PR #25, opened 2017-12-19
Do not overwrite transform in render_to_canvas

niklasf:
Hi, I just saw your announcement on reddit and tried [replacing](https://github.com/niklasf/rust-chessground/compare/resvg) librsvg with *libresvg* in my experimental chessboard project https://github.com/niklasf/rust-chessground.

It's using SVG to render the pieces and has to be able to apply arbitrary transforms before rendering. However `render_cairo::render_to_canvas` seems to disregard any previous transforms.

I tried to change that and it looks fine in my tests, but I am not sure if this is really correct.


RazrFalcon:
Yes. This a good idea. `render_to_canvas` is very experimental.

Update Qt backend as well and I will merge it.

https://github.com/RazrFalcon/libresvg/blob/master/src/render_qt/mod.rs#L80

```diff
-painter.set_transform(&ts);
+painter.apply_transform(&ts);
```

niklasf:
Done, thanks.

RazrFalcon:
Good. Thanks.

PR #57, opened 2018-05-14
Fix position of image border in examples/qt-demo app

N3WWN:


N3WWN:
> Its needed to center the image. Why should we remove it?

Ah, I see now... I hadn't noticed that my change left the image off-center.

Here's what I am seeing with the current code:

resvg demo - current

resvg demo - current - original size

Here's the result of my `4feada9` commit:

resvg demo - n3wwn commit 4feada9

Since seeing the `p.translate()` was needed for centering properly and fixing the centering:

resvg demo - n3wwn - new - fit

resvg demo - n3wwn - new - original

I'll be updating my PR to center properly in a few minutes.

RazrFalcon:
Fixed by https://github.com/RazrFalcon/resvg/commit/1bb82fbebd5695af75608db11563b8585be73034

PR #74, opened 2018-09-29
Added instructions for installing cairo-backend with homebrew

ocohen:


RazrFalcon:
Thanks. I don't have time to actually test it now, but it looks ok.

PR #76, opened 2018-10-05
Update cairo, pango and pangocairo

selaux:
Upgrade to version `0.5`.

Also changes shebangs of test scripts to `python3`, so they run on systems that do not have exactly `python3.6` installed.


RazrFalcon:
Thanks!

I depend on Python 3.6, so yes...

I will look up what's wrong with the example, so you can fix the CI.

selaux:
The only issue with the example now is that it requires rust version `1.27` because the `glib` trait uses some `dyn Trait` syntax...

RazrFalcon:
>The only issue with the example now is that it requires rust version 1.27 because the glib trait uses some dyn Trait syntax...

Hmm... the problem is that `resvg` trying to support Rust >= 1.22

But if this affects only examples - it's probably fine. Except we have to skip the example build on travis for 1.22.

RazrFalcon:
Looks like we should check for `TRAVIS_RUST_VERSION` variable in [here](https://github.com/RazrFalcon/resvg/blob/b35e1d514d0c585a0b9ef1b4d71e68b97507600f/testing_tools/run-tests.py#L102).

For this pull request you can simply comment this line and I will figure out how to make it optional.

selaux:
Looking at the latest CI run, this also affects building the library.

```
use of unstable library feature 'thread_local_state': state querying was recently added (see issue #27716)
```

As far as I can see `thread_local_state` was made stable in 1.26.

The problem with not updating cairo is that from my experience rust libraries / applications that use resvg and also have a dependency on `cairo`, `pango` and maybe other `gtk-rs` libraries have a high chance of failing with a build result like

```
package `cairo-sys-rs v0.6.0`
    ... which is depended on by `resvg-capi v0.3.0 (file:///home/travis/build/RazrFalcon/resvg/capi)`
links to native library `cairo`
package `cairo-sys-rs v0.7.0`
    ... which is depended on by `pangocairo-sys v0.8.0`
    ... which is depended on by `pangocairo v0.6.0`
    ... which is depended on by `resvg v0.3.0 (file:///home/travis/build/RazrFalcon/resvg)`
    ... which is depended on by `resvg-capi v0.3.0 (file:///home/travis/build/RazrFalcon/resvg/capi)`
```
With conflicting `*-sys` versions of crates they are forced to not update as well. But maybe there is a way around this that I dont know?

RazrFalcon:
The only solution I know is too use the latest stable... As everyone does.

I think we can simply remove `1.22.0` from travis for now.

PR #112, opened 2019-01-21
WIP some SVGView improvements

flying-sheep:
I improved the look and added code to use the actual screen’s DPI.

Sadly I’m not very good in C++, so it crashes. I guess the fact that I bring in references to the QWindow in from its thread into the worker thread is a problem?

In that case we can go another route, e.g. storing it at a fixed memory location, or going over events.

### DIFF
--- a/tools/viewsvg/mainwindow.ui
+++ b/tools/viewsvg/mainwindow.ui
@@ -15,14 +15,8 @@
   </property>
   <widget class="QWidget" name="centralWidget">
    <layout class="QVBoxLayout" name="verticalLayout_2">
-    <property name="spacing">
-     <number>6</number>
-    </property>
     <property name="leftMargin">
      <number>0</number>
-    </property>
-    <property name="topMargin">
-     <number>6</number>
     </property>
     <property name="rightMargin">
      <number>0</number>
@@ -33,7 +27,7 @@
     <item>
      <layout class="QHBoxLayout" name="horizontalLayout">
       <property name="leftMargin">
-       <number>2</number>
+       <number>6</number>
       </property>
       <item>
        <widget class="QLabel" name="label_3">
@@ -151,10 +145,10 @@
        </sizepolicy>
       </property>
       <property name="frameShape">
-       <enum>QFrame::StyledPanel</enum>
+       <enum>QFrame::NoFrame</enum>
       </property>
       <property name="frameShadow">
-       <enum>QFrame::Sunken</enum>
+       <enum>QFrame::Plain</enum>
       </property>
      </widget>
     </item>
--- a/tools/viewsvg/svgview.cpp
+++ b/tools/viewsvg/svgview.cpp
@@ -5,16 +5,13 @@
 #include <QPainter>
 #include <QThread>
 #include <QTimer>
-#include <QWindow>
 
 #include "svgview.h"
 
-SvgViewWorker::SvgViewWorker(QWidget *parent)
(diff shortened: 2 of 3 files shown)


tools/viewsvg/mainwindow.ui
```diff
@@ -15,15 +15,9 @@
   </property>
   <widget class="QWidget" name="centralWidget">
    <layout class="QVBoxLayout" name="verticalLayout_2">
-    <property name="spacing">
```
flying-sheep: This is the default right? So I just removed it.

tools/viewsvg/mainwindow.ui
```diff
@@ -33,7 +27,7 @@
     <item>
      <layout class="QHBoxLayout" name="horizontalLayout">
       <property name="leftMargin">
-       <number>2</number>
+       <number>6</number>
```
flying-sheep: The text was very close to the window border on my system.

tools/viewsvg/mainwindow.ui
```diff
@@ -151,10 +145,10 @@
        </sizepolicy>
       </property>
       <property name="frameShape">
-       <enum>QFrame::StyledPanel</enum>
+       <enum>QFrame::NoFrame</enum>
```
flying-sheep: This and the next change remove all border from the area, allowing an elegant borderless window

RazrFalcon:
Yes. You cannot pass QWidget to a thread. 

What are you trying to achieve? Dynamic DPI change?

tools/viewsvg/svgview.cpp
```diff
@@ -44,6 +47,11 @@ QString SvgViewWorker::loadFile(const QString &path)
     return QString();
 }
 
+void SvgViewWorker::setScreenDPI(const QScreen *screen)
+{
+    m_dpiRatio = screen->devicePixelRatio();
```
flying-sheep: I tried to catch a nullpointer here, but it doesn’t help with the crashes. Probably still a good idea…

RazrFalcon:
I think it will be easier for me to implement this.

Also, the frame looks nice on my OS. Which one do you use?

flying-sheep:
> Yes. You cannot pass QWidget to a thread.
> 
> What are you trying to achieve? Dynamic DPI change?

If you move the window to a screen with a different DPI, the worker will retain the old one.

Also, `qApp->screens().first()` is just generally wrong. `aApp->primaryScreen()` is a bit more correct (It’s the screen on which shown QWindows pop up), but `widget->windowHandle()->screen()` is more correct.

flying-sheep:
> Also, the frame looks nice on my OS. Which one do you use?

KDE with breeze window decorations

RazrFalcon:
Ok. I will see what I can do.

flying-sheep:
I think `QFrame::NoFrame` and `QFrame::Plain` are a good idea when you have a main viewing panel. All viewers and editors in KDE do it: Gwenview, Krita, Kate, Okular, …

And as said: That’ll make it look good on macOS too! What OS do you have?

RazrFalcon:
I have KDE too. With a breeze dark. But I don't think that removing a frame border is a good idea.

flying-sheep:
Why not? As said: all the viewers do it, and this fits right into that category.

flying-sheep:
So what do you say? I still think it should get rid of the border.

RazrFalcon:
Done.

flying-sheep:
great, thank you!

PR #128, opened 2019-04-02
[WIP] PoC of resvg-convert (rsvg-convert like tool)

Rush:
**NOTE: This is my first ever written Rust code - I won't lie, a lot of it was trial and error**

What is working (and is working well):
- Output to STDOUT `resvg-convert in.svg -f png`
- Conversion to SVG with text converted to path `resvg-convert in.svg -f svg -o out.svg`
- .. and to other vector based formats "pdf", "eps", "ps".

What is not working:
- `--export-id` is exporting an object in raster images
- Not done in the Qt backend. Not sure if even doable?

What is left to do:
- Fix build
- Add tests
- Taking SVG input from standard input
- Go through other rsvg-convert options
- Get reviews and feedback.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -62,6 +62,11 @@ version = "1.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
+name = "c_vec"
+version = "1.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "cairo-example"
 version = "0.1.0"
 dependencies = [
@@ -529,7 +534,9 @@ dependencies = [
 name = "resvg"
 version = "0.6.1"
 dependencies = [
+ "c_vec 1.3.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "cairo-rs 0.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "cairo-sys-rs 0.8.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "gdk-pixbuf 0.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.4.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "pango 0.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
@@ -547,6 +554,17 @@ dependencies = [
  "fern 0.5.7 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.4.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "resvg 0.6.1",
+]
+
+[[package]]
+name = "resvg-convert"
+version = "0.6.1"
+dependencies = [
+ "fern 0.5.7 (registry+https://github.com/rust-lang/crates.io-index)",
+ "gumdrop 0.5.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "log 0.4.6 (registry+https://github.com/rust-lang/crates.io-index)",
+ "resvg 0.6.1",
+ "time 0.1.42 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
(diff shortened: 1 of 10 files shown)


RazrFalcon:
>Not done in the Qt backend. Not sure if even doable?

Qt also supports PDF. And `resvg` must support the same set of features for all backends. Otherwise it's no go.

>Conversion to SVG with text converted to path

This is very bad. I want to omit this. And it's not possible at the moment.

>Taking SVG input from standard input

This is a separate issue.

Anyway, as I said previously, I don't want to rely on backend's internal PDF support. Mainly because there will be more backends in the future, so the PDF backend should be standalone and handwritten. Also, they are pretty unpredictable and I want a full control over the output.

Thanks for your effort, but I can't accept this pull request.

Rush:
So what's your advice? I need this tool. Should I maintain a fork of `resvg` in my own repo then?

Rush:
> This is very bad. I want to omit this. And it's not possible at the moment.
Can you explain what's very bad? As I said rsvg-convert exports SVGs with texts converted to path and I need this very behavior. At least for my use case this PR does the job.

In fact it's already running in a production environment. The question is now on how do I tie it in into `resvg`.

RazrFalcon:
>I need this tool.

I'm not sure what exactly do you need. As I said before, if we will use backend's PDF implementation it will create raster images anyway. So it's easier to embed PNG directly.

>Should I maintain a fork of resvg in my own repo then?

If you want it right now - yes. I will not use backend's PDF implementation anyway, as I explained earlier. 

Could you explain why do you need this and in what form do you plan to use it?

1. Do you need a single page PDF?
1. Do you need an ability to embed SVG into an existing PDF? In this case `resvg` should produce some intermediate output.
1. Why not PNG? Because of scaling? If so, SVG cannot be converted without raster layers anyway (not sure what features does PDF support).
1. Why do you need text to path conversion? It will create a huge file with poor performance. On the other hand, we have to embed custom fonts to the PDF to use text.
1. What about filters? How can we translate Gaussian blur to PDF? Should it be ignored? Should it be optional?

I'm not against the PDF support, but I'm not sure what result do you expect to achieve.

Rush:
> if we will use backend's PDF implementation it will create raster images anyway. So it's easier to embed PNG directly.

I will reply to your other questions in depth later but I want to answer the above first.
Given this SVG https://rush.sh/Gjv2o6FVC3hmT0edWku7Dr7oVPE/0c1cUj.raw.svg (which is representative of what I'm doing)

`resvg-convert 0c1cUj.raw.svg -f pdf -o 0c1cUj.pdf` 

It generates a PDF that is fully vector.

Rush:
> I'm not sure what exactly do you need. As I said before, if we will use backend's PDF implementation it will create raster images anyway. So it's easier to embed PNG directly.

I need `rsvg-convert`-like functionality to pipe-in SVG from stdin and pipe out results to std-out in the following formats:
- SVG with text converted to paths. I don't care that the resulting SVG is huge - I care for it to render correctly even in web browsers which don't have the fonts installed.
- PDF with vector properties preserved
- PNG with configurable DPI (for example small DPI is used to generate thumbnails)

`resvg-convert` which this PR implements satisfies all of the above requirements. But I would like to avoid maintaining a fork and rather (at least) contribute some parts of it to resvg itself. I'm fine maintaining the `resvg-convert` tool itself.

> Could you explain why do you need this and in what form do you plan to use it?
I use it as a utility function in a SaaS product (I can share which one but I don't want to be accused of using this thread for advertising). The results are viewed through a web browser. 

> * Do you need a single page PDF?
Yes, just one page with the dimensions of the original SVG file.

> * Do you need an ability to embed SVG into an existing PDF? In this case `resvg` should produce some intermediate output.
No - just pure SVG -> PDF. Same as `rsvg-convert` does. 

> * Why not PNG? Because of scaling? If so, SVG cannot be converted without raster layers anyway (not sure what features does PDF support).
 
PNG is not good for printing, not good for zooming. In my case SVG files are also smaller than respective rendered SVG file. SVG can also be post-processed and DOM events can be attached to it.

> * Why do you need text to path conversion? It will create a huge file with poor performance. On the other hand, we have to embed custom fonts to the PDF to use text.

It's for the web. Users may not have the right fonts installed. Huge file is not a concern as it's actually not that huge in practice. (smaller than a PNG file anyway) Performance seems OK (remember I've been using rsvg-convert before)

> 5\. What about filters? How can we translate Gaussian blur to PDF? Should it be ignored? Should it be optional?

Preferably rasterized - honoring the DPI setting.

RazrFalcon:
Thanks for feedback. I will look into it.

Rush:
@RazrFalcon I hope you're doing well. Do you think resvg would serve my use-case well today? I am still using my modified resvg library from a year ago. Can I update and throw away my old code?

RazrFalcon:
Sadly, no. The next release will specifically forbid such use case. svg2pdf must be written from scratch.

Rush:
That's unfortunate as my PR has served me well for the last year. I guess I will just keep using the old call.

RazrFalcon:
Sure, but it works only in very specific range. And resvg aims to support most of the spec.

Hopefully, it will be implemented one day.

PR #137, opened 2019-05-14
Raqote backend

### LINKED ISSUE #134: Implement a raqote backend
@RazrFalcon have you started on this?

RazrFalcon:
TODO:

- [x] Ugly stroking: `e-image-003` (especially `S`), `e-image-017`, `e-image-022` (clearly visible)  (raqote bug)
- [x] `a-stroke-dasharray-004` (raqote bug)
- [x] `a-stroke-linecap-004` (should be done on the resvg size)
- [x] `a-stroke-linecap-005` (should be done on the resvg size)
- [x] `e-clipPath-019`
- [x] `e-feBlend-002` (raqote bug, `BlendMode::Multiply` produce nothing)
- [x] `e-image-010` (raqote bug? Looks like some issue with a pattern.)
- [x] `e-mask-007` (~~raqote bug? Looks like something is wrong with clip~~)
- [x] `e-path-013` (raqote bug)
- [x] `e-radialGradient-011` (raqote bug)
- [x] `e-radialGradient-029`
- [x] ~~`e-stop-016` (raqote bug)~~
- [x] Markers are very slow, aka multiple sublayers.
- [x] Optional anti-aliasing.

Closes #134


RazrFalcon:
@jrmuizel here is the current progress.

jrmuizel:
What would you say is the highest priority functionality that you need from raqote at this point?

RazrFalcon:
Gradients and layers compositing with opacity.

jrmuizel:
Ok, I'm going to work on the opacity problem next. What's missing for gradients?

RazrFalcon:
>What's missing for gradients?

Everything. There are no spread method, coordinates (see [linear](https://doc.qt.io/qt-5/qlineargradient.html#QLinearGradient-2) and [radial](https://doc.qt.io/qt-5/qradialgradient.html#QRadialGradient-6) constructors in Qt), object bbox transform.

jrmuizel:
Gradients now have spread.

jrmuizel:
It seems like gradients and layer opacity are in decent shape now. What's important to get next?

RazrFalcon:
Current tasks should be written on the resvg side first. So nothing important at the moment.

jrmuizel:
Do you have a more up to date version of this locally? i.e. one that includes gradient support? I'd like to look into hooking up more of the missing functionality.

RazrFalcon:
I've pushed the latest changes. It's not much, since I'm working on the text layout right now.

jrmuizel:
I started working on clipping here: https://github.com/jrmuizel/resvg/tree/clip

RazrFalcon:
Nice.

jrmuizel:
I have untested image and filter support now too.
https://github.com/jrmuizel/resvg/tree/raqote

RazrFalcon:
You can test it via [vdiff](https://github.com/RazrFalcon/resvg-test-suite/tree/master/tools/vdiff). Just replace `cairo` or `qt` with `raqote` [here](https://github.com/RazrFalcon/resvg-test-suite/blob/c2b3384f21fdeb38db00ddfbfb8bba48e94272bb/tools/vdiff/src/render.cpp#L121).

jrmuizel:
Great. That's just what I was going to ask.

RazrFalcon:
Could you merge current changes to the branch? I plan to finish this backend soon and release a new version.

jrmuizel:
My latest changes are here: https://github.com/jrmuizel/resvg/tree/raqote. The filter code is probably the area with the most bugs remaining right now.

RazrFalcon:
There are strange artifacts on `e-image-017`. An inner, green path is broken down.

RazrFalcon:
It mostly works. This is very good! Patters are nice and crisp. I spend weeks writing them. cairo backend still has some bugs.

Main issues: filters, stroke-dasharray, optional anti-aliasing, blurry clip-path (`e-clipPath-019`), masking (`e-mask-007`), some paths are still broken (`e-path-013`), radial gradients are still a bit off, 

Some images take a lot of time to render. For example, `e-marker-022`.

jrmuizel:
The performance problems are probably caused by the fact that we composite the entire canvas regardless of the bounds of the path. This shouldn't be too hard to fix but I've wanted to get feature complete and decent correctness before looking at fixing performance.

I'll take a look a look at the stroke-dasharray problem next.

RazrFalcon:
Yes, region based rendering is a long-term optimization. It wasn't possible previously, because of text. But now, since we know all bboxes before hand we can do this.

jrmuizel:
I actually meant inside raqote. Currently if you fill a 1x1 rect it takes time proportional to the size of the canvas.

Some more of the stroke-dasharray tests pass now. It looks like there's still an issue with the last segment that I'll look at when I next get a chance.

RazrFalcon:
Oh, I see.

RazrFalcon:
I've merged your changes.

RazrFalcon:
Don't worry about CI, it will always fail due to how regression testing is implemented.

RazrFalcon:
I've updated the first post. We almost there!

RazrFalcon:
Ran some benchmarks:

- cairo - ~250sec
- raqote - ~450sec

This is time to render Oxygen Icon theme.

jrmuizel:
Are there instructions for how to run the benchmarks?

I've fixed the dasharray tests. It looks like the negative dash offset test is still or started failing. I'll look at that next.

RazrFalcon:
>Are there instructions for how to run the benchmarks?

I simply converting images one by one. [Here is](https://github.com/RazrFalcon/resvg-test-suite/tree/master/tools/perf) the source code. [Download link](https://www.archlinux.org/packages/extra/any/oxygen-icons-svg/download/) for Oxygen Icon Theme. I couldn't find the 4.12 one, but it should be fine. 

I'm testing svg files and not svgz one. You can do this [via](https://github.com/RazrFalcon/svgcleaner/blob/master/tools/files-testing/README.md#convert-svgz-to-svg).

RazrFalcon:
About benchmarking, here is a better option. Try rendering this file.

```
> rendersvg -backend=raqote -w 1000 --perf test.svg out.png
Preprocessing: 15.80ms
Rendering: 7449.89ms
Saving: 40.68ms

> rendersvg -backend=cairo -w 1000 --perf test.svg out.png
Preprocessing: 15.41ms
Rendering: 196.20ms
Saving: 124.41ms

> rendersvg -backend=qt -w 1000 --perf test.svg out.png
Preprocessing: 15.11ms
Rendering: 209.96ms
Saving: 120.96ms
```

PNG generation in Rust/image/raqote is suspiciously fast.

jrmuizel:
Negative dash offsets are now fixed.

jrmuizel:
I had a brief look at `e-feBlend-002`. It looks like it's missing a white background from some place. From looking at the svg it seems like the `lightblue` should blend with transparency to be transparent. I'm not sure where the white is supposed to come from.

RazrFalcon:
`e-feBlend-002` blends `lightblue` with `green`. There are no opacity.

`a-stroke-dasharray-004` was fixed.

jrmuizel:
e-feBlend-002 is fixed now.

RazrFalcon:
Yep.

jrmuizel:
`e-path-013` is a flattening tolerance issue.

jrmuizel:
The e-image failures are flattening tolerance problems as well.

RazrFalcon:
Is it lyon bug?

jrmuizel:
I'm not sure yet. I'll need to dig in further.

jrmuizel:
I had a look at performance of the test.svg you posted. Most of the time is being spent in draw_image_at(). There's quite a bit of potential for adding fast paths to this so there should be lots of room for speed up.

There's also a lot of time being spent in clear().
The following change brought rendering time from 6391.86ms down to 5465.51ms for me.

```patch
diff --git a/src/backend_raqote/mod.rs b/src/backend_raqote/mod.rs
index 2702d13..47d3a05 100644
--- a/src/backend_raqote/mod.rs
+++ b/src/backend_raqote/mod.rs
@@ -324,13 +324,14 @@ fn create_layers(img_size: ScreenSize, opt: &Options) -> RaqoteLayers {
 }

 fn create_subsurface(
     size: ScreenSize,
     _: f64,
 ) -> Option<raqote::DrawTarget> {
     Some(raqote::DrawTarget::new(size.width() as i32, size.height() as i32))
 }

 fn clear_subsurface(dt: &mut raqote::DrawTarget) {
-    dt.set_transform(&raqote::Transform::identity());
-    dt.clear(raqote::SolidSource { r: 0, g: 0, b: 0, a: 0 });
+    for i in dt.get_data_u8_mut() {
+        *i = 0;
+    }
 }
```

RazrFalcon:
I thought that `clear` does implemented like this. At least in cairo/Qt this method is basically a `fill()`

jrmuizel:
DrawTarget::clear() needs to respect the clip and so in raqote it just goes through the generic composite path. Cairo detects situations where it can take a memset like fast path but raqote doesn't have any blitter fast paths yet.

RazrFalcon:
I see. I will add this fix for now.

jrmuizel:
https://github.com/jrmuizel/raqote/tree/rasterizer-bounds has proof-of-concept implementation restricting composition to the path bounds. It speeds things up some on the test.svg but there are bigger problems elsewhere like https://github.com/jrmuizel/raqote/issues/61

RazrFalcon:
`rasterizer-bounds` almost twice as fast. Impressive.

RazrFalcon:
Oxygen "benchmark" dropped down from ~450sec to ~390sec with `rasterizer-bounds` and faster `clear`.

jrmuizel:
e-stop-016 is caused by https://github.com/jrmuizel/raqote/issues/62. I haven't decided what to do about it yet.

jrmuizel:
`e-mask-007` is caused by

```rust
        let mut pb = raqote::PathBuilder::new();
        pb.rect(r.x() as f32, r.y() as f32, r.width() as f32, r.height() as f32);
        mask_dt.push_clip(&pb.finish());
```

with `r =  (x = -4000, y = -4000, width = 48000, height = 48000)`

This runs into some issues that raqote has with handling edges that are much larger than the canvas.

RazrFalcon:
It's also a `resvg` bug. Mask rect should be `Rect(0.1 0.1 1.2 1.2)` and not `Rect(-20 -20 240 240)`. But since it affects only clip-rect - it does work as expected.

I will add a new test for this and fix this issue in resvg.

RazrFalcon:
Fixed.

jrmuizel:
I added a fix for https://github.com/jrmuizel/raqote/issues/61 that should help performance quite a bit.

RazrFalcon:
Well done! It's ~360sec now.

Elementary OS icon theme rendering is just ~22sec, exactly the same as cairo. So in rendering single layer, single color, curves only images they are mostly on par.

jrmuizel:
Are you sure about those numbers? I don't expect it to be that good yet.

RazrFalcon:
It's only according to this benchmark. I think the issue is that cairo/Qt are using high level PNG compression. While raqote/image creates a bigger file, but faster. Here is an example (rendering to 1000x1000px):

cairo:
```
Preprocessing: 1.17ms
Rendering: 42.42ms
Saving: 93.47ms
```

raqote:
```
Preprocessing: 1.18ms
Rendering: 389.47ms
Saving: 37.65ms
```

[camera.zip](https://github.com/RazrFalcon/resvg/files/3291432/camera.zip)

PS: this image also has broken outlining on inner circles.

RazrFalcon:
Raqote also doesn't scale well. By rendering 500x500 and 1000x1000 image, cairo is 3.2x slower, while raqote is 3.8 slower (almost linear, actually).

jrmuizel:
`e-image-003`, `e-image-017`, `e-image-022` should all be fixed now. camera.svg also looks better. Is the issue you were seeing fixed?

RazrFalcon:
Yes, it's way better. But you still have a small artifact on `e-image-017` and `e-image-022`.

jrmuizel:
The artifact that I think you're talking about can be fixed by closing the path for the rounded rect in `convert_rect`

RazrFalcon:
You're right. No sure why it works in other backends. I'm closing this one.

jrmuizel:
A couple of new things:
- I pushed a change that seems to fix `e-image-010` and similar tests.
- https://github.com/jrmuizel/raqote/tree/rasterizer-bounds has a fix that should improve performance a little more.
- I looked at `e-path-013` a little more. What's basically happening is that after we flatten the path we we end up with line segments at sharp angles that get miter joined. Some other rasterizers like agg, ghostscript and old versions of Adobe Reader run into the same issue (try loading the attached pdf in Chrome vs pdf.js)
[cusp-join.pdf](https://github.com/RazrFalcon/resvg/files/3294458/cusp-join.pdf)
. Try loading Raqote seems to hit it a bit more often so I'm going to investigate it further but I don't think it should block merging the backend.

RazrFalcon:
Yes, looks like image rendering is fixed. But the scaling algorithm itself has visible "steps". No sure how it's called. cairo/Qt is much more smooth.

I've already tested the `rasterizer-bounds`. The last update in that branch was three days ago.

I agree, that `e-path-013` is a minor bug.

jrmuizel:
The steps in image scaling is somewhat intentional. It's the cost of a performance optimization. Skia has the same issue and you should be able to see it in Firefox or Chrome if you disable hardware acceleration.

I forgot to push the update to the `rasterizer-bounds` branch. It's pushed now.

jrmuizel:
I was looking at some of the gradient tests and notice a typo.
Line 185 of style.rs should be
```rust
    if g.fx == g.cx && g.fy == g.cy {
```

instead of
```rust
    if g.fx == g.cx && g.fy == g.fy {
```

RazrFalcon:
Thanks! This fixes `e-radialGradient-029`.

jrmuizel:
`e-radialGradient-011` should be fixed now.

RazrFalcon:
Nice! There are still small artifacts on `e-radialGradient-022`, but we can ignore them for now.

RazrFalcon:
Only two bugs left.

jrmuizel:
I've added optional anti-aliasing support. It can be turned off with DrawOptions::antialias = AntialiasMode::None;

RazrFalcon:
Nice, but doesn't work on images.

jrmuizel:
For images do you mean disabling bilinear filtering? Do you have an example svg?

RazrFalcon:
Yes.

`a-image-rendering-001.svg`

RazrFalcon:
Found yet another great image for performance testing: [camera.tar.gz](https://github.com/RazrFalcon/resvg/files/3305878/camera.tar.gz)

Rendering to 500x500

```
# cairo
Preprocessing: 2132.77ms
Rendering: 1384.87ms
Saving: 30.91ms

# qt
Preprocessing: 2136.06ms
Rendering: 2040.55ms
Saving: 32.12ms

# raqote
Preprocessing: 2217.75ms
Rendering: 45793.24ms
Saving: 11.05ms
```

Almost 33 times  slower!

jrmuizel:
Yeah, the slowness here is caused by raqote's opacity image blend path still being quite a bit slower than it could be and the fact that we don't hit the fast path in apply_to_canvas because of the non-0,0 offset.

jrmuizel:
I've added support for Nearest filtering for images. That should let you finish up the optional image antialiasing feature.

I've looked at `e-stop-016` some more and I'd recommend not blocking on it. Firefox and Chrome had that bug until late 2015 (https://bugs.chromium.org/p/skia/issues/detail?id=517#c13) and I'm not exactly sure how I want to fix it yet.

RazrFalcon:
Looking good.

Ok, I can ignore `e-stop-016` for now, since raqote is still an experimental backend.

So I'm waiting for raqote's version bump and I'm ready to merge it and release resvg today.

jrmuizel:
I published raqote 0.5.0

RazrFalcon:
ClipPath became broken after https://github.com/jrmuizel/raqote/commit/30d08412bc00b142fc40cf74e955459a285c332f.

jrmuizel:
Looking now

jrmuizel:
0.5.1 fixes it.

RazrFalcon:
Yep, fixed.

RazrFalcon:
Hooray! Thanks a lot for you work.

jrmuizel:
And thank you for your work! Having resvg around really helped push raqote into something usable.

jrmuizel:
I pushed a 0.5.2 that has the rasterizer bounding and another change that helps some more with the performance of `camera.tar.gz`

RazrFalcon:
Updated.

Since we have a working regression testing, updating will be less painful and unpredictable.

On the other hand, travis takes almost half an hour to run tests. Mainly because of debug mode. cairo/Qt itself are built in release mode, unlike raqote.

PR #154, opened 2019-07-12
Skia backend

JaFenix:
This merge request contains the addition of the Skia backend.  It is source compatible with v0.7.0, commit ec197b4.


Brooooooklyn:
I recommend https://github.com/rust-skia/rust-skia as backend which support skia m74 and vulkan and so on.

PR #155, opened 2019-07-12
Skia backend

### LINKED ISSUE #44: Skia backend

RazrFalcon:
TODO:

- [x] Sync with master.
- [ ] e-pattern-18
- [ ] e-marker-22 is very slow
- [x] Support older Skia.
- [x] Update BUILD.adoc
- [x] Test on all platforms.
- [ ] ~~Test on CI.~~

Closes #44


RazrFalcon:
@JaFenix [e-pattern-018.svg](https://github.com/RazrFalcon/resvg-test-suite/blob/master/svg/e-pattern-018.svg) has no antialiasing for some reasons. Not sure why. Maybe we should set filter quality on a shader somewhere?

RazrFalcon:
By the way, this is the only one "broken" file. Everything else is working already. At least on test suite.

RazrFalcon:
Works fine on Windows using Skia m76.

RazrFalcon:
@JaFenix I'm ready to merge. I've marked macOS as an unsupported Skia platform for now, because I don't have an ability to test it there right now. And everything else is working fine (except to minor problems noted in the first comment).

Since the majority of work was done by you, I think you should fork this branch and resubmit it as a new pull request with a singe commit. So it will be marked as your work in the git history.

JaFenix:
Ok thanks, looking at this now.

JaFenix:
@RazrFalcon  For M58, kRGBA_8888_SkColorType is not a valid color type for raster surfaces, so a null surface is returned by skiac_surface_create.  (Could this be the problem you saw on Linux?)  Apparently, it must be  kN32_SkColorType instead.  It looks like the color type parameter identifies the pre-multiplied color format, but it's configuration dependent, as defined by macros in SkPostConfig.h.  So kN32_SkColorType resolves to the correct byte order at compile type and is one of the only valid parameters to SkSurface::MakeRaster.   The unpremultiplied color order, however, appears to be the same regardless of the build config.

I didn't notice any issues with pixel manipulation, e.g. during filter processing.  Did you see any color problems using kN32_SkColorType in M58 and/or M76 for Linux/Windows?

RazrFalcon:
`kN32_SkColorType` breaks raster images rendering on Windows. Not sure why, but on Windows the channels order is inverted.

>Could this be the problem you saw on Linux?

No, they both work correctly on Linux.

RazrFalcon:
If we want to use `kN32_SkColorType`, we have to export it to resvg, so it can switch bytes order during [raster image rendering](https://github.com/RazrFalcon/resvg/blob/skia-backend/src/backend_utils/image.rs#L139).
I don't understand why bytes order is different between Linux in Windows. It should be the same.

JaFenix:
Is raster image rendering working on a premultiplied or rgba (unpremultiplied surface)?

RazrFalcon:
For skia I'm using unpremultiplied one, because I'm not sure what alpha multiplication algorithm it uses. But it doesn't matter anyway, because `kN32_SkColorType` uses BGRA on Windows.

JaFenix:
It says BGRA, but "PM" in the macros implies pre-multiplied, so I thought it might not apply for unpremultiplied surfaces:

Line 143 of SkColor.h:
`////////////////////////////////////////////////////////////////////////

/** 32 bit ARGB color value, premultiplied. The byte order for this value is
    configuration dependent, matching the format of kARGB32 bitmaps. This is different
    from SkColor, which is nonpremultiplied, and is always in the same byte order.
*/
typedef uint32_t SkPMColor;`

Line 172 of SkPostConfig.h
`/**
 * SK_PMCOLOR_BYTE_ORDER can be used to query the byte order of SkPMColor at compile time. The
 * relationship between the byte order and shift values depends on machine endianness. If the shift
 * order is R=0, G=8, B=16, A=24 then ((char*)&pmcolor)[0] will produce the R channel on a little
 * endian machine and the A channel on a big endian machine. Thus, given those shifts values,
 * SK_PMCOLOR_BYTE_ORDER(R,G,B,A) will be true on a little endian machine and
 * SK_PMCOLOR_BYTE_ORDER(A,B,G,R) will be true on a big endian machine.
 */`

I'll look into it some more.

JaFenix:
The SkSurface_Raster::Valid method is removed in m76, but is present and returns false in m58 unless the color type is kN32_SkColorType, in which case it defaults to BGRA for Windows which is apparently the "native" color type.

In v0.7.0, it looks like backend_utils::image::image_to_surface() wasn't being called in draw_raster (?), so this may be why the raster images looked OK when I was testing, because they were drawn directly to the surface.

So, I agree that the backend will have to say what color type it's using, so that backend_utils::image::image_to_surface can do the right thing.  Otherwise, copying an image to a surface could revert to being done by the native code.

RazrFalcon:
Will fix this soon.

>Otherwise, copying an image to a surface could revert to being done by the native code.

I'm trying to minimize an amount of an external code.

RazrFalcon:
Works fine for me now. Not sure about m58.

JaFenix:
Checking it out now

JaFenix:
m58 is working!  Looks OK.

Should I wait until the CI completes before issuing the pull request?

RazrFalcon:
No, you can do it right now. CI doesn't check Skia right now.

I'm closing this pr in favor of yours.

PR #156, opened 2019-07-15
Skia backend

JaFenix:


RazrFalcon:
Done! :tada: 

Thanks again for you work. I will add Skia to CI later, but only m76. If something will broke on m58 just let me know.

PS: maybe [a better email](https://github.com/RazrFalcon/resvg/blob/fba2345e074ac7d12f760530bcb24a608474fc23/bindings/resvg-skia/Cargo.toml#L4)?

JaFenix:
Thank you!   We're glad to see Skia included and we appreciate your help and the effort you've into it.  I'm looking forward to seeing the performance numbers for Skia and how it stacks up.  I'll keep the fork open on my account as I need to keep a few custom changes in, and I'll issue another pull request for the email.

RazrFalcon:
Performance is surprisingly low. Looks like a layer composition is pretty slow. But it is required for a proper clipping. I'm looking for a way to optimize it (https://github.com/RazrFalcon/resvg/issues/146).

Currently, it is 2x slower that Qt backend. I think there is a bug somewhere, which makes layers composition that slow.

RazrFalcon:
Looks like most of the time is spend in `avx::clamp_gamut(unsigned long, void**, unsigned long, unsigned long, float, float, float, float, float, float, float, float)`. Which is invoked by layer composition, aka `SkiaFlatRender::pop_layer`.

The same method is 22 times faster in Qt. I do not understand this. Maybe I've build skia in debug mode somehow, but skia.so is just 9MiB, so it's probably a release.

JaFenix:
Have you seen this [build note](https://skia.org/user/build) about Clang vs. MSVC?

> If you depend on software rasterization, image decoding, or color space conversion and compile Skia with GCC, MSVC or another compiler, you will see dramatically worse performance than if you use Clang.

This appears to be something new for later versions > m58.  I don't think they were using Clang then.

RazrFalcon:
I've missed this paragraph somehow. Will try with clang then.

RazrFalcon:
That's ridiculous! Now `SkiaFlatRender::pop_layer` is 22 times faster. Absurd!

The main difference is that Skia built with clang uses `avx::lowp::load_8888` and not `avx::load_8888`.

JaFenix:
Does that mean it has about the same performance as Qt using Clang?

RazrFalcon:
I've ran benchmarks again and Skia (clang) is ~5% faster than Qt (gcc) now during Oxygen icon theme rendering. And 60% faster during elementary icon theme rendering (Qt has a very slow path rendering algorithm). The same as cairo.

Which is an expected performance and make Skia the fastest backend.

JaFenix:
Awesome!

PR #162, opened 2019-08-01
Update resvg.h for use with C++

Tibbel:
Added extern "C" for use with C++


RazrFalcon:
Agree. You should then remove this line:

https://github.com/RazrFalcon/resvg/blob/3c1adbd09942c0e6f4fd5c6efdba59f2782b1916/capi/include/ResvgQt.h#L18

PR #171, opened 2019-09-03
README: Add a link to BUILD.adoc

ngirard:
Not a big deal, but otherwise, it's easy to miss it.


RazrFalcon:
I don't think that this is necessary. This is a generic file, like CHANGELOG, INSTALL, README, LICENSE, etc. No need to add it to readme.

Issue #208, opened 2020-02-19
how to run/use resvg


JoKalliauer:
I would like to render a svg as png, but I do not know how to start resvg. I tried `resvg` and `cargo run resvg`, but it did not work.

I already succesfully build:
```bash
sudo apt install qtbase5-dev libharfbuzz-dev libcairo2-dev
cargo build --release --features "qt-backend"
cargo build --release --features "cairo-backend"
cargo build --release --features "raqote-backend"
```

RazrFalcon:
You should use [rendersvg](https://github.com/RazrFalcon/resvg/tree/master/tools/rendersvg), which is a CLI tool. `resvg` is just a library. 

I should note it somewhere...

RazrFalcon:
I guess we can close this now.

PR #244, opened 2020-02-28
solves #208

JoKalliauer:
source: https://github.com/RazrFalcon/resvg/issues/208#issuecomment-588319920

author: RazrFalcon


RazrFalcon:
This should be added to [this](https://github.com/RazrFalcon/resvg/blob/master/tools/rendersvg/README.md) readme.

PR #258, opened 2020-03-02
solves #208 #244

JoKalliauer:
I think there should be also added a link to https://github.com/RazrFalcon/resvg/blob/master/tools/rendersvg/README.md otherwise nobody will find it.


RazrFalcon:
A link to rendersvg is right at the begining: https://github.com/RazrFalcon/resvg#purpose 

>a CLI application (see tools/rendersvg)

Issue #195, opened 2020-01-02
Remove all backends except raqote


RazrFalcon:
Originally, when `resvg` was started (2.5 years ago) there was no [raqote](https://github.com/jrmuizel/raqote). So the only solution was to use an existing 2D libraries like cairo and Qt.

The initial requirements for a 2D library were:

- Shapes rendering (filling, stroking, gradients, patters).
- Layers compositing.
- Text rendering (the most complex one).
- JPEG/PNG handling.

But that was a long time ago. Right now, `resvg` requires only the first two points. Text rendering, JPEG/PNG handling, and filters are done in a backned-independent manner already. So the only reason to have multiple backends is to render very simple SVG images (no opacity, clipping, masking, filters, images) on an existing canvas.

So the idea is to remove all backends except `raqote`, while C API will be changed to return just a `Vec<u8>` as a resulting image (not PNG encoded).

This will simplify things quite a bit. The only downside is that `raqote` is a bit slow and unstable. But this can (and will) be fixed.

@JaFenix, @therealbnut if you are still using `resvg`, what are your thoughts about this change? Is it absolutely necessary to have a Skia based backend for you?

Shnatsel:
For context, this does not impact out-of-tree "backends" such as Lyon and Pathfinder - they use `usvg` crate only.

RazrFalcon:
Yes, `usvg` isn't going anywhere. `resvg` will still be split into two modules.

JaFenix:
For rendering purposes, there isn't anything specific about Skia that would be necessary for me.  Provided that Raqote executes safely and efficiently inside of an OpenGL (hardware accelerated), Skia-based application, then it should not be necessary to have a Skia specific backend.  It would be sufficient for the C API to return a pointer to a rendered bitmap.  

Performance and memory consumption are the only concerns.  The performance will ideally have to be as good as Skia's software renderer, or at least as good as a web browser (some of which use Skia), because our application competes with the browsers in terms of rendering performance.  I imagine there will be some additional memory overhead with loading the Raqote library into the application's process in addition to Skia.  Currently, we only load Skia, so this affords us a nice optimization.

[...]

PR #278, opened 2020-04-11
Add meson.build for distro packagers

pontaoski:
A meson.build wraps around Cargo.toml to manage installation of necessary files, allowing distro packagers to have an easier time packaging.

This also generates a pkgconfig file in order for C/C++ projects to be able to depend on resvg without needing to vendor it.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,28 @@
+project('resvg', 'c', version: '0.9.0')
+
+cargo = find_program('cargo', required: true)
+
+cargo_script = find_program('scripts/build.sh')
+
+lib = 'libresvg.so.0.'+meson.project_version()
+
+target = custom_target('resvg',
+                       build_by_default: true,
+                       install: true,
+                       install_dir: lib,
+                       console: true,
+                       output: 'libresvg.so.0.'+meson.project_version(),
+                       command: [cargo_script, '@SOURCE_ROOT@', '@OUTPUT@', meson.build_root(), get_option('buildtype'), get_option('backends')])
+
+if get_option('backends') == 'both' or get_option('backends') == 'qt'
+    install_headers(['capi/include/resvg.h', 'capi/include/ResvgQt.h'], subdir: 'resvg')
+endif
+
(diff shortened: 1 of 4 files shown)


RazrFalcon:
Thanks, that's a good idea.

Why do we need a shell script? Shouldn't this be handled by meson. I mean, usage of bash will make it linux only.

Also, why it supports only Qt and cairo? I plan to remove those backends completely. See #195 

PS: could you add newline to the end of the files?

pontaoski:
> Thanks, that's a good idea.
> 
> Why do we need a shell script? Shouldn't this be handled by meson. I mean, usage of bash will make it linux only.

Meson is intentionally designed to not be Turing-complete, and calling out to helper scripts for complex tasks is generally preferred to making your `meson.build` complex. I could replace it with a Python script if it was necessary, but a shell script will do here.

Also, being made for Linux/BSD is kinda the point. On Windows or macOS, you'd be much more likely to submodule this and access via Cargo because that's the convention for those platforms, but on Linux/BSD, you'd use a distro-provided shared library and include files instead of bundling with a project.
 
> Also, why it supports only Qt and cairo? I plan to remove those backends completely. See #195

Qt and Cairo are the only backends that you'd be packaging shared libraries for in distros. Skia is seldom packaged standalone, and Raqote, being Rust, would be distributed via source instead of compiled for the usage of building other packages.

Also, dropping Qt backend would kinda defeat the reason why I made this patch in the first place: so that it would be possible to depend on this without needing to submodule for KIconLoader.

> PS: could you add newline to the end of the files?

Sure.

RazrFalcon:
resvg is a crossplatform library. I do not plan to add any OS-specific stuff. So I cannot accept this request. Sorry.

>Also, dropping Qt backend would kinda defeat the reason why I made this patch in the first place: so that it would be possible to depend on this without needing to submodule for KIconLoader.

The point in dropping cairo and Qt is that they are not necessary. raqote is more than enough. You still can use resvg from KIconLoader.

PR #282, opened 2020-04-24
Add -h flag to usvg-cli to match usage text

kneitinger:
Hello, and thank you for your work on this crate, it is greatly appreciated!

I was using `usvg` and attempted to use `-h` which was not recognized.  `--help` shows it as a valid option, so this extraordinarily small PR just adds the flag in.


RazrFalcon:
Thanks!

PR #291, opened 2020-06-02
Bugfix: fixes dy calculation for textPath

Stoeoef:
There seems to be a subtle error when using `text` `textPath` in combination with the `dy` attribute:

Original (upper `textPath` has `dy` attribute, lower path doesn't):
Original

With this PR:
Fixed

Since the position is only off by a distance of 1, the error is only noticeable for small coordinate values.

Svg-file:
[TwoPathsWithDy.zip](https://github.com/RazrFalcon/resvg/files/4716439/TwoPathsWithDy.zip)

I'm a bit unsure on how to run the test suite and haven't done that - sorry! I can dig further into this if required.

```xml
<svg id="svg1" viewBox="0 0 2 2" width="500" height="500" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
	<title>
		Two paths with dy
	</title>
	<path id="pathForText1" d="M 0.20 0.73 C 0.35 1.08 0.85 1.08 1.00 0.73 C 1.15 0.38 1.65 0.38 1.80 0.73" fill="none" stroke="gray" stroke-width="0.03" />
	<path id="pathForText2" d="M 0.20 1.27 C 0.35 1.62 0.85 1.62 1.00 1.27 C 1.15 0.92 1.65 0.92 1.80 1.27" fill="none" stroke="gray" stroke-width="0.03" />
	<text id="text1" font-family="serif" font-size="0.24" stroke="none" dy="0.1">
		<textPath id="textPath1" xlink:href="#pathForText1">
			Some long text
		</textPath>
		<textPath id="textPath2" xlink:href="#pathForText2">
			Some long text
		</textPath>
	</text>
	<!-- image frame -->
	<rect id="frame" x="0.01" y="0.01" width="1.98" height="1.98" fill="none" stroke="black" stroke-width="0.1" />
</svg>
```

### DIFF
--- a/testing-tools/regression/allow-cairo.txt
+++ b/testing-tools/regression/allow-cairo.txt
@@ -1,1 +1,5 @@
+e-textPath-023.svg
+e-textPath-027.svg
+e-textPath-032.svg
+e-textPath-033.svg
 e-textPath-035.svg
--- a/testing-tools/regression/allow-cairo.txt
+++ b/testing-tools/regression/allow-cairo.txt
@@ -1,2 +1,1 @@
-e-image-038.svg
-e-image-039.svg
+e-textPath-035.svg
--- a/testing-tools/regression/allow-qt.txt
+++ b/testing-tools/regression/allow-qt.txt
@@ -1,1 +1,5 @@
+e-textPath-023.svg
+e-textPath-027.svg
+e-textPath-032.svg
+e-textPath-033.svg
 e-textPath-035.svg
--- a/testing-tools/regression/allow-qt.txt
+++ b/testing-tools/regression/allow-qt.txt
@@ -1,2 +1,1 @@
-e-image-038.svg
-e-image-039.svg
+e-textPath-035.svg
--- a/testing-tools/regression/allow-raqote.txt
+++ b/testing-tools/regression/allow-raqote.txt
@@ -1,1 +1,5 @@
+e-textPath-023.svg
+e-textPath-027.svg
+e-textPath-032.svg
+e-textPath-033.svg
 e-textPath-035.svg
--- a/testing-tools/regression/allow-raqote.txt
+++ b/testing-tools/regression/allow-raqote.txt
@@ -1,2 +1,1 @@
-e-image-038.svg
-e-image-039.svg
+e-textPath-035.svg
--- a/testing-tools/regression/allow-skia.txt
+++ b/testing-tools/regression/allow-skia.txt
@@ -1,1 +1,5 @@
+e-textPath-023.svg
+e-textPath-027.svg
+e-textPath-032.svg
+e-textPath-033.svg
 e-textPath-035.svg
--- a/testing-tools/regression/allow-skia.txt
+++ b/testing-tools/regression/allow-skia.txt
@@ -1,2 +1,1 @@
-e-image-038.svg
-e-image-039.svg
+e-textPath-035.svg
--- a/usvg/src/convert/text/shaper.rs
+++ b/usvg/src/convert/text/shaper.rs
@@ -556,7 +556,7 @@ fn resolve_clusters_positions_path(
         // Shift only by `dy` since we already applied `dx`
         // during offset along the path calculation.
         if !dy.is_fuzzy_zero() || !baseline_shift.is_fuzzy_zero() {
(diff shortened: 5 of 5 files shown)


RazrFalcon:
Yes, testing is a bit complicated, but looks like all tests were passed anyway. Which is strange. I though that removing an angle will affect it. 

Honestly, I wrote that code almost a year ago, so I don't really remember how it works. But since everything is ok - I guess it's fine.

I will add your test to the test suite and reran the CI.

Stoeoef:
Yep, I wondered about the original intent as well. The most likely interpretation IMO is that it's just a bug that got unnoticed as it only becomes clear once you go down to small coordinate values.

RazrFalcon:
Yes, that's why I have 1400 test files to not to worry about it =)

I've rerun the tests, and they fails as expected. Would you mind adding `e-textPath-035.svg` to `resvg/testing-tools/regression/allow-*.txt`

I know that testing is very complicated, but this is because SVG is complicated.

Buy the way, this is how it looks in other applications:

e-textPath-035

Looks like only Firefox works correctly.

Stoeoef:
> Buy the way, this is how it looks in other applications:

What a mess :) !

By the way, I'm really unsure if Chrome is not actually doing the right thing (and the rest isn't).
If I interpret [the spec](https://www.w3.org/TR/SVG11/text.html#TextElementDYAttribute) correctly, the second `textPath` should also be affected by `dy` as it is also a descendant. So we may have two different defects here of which only one is fixed by this PR!

Any opinion on how to proceed? We may want to create a test case only containing the first `textPath` until it's clear what the correct solution actually is!

Here's the svg doing that:
```xml
<svg id="svg1" viewBox="0 0 2 2" width="400" height="400" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
	<title>
		textPath with dy
	</title>
	<path  id="pathForText1" d="M 0.20 0.73 C 0.35 1.08 0.85 1.08 1.00 0.73 C 1.15 0.38 1.65 0.38 1.80 0.73" fill="none" stroke="gray" stroke-width="0.03" />
	<text id="text1" font-family="serif" font-size="0.24" stroke="none" dy="0.1">
		<textPath id="textPath1" xlink:href="#pathForText1">
			Some long text
		</textPath>
	</text>
	<!-- image frame -->
	<rect id="frame" x="0.01" y="0.01" width="1.98" height="1.98" fill="none" stroke="black" stroke-width="0.1" />
</svg>
```

Sorry for not seeing that earlier!
I'm currently investigating a different issue (`textPath` seems to ignore any `transform`ation of its path), maybe I'll find some time to find out how to replicate the chrome behavior.

RazrFalcon:
You can make other changes in separate pull request. It's fine to merge this one as is.

As for Y offset for the second `textPath`, I think it should be discarded. Since it affect only the first char and not the whole text. But I might be wrong. `textPath` in general is mostly undocumented and badly supported. So there is no correct way.

>textPath seems to ignore any transformation of its path

Yes, I've just noticed this too.

Stoeoef:
Thanks for the quick feedback. I've adjusted the allow-*.txt files as requested.

RazrFalcon:
Strange. It produced some changes this time. Looks like it affected aliasing in all backends except Qt. Which is even stranger.

As I said, SVG tests is a bit magical...

Basically, it's not that big of a deal, so we can push it as is. Just add 

```
e-textPath-023.svg
e-textPath-027.svg
e-textPath-032.svg
e-textPath-033.svg
```

to the same files.

Stoeoef:
I've added the 4 files as requested.

RazrFalcon:
Everything looks good now. Thanks for the contribution.

RazrFalcon:
I've added `e-textPath-036.svg` and `e-textPath-037.svg` in case you want to fix the incorrect transformation issue.

Stoeoef:
> I've added e-textPath-036.svg and e-textPath-037.svg in case you want to fix the incorrect transformation issue.

Thanks! I'll have a look at them but can't make any promises.

RazrFalcon:
Let me know about the results, so I could fix it myself.

PR #292, opened 2020-06-02
Bugfix: TextPath ignored transformed paths

Stoeoef:
Okay, here goes nothing. Transformation is now (somewhat hacky) calculated when the textPath is resolved.

### DIFF
--- a/usvg/src/convert/text/convert.rs
+++ b/usvg/src/convert/text/convert.rs
@@ -6,7 +6,7 @@ use std::cmp;
 use std::rc::Rc;
 
 use crate::{fontdb, svgtree, tree};
-use crate::convert::{prelude::*, style, units};
+use crate::{Transform, convert::{prelude::*, style, units}};
 use super::TextNode;
 
 
@@ -60,6 +60,7 @@ pub struct TextPath {
     pub start_offset: f64,
 
     pub path: tree::SharedPathData,
+    pub transform: Transform,
 }
 
 
@@ -317,9 +318,30 @@ fn resolve_text_flow(
         node.resolve_length(AId::StartOffset, state, 0.0)
     };
 
+    let node_transform = path_node.attribute::<Transform>(AId::Transform).unwrap_or_default();
+
+    // Get the absolute transform by looking at all parent elements
+    // Unfortunately, we can't use NodeExt::abs_transform since we're still 
+    // only dealing with svgtree::node
+    let mut transform_stack = vec![node_transform];
+
+    let mut current_node = path_node;
+    while let Some(parent) = current_node.parent_element() {
+        if let Some(parent_transform) = parent.attribute::<Transform>(AId::Transform) {
+            transform_stack.push(parent_transform);
+        }
+        current_node = parent;
+    }
+
+    let mut absolute_transform = Transform::default();
+    for ts in transform_stack.iter().rev() {
+        absolute_transform.append(ts);
+    }
+    
     Some(TextFlow::Path(Rc::new(TextPath {
         start_offset,
         path,
+        transform: absolute_transform,
     })))
 }
 
--- a/usvg/src/convert/text/convert.rs
+++ b/usvg/src/convert/text/convert.rs
@@ -5,8 +5,8 @@
 use std::cmp;
 use std::rc::Rc;
 
-use crate::{fontdb, svgtree, tree};
-use crate::{Transform, convert::{prelude::*, style, units}};
+use crate::{fontdb, svgtree, tree, Transform};
+use crate::convert::{prelude::*, style, units};
 use super::TextNode;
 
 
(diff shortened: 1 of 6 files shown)


usvg/src/convert/text/convert.rs
```diff
@@ -60,6 +60,7 @@ pub struct TextPath {
     pub start_offset: f64,
 
     pub path: tree::SharedPathData,
+    pub transform: Transform,
```
RazrFalcon, line 63: I guess we can apply the transform immediately. No need to store the transform separately.

usvg/src/convert/text/convert.rs
```diff
@@ -317,9 +318,30 @@ fn resolve_text_flow(
...
     };
 
+    let node_transform = path_node.attribute::<Transform>(AId::Transform).unwrap_or_default();
+
+    // Get the absolute transform by looking at all parent elements
+    // Unfortunately, we can't use NodeExt::abs_transform since we're still 
+    // only dealing with svgtree::node
+    let mut transform_stack = vec![node_transform];
+
+    let mut current_node = path_node;
```
RazrFalcon, line 328: Are you sure that we need an absolute transform? From my brief tests it look like only the element transform is used.

RazrFalcon:
Also, clear the `allow-*.txt` files.

PS: I'm in process of simplifying the testing procedure...

RazrFalcon:
It broken down only on a single backend, which bothers me. Changes to the `usvg` should affect all backends. Also, `e-textPath-037.svg` should not be affected.

usvg/src/convert/text/convert.rs
Stoeoef: Whoops - you are right! That wasn't my intuition, but the spec is pretty unambigouous here:

> The path data coordinates within the referenced ‘path’ element are assumed to be in the same coordinate system as the current ‘text’ element, **not in the coordinate system where the ‘path’ element is defined.** The ‘transform’ attribute on the referenced ‘path’ element represents a supplemental transformation relative to the current user coordinate system for the current ‘text’ element, including any adjustments to the current user coordinate system due to a possible ‘transform’ attribute on the current ‘text’ element.

Now I know. Ah well...
Thanks for the feedback!

Stoeoef: At least that explains why some of my svgs I was creating look so bogus :)

usvg/src/convert/text/convert.rs
Stoeoef: Good point. Done!

usvg/src/convert/text/convert.rs
```diff
@@ -317,6 +317,14 @@ fn resolve_text_flow(
         node.resolve_length(AId::StartOffset, state, 0.0)
     };
 
+    let path = if let Some(node_transform) = path_node.attribute::<Transform>(AId::Transform) {
```
RazrFalcon, line 320: Initially, I though that you are making an unnecessary cloning, but since the same path can be referenced by multiple textPath's, your version is actually correct.

Also, could you move it before `start_offset`?

usvg/src/convert/text/convert.rs
```diff
@@ -6,7 +6,7 @@ use std::cmp;
 use std::rc::Rc;
 
 use crate::{fontdb, svgtree, tree};
-use crate::convert::{prelude::*, style, units};
+use crate::{Transform, convert::{prelude::*, style, units}};
```
RazrFalcon, line 9: Would you mind changing it to:

```rust
use crate::{fontdb, svgtree, tree, Transform};
use crate::convert::{prelude::*, style, units};
```

Just for style =)

RazrFalcon:
Everything looks good. Remember to add `e-textPath-036.svg` to the `allow-*` files, since this file was affected by this change and it's ok to change.

usvg/src/convert/text/convert.rs
Stoeoef: Rust wouldn't even let me mutate the `Rc`, so I didn't have a choice. The copy on write should work out nicely I think.

Stoeoef:
All points addressed. Sorry for taking so much of your time and thank you for the quick feedback!
Should I do anything about the failing skia back end? It's a bit odd why just that fails...

Stoeoef:
Hmm, now some more back ends seem to fail. I'm confused and don't know what to include into the allow-*.txt files. At least all backends fail.
```
Failed: e-textPath-023.svg
Failed: e-textPath-027.svg
Failed: e-textPath-032.svg
Failed: e-textPath-033.svg
Failed: e-textPath-035.svg
```
I can't really judge if these failures are due to my changes or acceptable regressions. Let me know if I should modify something or feel free to take over - I think the implementation should be done by now.

RazrFalcon:
Looks like branch detection is broken in the Travis script and it uses previous master commit for pull requests instead of the actual master.

Stoeoef:
Okay. Thanks for looking into this.

RazrFalcon:
The current pull request is correct. The bug is in `testing-tools/run-tests.py`. I guess we can merge it like this. It should work correctly on master.

Thanks again! And don't worry about asking questions. I'm glad that someone is looking into this.

PR #296, opened 2020-06-08
Enhance viewsvg.pro with RPATH, debug_and_release

ferdnyc:
- Config set up for `debug_and_release` (`release` is default)
  - `make` or `make release` will use release configs
  - `make debug` will build in debug mode
    (`qmake` does not need to be re-run to switch mode)
- `QMAKE_RPATHDIR` is set for both build modes, to avoid requiring `LD_LIBRARY_PATH` at runtime
- Readme updated accordingly


RazrFalcon:
Thanks.

PR #315, opened 2020-08-11
Rewrite explorer-thumbnailer in Rust

gentoo90:
I wanted to get my hands on [`com-rs`](https://crates.io/crates/com) and this seemed like a good place to start.

It's a little messy, lacks proper error handling and installer, but it works and I thought I should open a pull-request and see if you'd be willing to merge it before cleaning things up.


RazrFalcon:
Wow! This is really cool. I didn't know you can do this in Rust. I'm looking forward to it.

tools/explorer-thumbnailer/src/logging.rs
```diff
@@ -0,0 +1,32 @@
+#[cfg(feature = "winlog")]
```
RazrFalcon, line 1: Why it's optional?

tools/explorer-thumbnailer/src/logging.rs
```diff
@@ -0,0 +1,32 @@
+#[cfg(feature = "winlog")]
+static WINLOG_SOURCE: &'static str = "SVG Thumbnailer";
```
RazrFalcon, line 2: Maybe it should be resvg thumbnailer?

gentoo90:
That should be all.
Not sure where to get `vc_redist.x64.exe` for the installer, so I'll leave this to you.
Please note that due to microsoft/com-rs#157 filename can only be `server.dll`.

P.S.: I also made a [small CLI utility](https://github.com/gentoo90/DumpThumbnail) which calls a thumbnailer and dumps its output into a PNG file. Can be used for integration testing or whatever.

tools/explorer-thumbnailer/src/logging.rs
gentoo90: I thought I'd add it just for debugging, but if you're ok with it...

tools/explorer-thumbnailer/src/logging.rs
gentoo90: Done.

RazrFalcon:
1. You should remove `explorer-thumbnailer-old` completely.
1. Exclude thumbnailer from the global cargo workspace. 
1. We can change the licence now to MPL2.0, just like resvg itself.

tools/explorer-thumbnailer/src/lib.rs
```diff
@@ -0,0 +1,88 @@
+use com::registration::{RegistryKeyInfo, dll_register_server, dll_unregister_server};
+use thumbnail_provider::{CLSID_THUMBNAIL_PROVIDER_CLASS, ThumbnailProvider};
+
+mod error;
+mod utils;
+mod interfaces;
+mod thumbnail_provider;
+
+// we replace the com::registration::inproc_dll_module macro in order to be able
+// to modify DllRegisterServer and DllUnregisterServer functions
```
rylev, line 10: COM-rs maintainer here 👋. I wonder if we should provide hooks so that you don't have to copy/paste these definitions over.

gentoo90:
> * You should remove `explorer-thumbnailer-old` completely.
> * Exclude thumbnailer from the global cargo workspace.
> * We can change the licence now to MPL2.0, just like resvg itself.

Done

tools/explorer-thumbnailer/src/lib.rs
gentoo90: Yes, that would be great! Also `impl Display for GUID` please, so we can use it in registry key paths like

https://github.com/gentoo90/resvg/blob/explorer-thumbnailer/tools/explorer-thumbnailer/src/lib.rs#L67

RazrFalcon:
Good. Should we wait for the `com-rs` update or we can merge it like this for now?

gentoo90:
Well, I don't see any showstoppers. I think this can be merged.

RazrFalcon:
Nice. Thanks again!

rylev:
I'd be happy to help update this to the new version of the `com` crate when we release it. It's very helpful to see the crate used in a project.

RazrFalcon:
Thanks. I do not plan releasing it any time soon, so there is no hurry.

PR #326, opened 2020-11-03
Fix text path placement bug

chubei-oppen:

PR #327, opened 2020-11-08
Update kurbo

cmyr:
This should unblock #326.


RazrFalcon:
Thanks. I think @chubei-oppen should do this as part of #326

cmyr:
sounds good :)

PR #329, opened 2020-11-17
(usvg) Fix wrong letter spacing of last cluster

chubei-oppen:
Previously we decide if a cluster is the last one using the last byte
index of the last span, which only works for 1-byte-long characters.

Example:
| lib | pic|
| ----------- | ----------- |
| svg |  |
| resvg(mater) | bad |
| resvg(PR) | good |

This PR also affects [e-textPath-030](https://github.com/RazrFalcon/resvg/blob/master/tests/svg/e-textPath-030.svg):
| lib | pic|
| ----------- | ----------- |
| resvg(master) | master |
| resvg(PR) | PR |


RazrFalcon:
Good catch. Looks like Firefox and Batik have the same bug.

PR #330, opened 2020-11-17
(usvg) Rotate cluster around half width

chubei-oppen:
Previously clusters are rotated around half advance when placed on `textPath`, which looks strange when letter spacing is big.

Example:
| lib | pic|
| ----------- | ----------- |
| svg |  |
| resvg(mater) | bad |
| resvg(PR) | good |

This PR also affects [e-textPath-030](https://github.com/RazrFalcon/resvg/blob/master/tests/svg/e-textPath-030.svg) and [e-textPath-031](https://github.com/RazrFalcon/resvg/blob/master/tests/svg/e-textPath-031.svg):
| lib | pic|
| ----------- | ----------- |
| resvg(master) | master |
| resvg(PR) | PR |
| resvg(master) | master |
| resvg(PR) | PR |


chubei-oppen:
I'm not sure what the correct behavior is. Firefox behaves like resvg(master) while Chrome like resvg(PR).

RazrFalcon:
Batik and Inkscape are producing the same result as Chrome, so looks like a Firefox (and resvg) bug.

`textPath` support in general is pretty bad everywhere, so such weird bugs is no surprise.

Issue #229, opened 2020-02-26
Render text in wasm32-unknown-unknown


dabreegster:
usvg currently works great in wasm environments as long as the optional "text" feature is disabled. I'd love to get text support working too. From my understanding so far, there are at least 3 issues:

- a dependency on memmap2, used for reading fonts
- reading fonts from system directories (#202)
- a dependency on harfbuzz

The first two are relatively straightforward. As part of #202, design an API to feed a FontDB into usvg that can be instantiated many different ways, one of them just being raw bytes of a font file. (include_bytes! is a low-effort way to get this working in a wasm environment.)

The harfbuzz dependency seems harder. My requirements are limited to simple Latin left-to-right text; harfbuzz is overkill. I found [font2svg](https://github.com/RazrFalcon/ttf-parser/blob/master/examples/font2svg.rs) and think it might be possible to make something much simpler. Before I go down this rabbit hole, I wanted to ask for your advice.

### Background

I'm using usvg in [A/B Street](https://github.com/dabreegster/abstreet) to turn both normal SVG files and text into paths for lyon, which then give me polygons. [This](https://github.com/dabreegster/abstreet/blob/master/ezgui/src/svg.rs) does the usvg->lyon->my simple Polygon abstraction path, based off [this example](https://github.com/nical/lyon/blob/0d0ee771180fb317b986d9cf30266722e0773e01/examples/wgpu_svg/src/main.rs) from lyon. I'm using this same pipeline for [text](https://github.com/dabreegster/abstreet/blob/1ffcd1e5b53e2cbe25421049797629af20058586/ezgui/src/text.rs#L264) as well. (There are many inefficient and silly things that the text pipeline does right now, like constantly re-load fonts.) In a few weeks, I'm hoping to properly split off my drawing, geometry, and GUI libraries as new crates for other people to use. These libraries attempt to give exactly the same experience on native and web -- not using system UI controls/style -- without any changes to user code needed for different platforms.

RazrFalcon:
>The harfbuzz dependency seems harder.

Believe me, I know =) https://github.com/RazrFalcon/rustybuzz

>My requirements are limited to simple Latin left-to-right text; harfbuzz is overkill.

The harfbuzz dependency is mandatory. I do not plan reducing text rendering to Latin only, since this is against the SVG spec.

Anyway, Rust version will be way smaller.

>like constantly re-load fonts

In your code or in usvg? Because usvg does this too.

[...]

PR #334, opened 2020-12-05
Add features to build for web-sys. #229

dabreegster:
This is the other half of https://github.com/RazrFalcon/resvg/issues/229. A user can depend on usvg with `default-features = false, features = ["text", "text_web-sys"]` to enable this new behavior.

After the PR in fontdb is merged, I'll change this PR to pin to the latest git commit on the upstream fontdb repo, and regenerate Cargo.lock for usvg.

I tested both of these PRs by pinning A/B Street web to my local copies.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,4 +33,4 @@ rayon = "1.3"
 
 [features]
 default = ["text"]
-text = ["usvg/text"] # enables SVG Text support
+text = ["usvg/system-fonts", "usvg/text"] # enables SVG Text support
--- a/usvg/Cargo.toml
+++ b/usvg/Cargo.toml
@@ -18,7 +18,7 @@ travis-ci = { repository = "RazrFalcon/resvg" }
 
 [[bin]]
 name = "usvg"
-required-features = ["text"]
+required-features = ["system-fonts", "text"]
 
 [dependencies]
 base64 = "0.13"
@@ -37,7 +37,7 @@ siphasher = "0.2"
 svgtypes = "0.5"
 
 # for text to path
-fontdb = { version = "0.4", optional = true }
+fontdb = { version = "0.4", optional = true, default-features = false }
 rustybuzz = { version = "0.3", optional = true }
 memmap2 = { version = "0.1", optional = true }
 ttf-parser = { version = "0.9", optional = true }
@@ -46,13 +46,16 @@ unicode-script = { version = "0.5", optional = true }
 unicode-vo = { version = "0.1", optional = true }
 
 [features]
-default = ["text"]
+default = ["system-fonts", "text"]
 text = [
     "fontdb",
-    "memmap2",
     "rustybuzz",
     "ttf-parser",
     "unicode-bidi",
     "unicode-script",
     "unicode-vo",
+]
+system-fonts = [
+    "fontdb/fs",
+    "memmap2",
 ]
--- a/usvg/src/lib.rs
+++ b/usvg/src/lib.rs
@@ -244,7 +244,7 @@ pub trait SystemFontDB {
     fn load_system_fonts(&mut self);
 }
 
-#[cfg(feature = "text")]
+#[cfg(feature = "system-fonts")]
 impl SystemFontDB for fontdb::Database {
     #[inline]
     fn load_system_fonts(&mut self) {


RazrFalcon:
The `text_native` name is a bit confusing, mainly because it clashes with [SVG Native](https://svgwg.org/specs/svg-native/). I would prefer something like `system-fonts`. Especially since I want to move system fonts scanning to fontdb eventually.

dabreegster:
Good point -- renamed to `system-fonts`. Not sure if it's misleading that `memmap2` is included in that feature.

RazrFalcon:
No, I think `memmap2` is fine, since it used only for system fonts.

dabreegster:
Not sure it's a meaningful diff, but https://github.com/dabreegster/abstreet/commit/b6108859a78fe4f4e0d51661a5dd8a5ab5fc16fb is the demonstration of this working (well, you'd have to run it locally I guess). Once https://github.com/RazrFalcon/fontdb/pull/6 is merged, this PR can pin to the upstream and should also be ready to go.

RazrFalcon:
Ufff... wasm-bindgen is a fat dependency. I'm trying to avoid proc macros as much as possible. Is there a way we can bypass it?

dabreegster:
> Is there a way we can bypass it?

It shouldn't get built at all unless somebody opts into the feature, right? And anybody doing that is almost certainly depending on wasm-bindgen anyway, since they're building for the web.

> Is it just for uuid? UUIDv4 should be like 10LOC...

Yes, just for fontdb. It looks like it needs it for an RNG: https://github.com/uuid-rs/uuid/blob/379e3e428b64ff5a3a78913b4646329053186c12/Cargo.toml#L99

RazrFalcon:
It would be included in vendored dependencies, which is important for me.

I should decide what to do with uuid/rand first. I basically need a bloat-less uuid generator.

dabreegster:
> I basically need a bloat-less uuid generator.

I've only taken a quick glance at fontdb, but could you do something kind of awful and just have a global mutable ID counter that's incremented every time `parse_face_info` is called? Unless you expect a single program to call it 2^32 or 2^64 times, it should be OK?

But actually since fonts are re-parsed, maybe this method is actually called frequently and hitting that limit is a risk?

RazrFalcon:
I need UUID, because you can remove fonts/faces from the DB, which means a global counter/index is no go.

RazrFalcon:
@dabreegster What web target do you use? I think it would be fine to use something like `oorandom` instead of `uuid`. It's has no dependencies and no_std by default. It's not cryptographic, but it doesn't matter in our case. So if `oorandom` would compile/work on on your target - we can switch to it.

RazrFalcon:
I never worked with random data and it looks pretty absurd. We need a random seed, but `std::time::Instant` doesn't work in WASM. So you need something like `instant`, which also depends on `wasm-bindgen`... I think the easiest way is to require a seed in the `fontdb::Database::new`, so it's a caller problem and not our.

By the way, the reason why `fontdb` uses uuid in the first place is because it's not a global database, so you can have multiple databases at once. Which means that you can use ID of one db in another one, which will return an invalid font.
Idk, maybe we can ignore this "issue" completely, so a caller will be responsible in not messing up the ids.

dabreegster:
> What web target do you use? 

web-sys. oorandom will work there.

> I think the easiest way is to require a seed in the fontdb::Database::new, so it's a caller problem and not our.

This works for my use case, and I can't imagine it causing problems for anybody else -- either they can use `std::time::Instant` on native, or they're on web and are forced to pull in wasm-bindgen anyway and can use `instant`.

> Idk, maybe we can ignore this "issue" completely, so a caller will be responsible in not messing up the ids.

Sounds reasonable to me. I think I've seen a fancier approach that adds an anonymous `PhantomData` type parameter to the object handing out IDs, retains that parameter on the IDs, and gets the compiler to check that the correct ones are later passed in. I can't find the article now, or how to get the compiler to create anonymous, unique types like this. I think this kind of thing would be complete overkill here and make the API very unergonomic.

I can make the changes to fontdb and resvg if you like, but not until later tonight or tomorrow. Thanks for digging into the dependency issue and finding a lightweight fix.

RazrFalcon:
The trick with `PantomData` works only when we have different types. And I'm not sure how we can create random types dynamically.

I will remove uuid from fontdb completely then and bump a version. Looks like it creates more complexity than solves.

RazrFalcon:
Done.

dabreegster:
Awesome, just a simple counter per font DB. Thanks for making that fix! I've updated this PR and re-tested with A/B St; this should be good to go.

PR #335, opened 2020-12-11
Add more common types of 'FeBlendMode'

suti:
add some common types of 'FeBlendMode', which is commonly used in CSS. 
refer to MDN documentation [blend-mode](https://developer.mozilla.org/zh-CN/docs/Web/CSS/blend-mode)


RazrFalcon:
Those modes are not part of the [SVG spec](https://www.w3.org/TR/SVG11/filters.html#feBlendModeAttribute).

PR #351, opened 2021-01-30
Make svgfilters crate no_std.

upsuper:


RazrFalcon:
Can you explain the reason to do so?

upsuper:
Well, just because it can :smile: It will allow this crate to be used where no_std is desirable, like embedded / wasm.

svgfilters/src/turbulence.rs
```diff
@@ -2,6 +2,9 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+use alloc::vec;
```
RazrFalcon, line 5: This line is redundant.

svgfilters/src/morphology.rs
```diff
@@ -2,6 +2,8 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+use alloc::vec;
```
RazrFalcon, line 5: Is it needed here?

svgfilters/src/convolve_matrix.rs
```diff
@@ -2,6 +2,8 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+use alloc::vec;
```
RazrFalcon, line 5: Is it needed here?

svgfilters/src/iir_blur.rs
```diff
@@ -26,6 +26,8 @@
 
 // TODO: Blurs right and bottom sides twice for some reason.
 
+use alloc::vec;
+use alloc::vec::Vec;
```
RazrFalcon, line 30: Is it needed here?

RazrFalcon (changes requested):

svgfilters/src/turbulence.rs
upsuper: It's not. If you remove this line it wouldn't compile, because with `#[no_std]`, the std prelude is not imported automatically, and thus there is no `vec` macro in the namespace.

RazrFalcon: Shouldn't it be marked as `#[macro_use]`?

RazrFalcon:
Thanks. I've merged it as is.

svgfilters/src/turbulence.rs
upsuper: Marking `extern crate alloc;` with `#[macro_use]` would work, yes, but I think with 2018 edition, it's generally recommended to import them individually rather than using `#[macro_use]`.

PR #401, opened 2021-06-26
fix some clippy warnings

remilauzier:


RazrFalcon:
If you plan more "fixes" - send them as a one patch, not 10.

remilauzier:
That's the last one and my brain can't manage everything when there's mores then hundred of warnings to fix. And i can't know what is an acceptable warning when there no configuration of clippy prior to that.

RazrFalcon:
You can use separate commits or a draft pull request for this. Or at least note that you plan to do more work.

PR #407, opened 2021-07-04
usvg: update rustybuzz dependency to 0.4

bpowers:
this is a trivial PR, but ensures we don't pull in 2 different versions of rustybuzz


RazrFalcon:
Thanks, but I'm usually updating dependencies right before the release. And I don't plan any.
Or do you plan to use master version of resvg?

PR #439, opened 2021-08-29
Forward memmap-fonts fontdb feature in cargo config

neinseg:
Companion PR for https://github.com/RazrFalcon/fontdb/pull/17

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,3 +32,4 @@ once_cell = "1.5"
 default = ["text", "system-fonts"]
 text = ["usvg/text"] # enables SVG Text support
 system-fonts = ["usvg/system-fonts"] # enables system fonts loading
+memmap = ["usvg/memmap"] # allows font files memory mapping, greatly improves performance
--- a/usvg/Cargo.toml
+++ b/usvg/Cargo.toml
@@ -54,5 +54,10 @@ text = [
 ]
 system-fonts = [
     "fontdb/fs",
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -32,3 +32,7 @@ once_cell = "1.5"
 default = ["text", "system-fonts"]
 text = ["usvg/text"] # enables SVG Text support
 system-fonts = ["usvg/system-fonts"] # enables system fonts loading
+memmap = ["usvg/memmap"] # enables font loading using memmap
+
+[patch.crates-io]
+fontdb = { path = "fontdb" }
```
RazrFalcon, line 38: You should simply replace

```toml
fontdb = { version = "0.6", optional = true, default-features = false }
```

with 

```toml
fontdb = { git = "https://github.com/RazrFalcon/fontdb", rev = "b009936", optional = true, default-features = false }
```

in `usvg/Cargo.toml`. Replace `b009936` with a hash we would get after merging your patch.

`[patch.crates-io]` section should be removed completely.

RazrFalcon (changes requested):

Cargo.toml
neinseg: Oops, that slipped past me.

RazrFalcon:
Done via https://github.com/RazrFalcon/resvg/commit/30bc5e1148402c9ec0ffcfe70a7edd77220bb4b8

Issue #40, opened 2018-02-14
Detect SVG size by bboxes


RazrFalcon:
Currently, we skip files which doesn't have `width`, `height` or `viewBox` attribute.

reknih:
Hey,
I ran across this issue when trying to use usvg to render some SVG glyphs in fonts. [Twitter Color Emoji](https://github.com/eosrei/twemoji-color-font) for example has stripped the `viewBox` from their files in the actual font file (they did not contain `width` or `height` attributes in the first place). <small>[SVGO also has the tendency to remove `viewBox`es and to cause trouble with that..](https://github.com/svg/svgo/issues/1128)</small>

I'd like to submit a PR allowing to convert most files missing the attributes to Micro SVG, but your input is needed on a few points:

- For SVGs that only use absolute units as lengths in paths etc. we can just assume an infinite user coordinate space and shrink that down using bboxes afterwards, no problems. However, when lengths do use percentage values (and `width` and `height` are defined in terms of viewport size), there is no single Micro SVG presentation that preserves the file's rescaling behavior. I thought of two possibilities of dealing with this:
  - Reject files that do not have the `width`, `height` attributes defined in absolute terms, do not specify a `viewBox` _and_ use percentage lengths.
  - Allow the user to specify a fallback size for the viewport in the options struct with respect to which the `width` and `height` attributes get resolved. If such a fallback size is specified, usvg does not fail even if there are percentage lengths. Since in some applications, viewport size is not known beforehand, this preference would be optional to set. If it remains unset, the behavior would be the same as in the first possibility.
- Both of these would likely mean that I'd need to add a `bool` field to the state that indicates that the user space coordinate boundries were not unambigously resolved (alternatively, all such sizes could become `Option`s) and that the `convert_children` function and its descendants would start to return `Option`s / `Result`s.

Which of the alternatives would you prefer? Are you happy with this proposed design? \
Cheers

RazrFalcon:
Yes, the main problem with this issue is that the SVG spec doesn't explain how it should be handled (afaik).
We can simply fit the viewBox into content bbox, but this would not work for documents with relative lengths.
I agree that the simplest solution is to require a fallback viewBox/viewport size. And a caller would be able to change it to whatever size needed.

We could potentially ignore files with % coordinates, but `<svg><rect width="100%" height="100%"/></svg>` is a perfectly valid SVG.

[...]

PR #444, opened 2021-08-31
Allow usvg to process files with no viewBox and relative root dimensions.

reknih:
As discussed in #40 today, here is a PR that allows files without a `viewBox` and with relative `width`s or `height`s to be converted into the Micro SVG format.

For this, the commit introduces the new `ViewportIntent` enumeration with the variants `Unknown` (default) and `PresetSize`.

- `Unknown` can be used for documents without relative units to fit a `viewBox` as tight around the path bounding boxes as possible
- `PresetSize` can also process relative units, but has to know the viewport size the SVG will be rendered into.

### DIFF
--- a/usvg/src/main.rs
+++ b/usvg/src/main.rs
@@ -113,6 +113,8 @@ struct Args {
     font_dirs: Vec<PathBuf>,
     skip_system_fonts: bool,
     list_fonts: bool,
+    default_width: u32,
+    default_height: u32,
 
     keep_named_groups: bool,
     id_prefix: Option<String>,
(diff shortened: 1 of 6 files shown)


RazrFalcon:
I think the current implementation is too strict. Instead of using `ViewportIntent`, we can simply add `default_size: Size` to `Options`, which is set to 100x100 by default. This is a reasonable fallback and will simplify code/error processing quite a lot.

RazrFalcon:
Also, I think we can reuse existing `--width` and `--height` arguments. Not sure if anyone would want to use different target size and viewport size.

reknih:
Alright, I force-pushed a (way smaller) patch in line with your suggestions onto here.

For archival purposes, the first patch version is preserved on the `try-1` branch on my fork.

usvg/src/main.rs
```diff
@@ -158,6 +160,8 @@ fn collect_args() -> Result<Args, pico_args::Error> {
         font_dirs:          input.values_from_str("--use-fonts-dir")?,
         skip_system_fonts:  input.contains("--skip-system-fonts"),
         list_fonts:         input.contains("--list-fonts"),
+        default_width:      input.opt_value_from_fn("--default-width", parse_length)?.unwrap_or(100),
```
RazrFalcon, line 163: Add to HELP.

RazrFalcon (changes requested):

RazrFalcon:
Looks good to me.

Don't worry about force-pushing, I will squash it anyway.

usvg/src/main.rs
RazrFalcon: https://github.com/RazrFalcon/resvg/blob/2fba92c4c8ea1d463b25c7ab82c225dc13bb5a19/usvg/src/main.rs#L13

RazrFalcon:
Thanks again. I will publish a new version soon.

PR #453, opened 2021-09-10
Don't wrap long words early

romanzes:


romanzes:
Sorry, I keep opening PRs against the original repo instead of my fork

PR #458, opened 2021-09-15
Improve turbulence filter perf

akindle:
Resolves https://github.com/RazrFalcon/resvg/issues/457


RazrFalcon:
How the profiler output looks now? Are there other hot spots?

akindle:
It's still very much the filter:
image
but it's at least more distributed throughout the filter math:
image
image

RazrFalcon:
Yeah. I would say that for a 5000 × 3400 image it's an expected result.

PR #465, opened 2021-09-26
cargo: Remove unneeded required-features

neinseg:
Having all default features listed as `required-features` makes it impossible to build the binaries with them off. This PR removes all `required-features` except for usvg's `export` feature from the binaries, as the binaries work fine without them. This is necessary for wasm builds, as wasm does not support the memmap-fonts feature.

Ideally I think this should be done by having these features default to "on", with an override to turn them off--but AFAICT cargo does not support such "negative" features.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,6 @@ members = ["c-api", "svgfilters", "usvg"]
 
 [[bin]]
 name = "resvg"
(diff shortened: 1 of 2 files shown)


RazrFalcon:
This is intentional. Why do you want resvg binary built for wasm? You should use the lib instead.

PR #472, opened 2021-10-20
DRAFT: Non-panicking C API

agerasev:
Hi!

I'm using resvg and it's really impressing. Thank you for your hard work!

But I've faced an issue with c-api. The issue is that C API functions can panic internally due to variety of reasons and therefore terminate the whole application. And I can't see any good way to handle these panics in C code.

Have you considered the possibility of making another C API that reports about all errors by returning an error code rather than panicking? (maybe along with current C API)

Here is my vision of such C API. Could you tell me what do you think about that? Has it a chance to be added to resvg?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -143,6 +143,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "macro_rules_attribute"
+version = "0.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "641c64af6cd80b81cf9c2f2f6ee382b1050c71ce63e20800499971a4a4195005"
+dependencies = [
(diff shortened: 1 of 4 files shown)


agerasev:
Ok, now I understand that resvg can only panic on breach of API contract. It was difficult to see at first.
Although I would prefer error code rather than panicking, it seems reasonable here because it makes the code much simpler.
Thank you for taking a look at the PR, I'm closing it.

bravikov:
I think it is a very useful pull request.

The library reacts too brutally to various mistakes.

RazrFalcon:
The library itself is very forgiving. C API is not. `assert` on contact violation is a normal practice for C code.

Can you provide some concrete examples when the C API behaved in an unexpected way?

PR #483, opened 2021-11-18
Accept a user-specified transform when rendering a tree.

iamralpht:


iamralpht:
I'm not sure if there's a better way of doing this (maybe by inserting a Group node and adding a transform?). I'm using it to translate a document which has content that goes beyond (0, 0) so that nothing is clipped by the origin, and to apply different scale factors to the document (targeting various display densities).

RazrFalcon:
I'm not sure about this one. `transform` should be provided to both `render` and `render_node`. We don't need a separate method just for this.

iamralpht:
Ok, I've updated the change so that `transform` is provided to `render` and `render_node`. We could also do a `RenderOpts` struct, then we could add other fields to it over time and provide a `Default` implementation...

RazrFalcon:
Yes, a separate struct might me useful in the feature.

PR #484, opened 2021-11-24
Export user-specified transform in C API.

jpap:


jpap:
I needed this for a SVG having canvas size `C` and RESVG_FIT_TO_ZOOM with a scaling factor `S`.  When rendered into a pixmap of size smaller than `C*S`, I had expected the render to be centered in that pixmap: instead it was just rendered to origin in the top-left.

By using a translating transform, I can get the SVG to be scaled up and centered in the above pixmap.

c-api/lib.rs
```diff
@@ -592,14 +604,24 @@ pub extern "C" fn resvg_render(
     let pixmap: &mut [u8] = unsafe { std::slice::from_raw_parts_mut(pixmap as *mut u8, pixmap_len) };
     let pixmap = tiny_skia::PixmapMut::from_bytes(pixmap, width, height).unwrap();
 
-    resvg::render(&tree.0, fit_to.to_usvg(), tiny_skia::Transform::default(), pixmap).unwrap()
+    let skia_xf = tiny_skia::Transform::from_row(
```
RazrFalcon, line 607: Prefer:

```rust
impl resvg_transform {
    #[inline]
    fn to_tiny_skia(&self) -> tiny_skia::Transform {
        tiny_skia::Transform::from_row(
            self.a as f32, 
            self.b as f32, 
            self.c as f32, 
            self.d as f32, 
            self.e as f32, 
            self.f as f32,
        );
    }
}
```

and then `transform.to_to_tiny_skia()`, just like with `resvg_fit_to`

c-api/resvg.h
```diff
@@ -166,6 +166,11 @@ typedef struct resvg_transform {
     double f;
 } resvg_transform;
 
+/**
+ * @brief Returns an identity transform.
+ */
+resvg_transform resvg_transform_default();
```
RazrFalcon, line 172: I think we should name it `resvg_transform_identity` to be more clear.

c-api/ResvgQt.h
```diff
@@ -530,6 +530,7 @@ class ResvgRenderer {
      */
     QImage renderToImage(const QSize &size = QSize()) const
     {
+        resvg_transform xf = resvg_transform_default();
```
RazrFalcon, line 533: This function should accept `QTransform` then, but I guess I can add this myself.
At least rename `xf` to `ts`.

RazrFalcon (changes requested):

c-api/resvg.h
jpap: I have made the requested change, but the the "default" name was chosen to mirror the same name you chose in tiny_skia.  I don't mind either way...

c-api/ResvgQt.h
jpap: Please go ahead and refactor to incorporate `QTransform`.

c-api/ResvgQt.h
```diff
@@ -441,15 +441,15 @@ class ResvgRenderer {
      */
     QRect viewBox() const
     {
-        return viewBoxF().toRect();
+        return viewBots().toRect();
```
RazrFalcon, line 444: ?!

RazrFalcon (changes requested):

c-api/ResvgQt.h
jpap: Thanks for catching this: was a slip in a case-insensitive search and replace.

RazrFalcon (approved):

RazrFalcon:
Thanks!

PR #488, opened 2021-12-01
Exclude test data from the crates.io tarball

Shnatsel:
Reduces the crate size on crates.io from 7Mb to less than 100Kb.

This matters to people on slow and flaky Internet connections, e.g. those living in China.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ license = "MPL-2.0"
 edition = "2018"
 description = "An SVG rendering library."
 repository = "https://github.com/RazrFalcon/resvg"
+exclude = ["tests"]
 
 [workspace]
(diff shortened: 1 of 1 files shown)


RazrFalcon:
Is this a common practice for crates.io? Excluding tests seems like a bad idea.

Shnatsel:
AFAIK, crates are downloaded from crates.io only when they are needed as a dependency of something else that you want to compile. In that case the resvg tests will never be run anyway. Ditto for `cargo install resvg`.

In fact, there is no built-in way to download the source code from crates.io via Cargo; that requires a [third-party plugin](https://github.com/Xion/cargo-download).

RazrFalcon:
I see. Ok.

PR #498, opened 2022-01-25
Add possibility to render with custom xlink:href resolver function

antmelnyk:
This is a working attempt to add another `render` function that would use a custom `xlink:href` resolver, as discussed in https://github.com/RazrFalcon/resvg/issues/479.

In that scenario, a special `ImageKind::CustomHref` kind (as well as two other new kinds for `Arc` references to PNG and JPEG to avoid copying buffers) is used that doesn't hold image data and does nothing when SVG tree is generated.

But, during rendering with new function `render_with_resolver` inside `image::draw_kind` that href string is passed to a provided resolver function, and on "runtime" it gets image buffer as `Vec<u8>`. Then, the resulting buffer is passed again to `draw_kind` to determine the final `ImageKind` (ref to PNG or JPEG).

This opens possibilities to render SVG files with arbitrary data inside `xlink:href`. Great example is just to treat href as HTTP link and perform a HTTP request for the image:
```rust
let image_url_resolver = |href: &str| {
     reqwest::blocking::get(href)
         .and_then(|data| data.bytes())
         .map(|bytes| bytes.to_vec())
         .ok()
         .map(Arc::new)
};

resvg::render_with_resolver(
  &rtree,
  usvg::FitTo::Original,
  pixmap.as_mut(),
  Box::new(image_url_resolver),
).unwrap();
```

But it also can be used in more sophisticated ways, for example in my use-case, I already have image buffers in memory and I don't need `resvg` to read the file or use base64 string, I just use the resolver function to get images from HashMap, href acts as image id:
```rust
let link_resolver = |href: &str| {
  my_hash_map
    .get(href)
    .map(Arc::clone)
};
```

---------

Implementation-wise I tried to change as little API as possible, there is just 1 more function to render and 1 more option for Options. Internally there is a bunch of `link_resolver` passing around to make it possible to use the function on-render. All feedback is welcomed since I'm Rust newbie 🙏🏻 🦀 

----------

I haven't run regression tests yet, I'm not sure how to actually, so there can be issues I could accidentally miss, but I will of course fix them if we move on with PR.


antmelnyk:
Hey @RazrFalcon, hope you're doing good.

Any possibility of this being checked out from your side? I'd highly appreciate it, and I think that feature would be a great addition to the crate.

I'm open for discussion and potential fixes/improvement/implementation adjustments on the feature.

antmelnyk:
Thank you for your feedback! 🙏🏻 

Your idea is actually close to what I've tried to [implement initially!](https://github.com/RazrFalcon/resvg/pull/498/commits/679f36be89da29c229810eccbd60972d458b386a)! There were 2 concerns though:

1. Resolver ideally can't be just `fn(&str) -> ImageKind` function on trait because in that case, you cannot use anything else besides `href` to determine the image, e.g. example with existing buffers. That's why I use closure in PR, it opens more possibilities. What do you think about it? The type will be this one, as I use it in PR:
```rust
pub type ImageHrefResolver<'a> = Box<dyn Fn(&str) -> Option<Arc<Vec<u8>>> + 'a>;
```

We can also use struct and do this, following your idea:
```rust
struct ImageHrefResolver<'a> {
    pub resolve_data: Box<dyn Fn(mime: &str, data: &[u8]) -> Option<ImageKind> + 'a>,
    pub resolve_string: Box<dyn Fn(&str) -> Option<ImageKind> + 'a>
}
```
That would be the best of 2 ideas in my opinion. Yes, the type is a bit ugly, but it's more powerful.

2. `Options` are provided *once* when calling `usvg::Tree::from_*`, which means that if I need to change resolver, I would need to reassign it to `Options`, then create again `OptionsRef` from it to call `usvg::Tree::from_*` again with new resolver. Or that's not a problem at all? I imagine it's too edge case anyway for most of the users and even if they will use custom resolver, it's going to be the same. If it's fine to just reassign `opt.image_href_resolver`, then I don't need to touch `resvg`, and it indeed can be done on `usvg` level.

I think we're on the same page about how it should work, the only main question is - do you see it implemented only as Trait with some `fn`s or do you agree that closure would be better, as I've described above?

>Also, we probably should store images in usvg::ImageKind as Arc<Vec<u8>> and not just Vec<u8>.
That's true, I've added 2 new kinds that are `Arc`s and they're there alongside already existing kinds. Or do you mean that in addition to the `ImageHrefResolver` trait/struct we would replace old kinds with refs as well?

RazrFalcon:
Yes, you're right, I've completely forgot about the context. Yes, `ImageHrefResolver` should use closures. Of course, it would require some fighting with lifetimes, but I don't see an other way.

>Options are provided once when calling usvg::Tree::from_*, which means that if I need to change resolver,

This is not an issue.

>That's true, I've added 2 new kinds that are Arcs and they're there alongside already existing kinds.

Yes, we don't need non-Arc one anymore.

antmelnyk:
> Of course, it would require some fighting with lifetimes, but I don't see another way.

Yep, lifetimes are for sure required there and I spent some time figuring out how to make them work, but it's no big deal in the end. With `Box<T> + 'a` for closure and `Arc<T>` for images it works just fine.

>This is not an issue.

>Yes, we don't need non-Arc one anymore.

Noted. All clear for now, I'll try to implement it in that way, thanks.

usvg/src/image_href_resolver/default_resolver.rs
```diff
@@ -0,0 +1,140 @@
+//! Default resolver that `usvg` uses to handle `<image>` elements `xlink:href` strings.
```
RazrFalcon, line 1: I don't think we need a separate `image_href_resolver` module. All this code can leave in `image.rs` just fine.

usvg/src/image_href_resolver/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+#[derive(Clone, Debug)]
+pub struct ImageHrefResolver<'a> {
+    /// Resolver function that will be used if `xlink:href` is a DataUrl with encoded base64 string.
+    pub resolve_data: Box<dyn ImageHrefDataResolver + 'a>,
+    /// Resolver function that will be used to handle arbitrary string in `xlink:href`.
+    pub resolve_string: Box<dyn ImageHrefStringResolver + 'a>,
+}
+
+/// Image `href` resolver that accepts mime type string and reference to decoded base64 image data.
+pub trait ImageHrefDataResolver: Fn(&str, Arc<Vec<u8>>) -> Option<ImageKind> {
```
RazrFalcon, line 22: Why do we need this monstrosity? Can't we simply have:

```rust
pub resolve_data: Box<dyn Fn(&str, Arc<Vec<u8>>) -> Option<ImageKind> + 'a>,
```

usvg/src/options.rs
```diff
@@ -132,11 +137,16 @@ pub struct Options {
     /// Default: empty
     #[cfg(feature = "text")]
     pub fontdb: fontdb::Database,
+
+    /// Specifies the way to parse `<image>` elements `xlink:href` value.
+    ///
+    /// Default: either parse `href` as DataUrl, or parse it as a path to the local image file.
+    pub image_href_resolver: Option<ImageHrefResolver<'a>>,
```
RazrFalcon, line 144: I don't think resolver should be optional. It should be set to default by default. And default resolver creation should be cheap.

RazrFalcon (changes requested):

examples/custom_href_resolver.rs
```diff
@@ -0,0 +1,53 @@
...
+    opt.resources_dir = std::fs::canonicalize(&args[1])
+        .ok()
+        .and_then(|p| p.parent().map(|p| p.to_path_buf()));
+    opt.fontdb.load_system_fonts();
+
+    // We know that our SVG won't have DataUrl hrefs.
+    let resolve_data = Box::new(|_: &str, _: Arc<Vec<u8>>| None);
+
+    // Let's treat href as URL and download PNG image via network.
+    let resolve_string = Box::new(|href: &str| {
```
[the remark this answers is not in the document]
RazrFalcon: This way we can even remove CLI args, since everything would be hardcoded.

usvg/src/image_href_resolver/mod.rs
antmelnyk: All these shenanigans are to derive Debug and Clone traits, otherwise can't have closure as a part of `Options` struct.

RazrFalcon: I would rather remove Debug and Clone from options.

usvg/src/options.rs
antmelnyk: The reason I did it is that:

>Current resvg resolver logic requires Options itself. So if we use `custom_href_resolver` as `Options` field we need to capture `Options` itself and it cannot be referenced, has to be owned Options.

Since existing default resolver logic requires `Options` (because inside the resolver there is SVG case where we recursively call render), I can't create a default resolver without having `Options` struct first. That's why `use_default_href_resolver` function exists as a separate step. First we create Options with resolver as None, then use them to define resolver itself.

So far don't see another way around it...

usvg/src/image_href_resolver/mod.rs
antmelnyk: That works too 🙂

usvg/src/options.rs
```diff
@@ -53,7 +58,7 @@ impl FitTo {
 
 /// Processing options.
 #[derive(Clone, Debug)]
-pub struct Options {
+pub struct Options<'a> {
```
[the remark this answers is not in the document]
antmelnyk: I am thinking about it. That lifetime bugs me too (I've tried this PR with my codebase that uses `resvg` and it's pretty annoying to deal with in complicated code).

usvg/src/image_href_resolver/mod.rs
antmelnyk: Done.

examples/custom_href_resolver.rs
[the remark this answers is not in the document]
antmelnyk: Done, now example is self contained and doesn't use any arguments.

examples/custom_href_resolver.rs
```diff
@@ -0,0 +1,53 @@
...
+        reqwest::blocking::get(href)
+            .and_then(|data| data.bytes())
+            .map(|bytes| bytes.to_vec())
+            .ok()
+            .map(|data: Vec<u8>| ImageKind::PNG(Arc::new(data)))
+    });
+
+    // Assign new ImageHrefResolver option using our closures.
+    // Now `href` strings in `<image>` elements will be treated as URLs to PNG images.
+    opt.image_href_resolver = Some(ImageHrefResolver {
```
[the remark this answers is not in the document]
antmelnyk: @RazrFalcon Just pushed the new commi. Now everything is pretty simple inside as well I think. Let me know what you think.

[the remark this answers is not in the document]
antmelnyk: Also, here's an example of new API usage in my real project:

```rust
let existing_buffer_resolver =
    |href: &str, _: &OptionsRef| generated_images.get(href).map(Arc::clone);

self.dependencies
    .svg_rendering_options
    .image_href_resolver
    .resolve_string = Box::new(existing_buffer_resolver);
```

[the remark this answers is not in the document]
antmelnyk: Also, here's an example of real world API usage for my use case:

```rust
let existing_buffer_resolver = |href: &str, _: &OptionsRef| {
    generated_images
        .get(href)
        .map(|data| ImageKind::PNG(Arc::clone(data)))
};

self.dependencies
    .svg_rendering_options
    .image_href_resolver
    .resolve_string = Box::new(existing_buffer_resolver);
```

RazrFalcon (approved):

usvg/src/image.rs
```diff
@@ -37,6 +39,22 @@ impl std::fmt::Debug for ImageKind {
...
+    /// Resolver function that will be used if `xlink:href` is a DataUrl with encoded base64 string.
+    pub resolve_data: Box<dyn Fn(&str, Arc<Vec<u8>>, &OptionsRef) -> Option<ImageKind>>,
+
+    /// Resolver function that will be used to handle arbitrary string in `xlink:href`.
+    pub resolve_string: Box<dyn Fn(&str, &OptionsRef) -> Option<ImageKind>>,
+}
+
+impl std::fmt::Debug for ImageHrefResolver {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str("ImageHrefResolver closures (..)")
```
RazrFalcon, line 55: `ImageHrefResolver { .. }` is more [correct](https://doc.rust-lang.org/std/fmt/struct.DebugStruct.html#method.finish_non_exhaustive) (this method is available since 1.53, we cannot use it yet)

usvg/src/image.rs
```diff
@@ -95,10 +112,12 @@ pub(crate) fn convert(
         aspect: node.attribute(AId::PreserveAspectRatio).unwrap_or_default(),
     };
 
-    let href = node.attribute(AId::Href)
+    let href = node
+        .attribute(AId::Href)
```
RazrFalcon, line 116: Keep original formatting.

usvg/src/image.rs
```diff
@@ -111,34 +130,45 @@ pub(crate) fn convert(
...
+        (opt.image_href_resolver.resolve_data)(&mime, Arc::new(data), opt)
     } else {
-        let path = opt.get_abs_path(std::path::Path::new(href));
+        (opt.image_href_resolver.resolve_string)(href, opt)
+    }
+}
+
+/// Create DataUrl resolver function that handles standard mime types for JPEG, PNG and SVG.
+#[allow(clippy::type_complexity)]
+pub fn create_default_data_resolver<'a>() -> Box<dyn Fn(&str, Arc<Vec<u8>>, &OptionsRef) -> Option<ImageKind> + 'a> {
```
RazrFalcon, line 151: I would prefer those methods to be part of `ImageHrefResolver` as static one. So we would not clatter the global namespace.

And I would rename them to something like `ImageHrefResolver::default_data_resolver`. I don't think we need create/new prefix here.

usvg/src/options.rs
```diff
@@ -150,6 +157,10 @@ impl Default for Options {
             default_size: Size::new(100.0, 100.0).unwrap(),
             #[cfg(feature = "text")]
             fontdb: fontdb::Database::new(),
+            image_href_resolver: ImageHrefResolver {
```
RazrFalcon, line 160: I think we could even implement `Default` for `ImageHrefResolver`.

RazrFalcon (changes requested):

RazrFalcon:
Everything looks nice and surprisingly simple. I'm satisfied. I would love some more stylistic changes mentioned above, but the core logic looks just fine to me.

antmelnyk:
> Everything looks nice and surprisingly simple. I'm satisfied. I would love some more stylistic changes mentioned above, but the core logic looks just fine to me.

I'll do mentioned adjustments, no problem.

But what about failing checks and this? Adding `Send` to type didn't helps seems like. Maybe need to do `Arc<Mutex<ImageKind>>` or something.
>`(dyn for<'r, 's, 't0> Fn(&'r str, Arc<Vec<u8>>, &'s OptionsRef<'t0>) -> Option<usvg::ImageKind> + 'static) 
` cannot be sent between threads safely

RazrFalcon:
Yes, it's a strange one. Maybe it wants `Arc` for closures?

antmelnyk:
I added `Send` and `Sync` and it helped.
I think it's because `dyn Trait` thing. When we use `dyn`, we lose all automatic traits.

RazrFalcon (approved):

RazrFalcon:
Thanks! I'm glad we were able to fit it into < 200 LOC patch. A nice little change which should be very useful.

If everything will go as planned, I would publish a new version on the weekends.

PR #502, opened 2022-02-03
Treat empty svg as 300x150

### LINKED ISSUE #501: Panics on svg with no width, height, viewBox, or contents
Given this input:

```svg
<svg xmlns="http://www.w3.org/2000/svg" />
```

Panics here:

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/main.rs:567:78
```

The computed size is f64::MAX × f64::MAX, which is too big of a pixmap to create.

One of these alternatives would be preferable to panicking (which e.g. causes the Nodejs bindings to abort).

1. Returning an error
2. Producing a 300x150 transparent png (this is how ImageMagick and web browsers behave)
3. Producing a 100x100 transparent png (this is how resvg treats a 100% rect in an svg with no size)

jleedev:
Fixes #501

### DIFF
--- a/usvg/src/lib.rs
+++ b/usvg/src/lib.rs
@@ -801,7 +801,11 @@ fn calc_node_bbox(
                 }
             }
 
-            Some(bbox)
+            if bbox.x() == f64::MAX && bbox.y() == f64::MAX && bbox.width() == 1.0 && bbox.height() == 1.0 {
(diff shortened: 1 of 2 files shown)


RazrFalcon:
Sorry, but this is not a solution.

PR #509, opened 2022-04-05
fix spurious warning when using --export-id

benoit-pierre:
Warn about `--export-area-drawing` when it's actually used in combination with `--export-id`, not the other way around.


RazrFalcon:
Thanks!

PR #517, opened 2022-06-03
Replace unwrap() on linked node lookup for pattern with check

FylmTM:
We have encountered an invalid SVG, on which `usvg` panicked.

Simple example:
```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1 1">
<pattern id="a">
        <rect fill="url(#b)" />
    </pattern>
</svg>
```

When child node of a pattern referenced non-existing URL, it hit code path that contained `.unwrap()`.
This change replaces that place with with `if let Some()` check, therefore now return None in such scenario.


FylmTM:
I should note that I am not sure whether this is _correct_ thing to do.
I guess `.unwrap()` was intentionally used there.

If there is more correct way to do it, so that panic could be avoid - I will be glad to look into that.

RazrFalcon:
Looks like a bug to me. Would you mind adding a test case? Just add `resvg/tests/svg/e-pattern-031.svg` and then render it using resvg and place it to `resvg/tests/png`

```xml
    <title>Child with invalid FuncIRI</title>

    <pattern id="patt1" patternUnits="userSpaceOnUse" width="20" height="20">
        <rect id="rect1" x="0" y="0" width="10" height="10" fill="url(#missing)"/>
    </pattern>
    <rect id="rect2" x="20" y="20" width="160" height="160" rx="20" ry="20"
          fill="url(#patt1)" stroke="darkblue"/>

    <!-- image frame -->
</svg>

```

FylmTM:
Done ✅

RazrFalcon:
Thanks!

PR #519, opened 2022-06-11
Path bbox calucalation should scale stroke width

growler:
Consider a path like

```
<path id="1" fill="none" stroke="#ffffff" stroke-width="4" transform="matrix(0.1 0 0 0.1 0 0)" d="M 0 0 L 0 20 L 20 20 L 20 0 Z"/>
```

A correct bbox for such a path would be something like -2 -2 24 24
scaled down to -0.2 -0.2 2.4 2.4

However, PathData::calc_bbox_with_transform uses original stroke width
added to a transformed path bbox and produces a wrong result of -2 -2 6 6

Scaling stroke width by \sqrt{|a*d - b*c|} helps.

### DIFF
--- a/usvg/src/pathdata.rs
+++ b/usvg/src/pathdata.rs
@@ -439,7 +439,11 @@ fn calc_bbox_with_transform(
     // TODO: find a better way
     // It's an approximation, but it's better than nothing.
     if let Some(stroke) = stroke {
-        let w = stroke.width.value() / 2.0;
+        let w = stroke.width.value() / if ts.is_default() {
+            2.0
+        } else {
+            2.0 / (ts.a * ts.d - ts.b * ts.c).abs().sqrt()
+        };
         minx -= w;
         miny -= w;
         maxx += w;


RazrFalcon:
Yes, makes sense.  But is this a correct formulae? We have `usvg::Transform::get_scale` and it looks like this:

```
let x_scale = (self.a * self.a + self.c * self.c).sqrt();
let y_scale = (self.b * self.b + self.d * self.d).sqrt();
```

growler:
Well, determinant of a 2d transformation matrix gives an area scaling factor, so sqrt(det(M)) results in a "visual zoom factor" for a stroke width when you'd prefer to keep the shape of the stroke itself unchanged, which is often the desired result when scaling a figure (I found a good visual example [here](https://tex.stackexchange.com/a/474035))

I think though that the exact formula should match the way resvg scales stroke while producing bitmap. If resvg transforms strokes' shapes, the bounding box calculation should reflect that.

For my purpose (a one-off tool coverting a bunch of svg icons into pgf) either way would fit. 

Do I get it right that the correct way would be 
```
    // TODO: find a better way
    // It's an approximation, but it's better than nothing.
    if let Some(stroke) = stroke {
        let w = stroke.width.value();
        let (x_scale, y_scale) = ts.get_scale();
        minx -= w * x_scale;
        miny -= w * y_scale;
        maxx += w * x_scale;
        maxy += w * y_scale;
    }
```
?

RazrFalcon:
I just thought that it would be better to use the same logic as `tiny-skia` uses during rendering. And it seems like it does the same as your code: https://github.com/RazrFalcon/tiny-skia/blob/3b5ec3c8de1ea16fadaf94f1099d01eeda172ea1/src/stroker.rs#L279 So it's fine.

`tiny-skia` is actually strokes the original path and then scales it. So there is not "stroke scaling" happening, afaiu. 
Anyway, it should be fine.

PR #523, opened 2022-07-14
feat: implement the writing-mode `horizontal-tb/vertical-lr` values

yisibl:
This is the standard value of the `writing-mode` property in [CSS Writing Modes Level 3](https://www.w3.org/TR/css-writing-modes-3/#svg-writing-mode-css), and we should be compatible with it.

> Name:	writing-mode
Value:	horizontal-tb | vertical-rl | vertical-lr
Initial:	horizontal-tb

Also there is a `vertical-lr` value in CSS3, how should it be handled in resvg?

CSS spec says:

> The SVG1.1 values were also present in an older of the CSS [writing-mode](https://www.w3.org/TR/css-writing-modes-3/#propdef-writing-mode) specification, which is obsoleted by this specification. The additional tb-lr value of that revision is replaced by [vertical-lr](https://www.w3.org/TR/css-writing-modes-3/#valdef-writing-mode-vertical-lr).


usvg/src/text/convert.rs
```diff
@@ -782,11 +786,13 @@ fn count_chars(node: svgtree::Node) -> usize {
...
 /// it should affect the rendering.
 ///
-/// [SVG 2.0]: https://www.w3.org/TR/SVG2/text.html#WritingModeProperty
+/// [SVG 2]: https://www.w3.org/TR/SVG2/text.html#WritingModeProperty
+/// [CSS Writing Modes Level 3]: https://www.w3.org/TR/css-writing-modes-3/#svg-writing-mode-css
 pub fn convert_writing_mode(text_node: TextNode) -> WritingMode {
     if let Some(n) = text_node.find_node_with_attribute(AId::WritingMode) {
         match n.attribute(AId::WritingMode).unwrap_or("lr-tb") {
-            "tb" | "tb-rl" => WritingMode::TopToBottom,
+            "horizontal-tb" => WritingMode::LeftToRight,
```
yisibl, line 794: Maybe we don't need this line?

docs/svg2-changelog.md
```diff
@@ -146,7 +146,7 @@ NOTE: This list is not final. This just things I was able to find so far. Patche
 
 - [x] [`textPath`](https://www.w3.org/TR/SVG2/text.html#TextPathElement) can reference [basic shapes](https://www.w3.org/TR/SVG2/shapes.html) now.
 - [ ] Since CSS Fonts Module Level 4, the [`font-weight`](https://www.w3.org/TR/css-fonts-4/#font-weight-prop) property allows any value in a 1..1000 range.
-- [ ] A [`writing-mode`](https://www.w3.org/TR/SVG2/text.html#WritingModeProperty) property has a new set of allowed values.
+- [x] The [`writing-mode`](https://www.w3.org/TR/SVG2/text.html#WritingModeProperty) property introduces the `horizontal-tb` and `vertical-lr` values from [CSS Writing Modes Level 3](https://www.w3.org/TR/css-writing-modes-3/#svg-writing-mode-css).
```
RazrFalcon, line 149: Keep `[x] A`, since this is the article I use everywhere (not sure correctly or not).

usvg/src/text/convert.rs
RazrFalcon, line 794: This branch is unnecessary.

RazrFalcon (changes requested):

RazrFalcon:
>Also there is a vertical-lr value in CSS3, how should it be handled in resvg?

As for now, resvg text layout is line based. There are not paragraphs. Therefore `vertical-lr` and `vertical-rl` act the same.

docs/svg2-changelog.md
yisibl: Done

usvg/src/text/convert.rs
yisibl: Done

RazrFalcon:
>So do we need to map vertical-lr to WritingMode::TopToBottom?

Yes. For now they will act the same.

RazrFalcon (approved):

yisibl:
Okay, hold on a second, let me add that.

Issue #422, opened 2021-07-11
Support `paint-order`


RazrFalcon:
https://www.w3.org/TR/SVG2/painting.html#PaintOrder

RazrFalcon:
Continues in #550

PR #525, opened 2022-07-19
feat: support paint order

tyutjohn:
This modifies desire to paint strokes below fills without having to duplicate an element.
[paint-order](https://www.w3.org/TR/SVG2/painting.html#PaintOrder) is a standard specification of svg2 as discussed [#422 ](https://github.com/RazrFalcon/resvg/issues/422)


yisibl:
Nice works!

I suggest adding more tests, such as.

 ```
paint-order: markers stroke;
paint-order: stroke fill;
```

RazrFalcon:
Thanks for the effort, but this implementation is very incomplete. First, `paint-order` is a list of 3 items. Second, you do not handle `markers` at all. I cannot accept it in the current state.
Would be willing to improve it? Correct markers order might be hard to implement. If this issue is critical for you, I might take a look into in on the weekends.

Also, because we resolve markers during the preprocessing stage, during rendering we do have only Fill and Stroke. Meaning we would need two `PaintOrder` types: internal and public.

tyutjohn:
> 



> during

My fault. At first I wanted to add the font's drawing rules, markers didn't seem to work, so I ignored it. Secondly in the rendering code I didn't find any handling of markers, so I chose to ignore this property. 
I will take some times to improve it, I'm not sure if it's a good idea to provide a function on marker rendering during the rendering

RazrFalcon:
Yes, markers are resolved during preprocessing in `usvg`. `paint-order` is not as simple feature as it looks like. As everything in SVG.

PR #527, opened 2022-07-24
Bump tiny-skia to 0.7.0

chrisduerr:
You mention that dependency update PRs will not be accepted, so feel free to just close this and do it yourself. However considering that `resvg` doesn't re-export the `tiny-skia` dependency while relying on its types in its public API, I think it would be a bad idea to keep the versions out of sync.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -339,9 +339,9 @@ dependencies = [
 
 [[package]]
 name = "tiny-skia"
-version = "0.6.5"
+version = "0.7.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "78b3e1db967020dd509b49cecc024025beba2b1b6cf204618610ba266269d6b9"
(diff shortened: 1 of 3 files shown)


RazrFalcon:
Well, it's not an issue unless you're using the master branch. 

And the next resvg version will reexport its dependencies.

chrisduerr:
> Well, it's not an issue unless you're using the master branch.

This is an issue I ran into while using the latest stable release.

RazrFalcon:
Well, you're suppose to use the crate version as resvg. This will be "fixed" by reexporting in the next release.

chrisduerr:
> Well, you're suppose to use the crate version as resvg. This will be "fixed" by reexporting in the next release.

I am and the latest version on crates.io will not work if you're also pulling the latest version of `tiny-skia` from crates.io. It shouldn't be a surprise really, sure it might be fixed in the next release that's great, but the status quo will have people running into this issue until the next resvg release.

Issue #371, opened 2021-04-06
slow on large dimensions


JoKalliauer:
resvg is extremly fast for simple files and small dimensions (compared to others), however it has more extremely slow renderings compared to others, which ruins the average time of renderings.

Inkscape has a bad performance for small files (e.g. resvg-test-suite), however for large files it is often the fastest.

Batik seems to use ~2CPUs, resvg is in my experience limited to one CPU, librsvg/Inkscape has i guess some, but hardly parallel-processing

## slow-file for 2.636 × 4.235
![](https://upload.wikimedia.org/wikipedia/commons/archive/9/91/20210406204111%21AntigenicShift_HiRes.svg)

| |resvg|librsvg|inkscape|batik (uses ~2CPUs)|
|---|---|---|---|---|
|real time|2m6,226s|0m11,346s|0m4,546s|0m8,992s

## 373 × 600
![](https://upload.wikimedia.org/wikipedia/commons/9/91/AntigenicShift_HiRes.svg)

| |resvg|librsvg|inkscape|batik (uses ~2CPUs)|
|---|---|---|---|---|
|real time|0m1,727s|0m0,385s|0m1,463s|0m1,463s

## Copyright

source: https://commons.wikimedia.org/wiki/File:AntigenicShift_HiRes.svg

author: https://commons.wikimedia.org/wiki/User:Mouagip

license: Public Domain

RazrFalcon:
resvg indeed has an inefficient large images handling, because of the way layers are handled (clipping is the most common one). But the provided image doesn't have layers, so this is a different issue. Will investigate.

JoKalliauer:
PS: The 373x600 image was optimized, so it is not only a scaled down svg of the other.

RazrFalcon:
Thanks to 4db63614ae06c1d9598ceeafad7e967c04929d57 resvg is only 2x slower on my machine now.

JoKalliauer:
> Thanks to [4db6361](https://github.com/RazrFalcon/resvg/commit/4db63614ae06c1d9598ceeafad7e967c04929d57) resvg is only 2x slower on my machine now.

Thanks Much faster now:

https://upload.wikimedia.org/wikipedia/commons/archive/9/91/20210406204111%21AntigenicShift_HiRes.svg limited to one cpu: (tested on different computer)

|resvg|librsvg|inkscape|batik|
|---|---|---|---|
|18,392s|6,317s|2,654s|6,387s|

Issue #146, opened 2019-06-21
Use native clip on simple clipPath


RazrFalcon:
No need to use layers when clipPath has only one child. This should be much faster, especially with markers (which are clipped by default).

RazrFalcon:
Turns out that this is a pretty complex task. 

- A native clipping can be used only in a very specific cases:
  - `clipPath` with a single child.
  - `userSpaceOnUse`
  - No `clip-path` on a `clipPath` element.
  - No `clip-path` on `clipPath` children.
- We have to manage our own clip path stack, because cairo and Qt doesn't support this (raqote does).

RazrFalcon:
Not needed.

PR #530, opened 2022-08-06
Improve `trim_transparency` and PNG rasterization

shuding:
For PNG rasterization, one quick optimization is to do `multiply_alpha` in the same loop assigning each pixel. When the PNG dimension is large the time here is not ignorable.

For `trim_transparency`, with the current solution we always loop over all pixels to find the boundary, so doing it n times will result in a time complexity of `O(n*WH)`.

This PR changes it to find the first non-transparent pixel starting from each edge. For example we loop over the (1) area until reaching a pixel. And then do the same for (2) (3) (4):

image

Still, no pixel is scanned twice but we are now skipping all the pixels inside the white area. In the worse case (transparent canvas) this algorithm has the same performance as our current one. But in average case this is a lot faster as the complexity becomes `O(n*(WH-s))` where s is the average non-transparent area size.

Since initially we always scan the pixels in the original data storage order, another optimization is to skip all-zero bytes quickly. That's because a transparent pixel is likely to have rgba(0, 0, 0, 0). Here I'm using `u128::from_ne_bytes((&pixels[idx..(idx + 16)]).try_into().unwrap())` to read 16 bytes which is 4 pixels at a time.

I tested the SVG in #371 (2636×4235), and it shows a pretty big improvement:

**Current master branch**

```
FontDB init: 61.98ms
Reading: 0.70ms
Parsing: 38.02ms
Rendering: 18958.88ms
Saving: 222.37ms
./target/release/resvg --perf test.svg test.png  17.42s user 1.81s system 95% cpu 20.207 total
```

**This PR**

```
FontDB init: 61.36ms
Reading: 0.59ms
Parsing: 33.39ms
Rendering: 8535.01ms
Saving: 232.73ms
./target/release/resvg --perf test.svg test.png  7.50s user 1.31s system 99% cpu 8.880 total
```

Which is 2.22x faster in _Rendering_ phase.

### DIFF
--- a/src/image.rs
+++ b/src/image.rs
@@ -117,25 +117,15 @@ fn image_to_pixmap(image: &Image, pixmap: &mut [u8]) {
         }
         ImageData::RGBA(data) => {
             for p in data.as_rgba() {
-                pixmap[i + 0] = p.r;
-                pixmap[i + 1] = p.g;
-                pixmap[i + 2] = p.b;
+                let a = p.a as f64 / 255.0;
+                pixmap[i + 0] = (p.r as f64 * a + 0.5) as u8;
+                pixmap[i + 1] = (p.g as f64 * a + 0.5) as u8;
+                pixmap[i + 2] = (p.b as f64 * a + 0.5) as u8;
                 pixmap[i + 3] = p.a;
 
                 i += tiny_skia::BYTES_PER_PIXEL;
             }
-
-            multiply_alpha(pixmap.as_rgba_mut());
         }
-    }
-}
-
-fn multiply_alpha(data: &mut [rgb::RGBA8]) {
-    for p in data {
-        let a = p.a as f64 / 255.0;
-        p.b = (p.b as f64 * a + 0.5) as u8;
-        p.g = (p.g as f64 * a + 0.5) as u8;
-        p.r = (p.r as f64 * a + 0.5) as u8;
     }
 }
 
--- a/src/render.rs
+++ b/src/render.rs
@@ -285,19 +285,19 @@ pub fn trim_transparency(pixmap: tiny_skia::Pixmap) -> Option<(i32, i32, tiny_sk
     let mut max_y = 0;
 
     let first_non_zero = {
-        let max_safe_index = pixels.len() >> 4;
+        let max_safe_index = pixels.len() / 8;
 
-        // Find first non-zero byte by looking at 16 bytes a time. If not found
+        // Find first non-zero byte by looking at 8 bytes a time. If not found
         // checking the remaining bytes. This is a lot faster than checking one
         // byte a time.
         (0..max_safe_index)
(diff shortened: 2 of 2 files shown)


RazrFalcon:
Thanks! Correctness and not performance is the current priority for me, therefore there are a lot of easy optimizations here and there. Clipping and text are the obvious one, but it would take a lot of time to implement. 

As for the patch, I'm a bit skeptical about the use of `u128`. Specifically from the portability standpoint. Will it work on WASM? Does it provide any benefits on 32bit systems? Wouldn't it be easier to use `u64`? Is the performance difference that big?

shuding:
Thanks for creating this library and taking a look the PR! It makes total sense that correctness is the most important thing. I'm digging into the code and trying to optimize it because I want to make it faster for my use case where both performance and size are critical. Noticed some easy optimizations and thought they're nice to have, as they don't affect the correctness regardless. :)

> As for the patch, I'm a bit skeptical about the use of u128.

Good points. I updated the PR to use `u64` instead, which isn't as fast as `u128` but still ~2x faster.

```
FontDB init: 67.42ms
Reading: 0.64ms
Parsing: 34.19ms
Rendering: 10685.33ms
Saving: 221.08ms
./target/release/resvg --perf test.svg test.png  9.11s user 1.85s system 99% cpu 11.030 total
```

RazrFalcon:
Thanks! Looks good.

shuding:
Kinda related, I know that you also have ideas like #146 planned, which is pretty complex and you are definitely busy. So I would like to know that are you open to accept performance optimization PRs like this one? And if so, do you have any suggestions on the direction (like SIMD which is more general)?

For example, I was also thinking about improving the `apply_drop_shadow` process. Right now both `box_blur` and `iir_blur` computes all RGBA channels but for the drop shadow use case, only alpha is needed. A quick test shows a 30% speed improvement for that filter, if we only compute blur for alpha channel.

RazrFalcon:
Clip path optimizations are endless...

1. Right now, our clip is raster and RGBA, but only Alpha is needed. Which means we should rewrite it to use `tiny_skia::ClipMask` at least. The problem is that `tiny_skia::ClipMask` is very limited at the moment and has to be updated accordingly. 
2. Neither `resvg` nor `tiny-skia` support Rectangle clip optimization, when we do not need a mask at all. This would require some, maybe even significant, changes to `tiny-skia`.
3. We could even port Skia's clipping algorithm, which doesn't use an alpha mask at all. Tons of work and code, but way faster.
4. Remember, that SVG clipping is not just a single path/layer. SVG clipPath can be nested. That's why we're using bitmap to begin with. So while we can use "pure vector clip path" optimization (will still be rasterized at some point) in most cases, there are still cases we're we would need to essentially blend clip paths together. 

As for filters, yeah, they are slow. Not sure if blur can be alpha-only thought.

PR #532, opened 2022-08-06
Make image decoding dependencies optional

### LINKED ISSUE #531: Make image decoders (gif, jpeg, png) optional
Would you accept a PR that makes these dependencies optional?

I understand that the SVG spec includes jpeg and png. If I know that my SVGs do not make use of these it would still be nice to eliminate those crates from the dependency graph.

e00E:
When building without default features or some combination of image decoding features there might be a couple of compilation warnings like `rendering_mode` being an unused variable. IMO it isn't worth it to pepper the code with more conditional compilation attributes to fix them.

fixes #531

Sidenote:
The PR template informs me that 

> Pull requests that include:
> - dependencies updates
> - code formatting fixes
> - clippy fixes
> - compiler warnings fixes
>
> will not be accepted.

Why is that? I find easier to work on projects that use rustfmt and that have less warnings. We don't need to debate it. I'm just curious why you decided on this. And do you not want dependency update PRs because you will do them yourself or because you don't want to update dependencies?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,7 @@ usvg = { path = "usvg", version = "0.23.0", default-features = false }
 once_cell = "1.5"
 
 [features]
-default = ["filter", "text", "system-fonts", "memmap-fonts", "gif", "jpeg-decoder", "png"]
+default = ["filter", "text", "system-fonts", "memmap-fonts", "image"]
 # enables SVG Filter support
 # adds around 100KiB to your binary
 filter = ["svgfilters", "usvg/filter"]
@@ -49,3 +49,5 @@ memmap-fonts = ["usvg/memmap-fonts"]
 # enables the --dump-svg flag in CLI
 # adds around 50KiB to your binary
 dump-svg = ["usvg/export"]
+# enables support for embedded images
+image = ["gif", "jpeg-decoder", "png"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,11 +17,14 @@ name = "resvg"
 required-features = ["filter", "text", "system-fonts", "memmap-fonts"]
 
 [dependencies]
-gif = "0.11"
-jpeg-decoder = { version = "0.2", default-features = false, features = ["platform_independent"] }
+# enables embedded GIF support
+gif = { version = "0.11", optional = true }
+# enables embedded JPEG support
+jpeg-decoder = { version = "0.2", default-features = false, features = ["platform_independent"], optional = true }
 log = "0.4"
 pico-args =  { version = "0.5", features = ["eq-separator"] }
-png = "0.17"
+# enables embedded PNG support
+png = { version = "0.17", optional = true }
 rgb = "0.8"
 svgfilters = { path = "svgfilters", version = "0.4", optional = true }
 svgtypes = "0.8"
@@ -32,7 +35,7 @@ usvg = { path = "usvg", version = "0.23.0", default-features = false }
 once_cell = "1.5"
 
 [features]
-default = ["filter", "text", "system-fonts", "memmap-fonts"]
+default = ["filter", "text", "system-fonts", "memmap-fonts", "gif", "jpeg-decoder", "png"]
 # enables SVG Filter support
 # adds around 100KiB to your binary
 filter = ["svgfilters", "usvg/filter"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,13 +17,10 @@ name = "resvg"
 required-features = ["filter", "text", "system-fonts", "memmap-fonts"]
 
 [dependencies]
-# enables embedded GIF support
 gif = { version = "0.11", optional = true }
-# enables embedded JPEG support
 jpeg-decoder = { version = "0.2", default-features = false, features = ["platform_independent"], optional = true }
 log = "0.4"
 pico-args =  { version = "0.5", features = ["eq-separator"] }
-# enables embedded PNG support
 png = { version = "0.17", optional = true }
 rgb = "0.8"
 svgfilters = { path = "svgfilters", version = "0.4", optional = true }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -32,7 +35,7 @@ usvg = { path = "usvg", version = "0.23.0", default-features = false }
 once_cell = "1.5"
 
 [features]
-default = ["filter", "text", "system-fonts", "memmap-fonts"]
+default = ["filter", "text", "system-fonts", "memmap-fonts", "gif", "jpeg-decoder", "png"]
```
RazrFalcon, line 38: I don't think we need that level of control. The `image` feature would be enough.

src/image.rs
```diff
@@ -20,30 +20,46 @@ pub fn draw_kind(
     canvas: &mut Canvas,
 ) {
     match kind {
+        #[cfg(any(feature = "jpeg-decoder"))]
```
RazrFalcon, line 23: All those `ImageKind`s should be feature-gated on the `usvg` side.

RazrFalcon (changes requested):

RazrFalcon:
>Why is that? I find easier to work on projects that use rustfmt and that have less warnings. We don't need to debate it. I'm just curious why you decided on this. And do you not want dependency update PRs because you will do them yourself or because you don't want to update dependencies?

Because it's my project and I write in a way I like?

>We don't need to debate it.

I don't care.

>And do you not want dependency update PRs because you will do them yourself or because you don't want to update dependencies?

I can do it myself.

RazrFalcon:
Also hide `usvg::ImageRendering` behind the feature.

e00E:
I have collapsed the three features into a single "image" feature.

> All those ImageKinds should be feature-gated on the usvg side.

ImageKind shows up in usvg's public api and usvg itself is a public (non internal) crate. Changing the public api based on a feature  is problematic. For example, you might be writing code against usvg without default features. You do not use the image feature so the enum only has the SVG variant. Then if at some point anywhere in your dependency tree usvg is pulled in with the image feature enabled your code will stop compiling.

In the special case of an enum we could solve this by marking it `non_exhaustive`. On the other hand there is little motivation to feature gate this in usvg in the first place because it only adds a couple of lines of code and no dependencies.

Let me know what you prefer: 
1. not adding the feature to usvg
2. adding the feature with non_exhaustive
3. adding the feature without non_exhaustiv

> Because it's my project and I write in a way I like?

Definitely. I'm curious *why* you don't like it. By giving your reasoning you might be able to change my opinion and if you don't mind debating it I might be able to change yours.

RazrFalcon:
Hm... yes, I guess we can apply this feature only to `resvg`, since it doesn't really affect `usvg`.

The problem is that this feature would disable only raster images, but not SVG images. I think we should call it `raster-images` and make it available only in `resvg`.

```toml
# enables decoding and rendering of raster images
# when disabled, `image` elements with SVG data will still be renderer
raster-images = ["gif", "jpeg-decoder", "png"]
```

>I'm curious why you don't like it.

I don't like getting _noise_-patches. I can run `cargo clippy`, `cargo update`, `cargo fmt` and so on by myself. It's not that hard. And I usually do it right before releasing a new version. Accepting an "update dependencies" patch and then releasing a new version of resvg 2-4 months later doesn't make any sense.

I'm grateful when people try to untangle the mess resvg/SVG is, but running `cargo clippy` (or any automated tool) is not a meaningful change.

Cargo.toml
```diff
@@ -17,11 +17,14 @@ name = "resvg"
 required-features = ["filter", "text", "system-fonts", "memmap-fonts"]
 
 [dependencies]
-gif = "0.11"
-jpeg-decoder = { version = "0.2", default-features = false, features = ["platform_independent"] }
+# enables embedded GIF support
```
RazrFalcon, line 20: Outdated comments?

src/image.rs
```diff
@@ -20,30 +20,38 @@ pub fn draw_kind(
...
             match read_gif(data) {
                 Some(image) => { draw_raster(&image, view_box, rendering_mode, canvas); }
                 None => log::warn!("Failed to decode a GIF image."),
             }
         }
-        usvg::ImageKind::SVG(ref subtree) => {
-            draw_svg(subtree, view_box, canvas);
+        #[cfg(not(feature = "image"))]
+        _ => {
+            log::debug!("Not decoding embedded image because feature is not enabled.");
```
RazrFalcon, line 49: Prefer `log::warn`.

Also, I would prefer: `Images decoding was disabled by a build feature.`

src/image.rs
```diff
@@ -20,30 +20,38 @@ pub fn draw_kind(
...
         }
     }
 }
 
+#[cfg(feature = "image")]
```
RazrFalcon, line 54: I think we could move all raster methods and types into a `raster_image` submodule. I.e. `mod raster_image {}`. This would introduce nesting, but would make our intentions clearer.

RazrFalcon (changes requested):

e00E:
Addressed your review comments. Diff is easier to read with "ignore whitespace'. Can squash if you prefer.

> I don't like getting _noise_-patches. I can run `cargo clippy`, `cargo update`, `cargo fmt` and so on by myself. It's not that hard. And I usually do it right before releasing a new version.

From my perspective with this contribution, I would have preferred already applied fmt and clippy. It's not a big deal but it just makes it a little easier to work with. When I first made the PR I habitually pressed my cargo fmt hotkey. Then I had to redo the changes after realizing I just touched most of the files in the project. If the goal is to generally have clippy and fmt applied after releases anyway then it doesn't feel like a downside to do it per PR already.

The point about dependencies I completely agree with.

RazrFalcon (approved):

RazrFalcon:
Thanks!

I plan to use `cargo fmt` eventually.

PR #554, opened 2022-10-30
Cargo.toml: Specify `rust-version` (currently 1.56)

EwoutH:
Specify the `rust-version` field in Cargo.toml, so that the minimum supported Rust version is enforced. The prevents installation on older Rust versions, and particularly useful when updating the minimum Rust version in the future (in which case older versions will keep working with older Rust versions).

See https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field


RazrFalcon:
Yes, I'm aware of this feature, but resvg still targets 1.51, which doesn't support it yet.

EwoutH:
You’re right, I mixed up the numbers. It wasn’t my intention to raise the requirement, so I will close this for now.

PR #567, opened 2022-12-27
Fix handling of `usvg::Options::font_family`

ANSH3LL:


RazrFalcon:
Sorry, but we must not add it by default. Only when the list is empty. The current code is correct.

PR #577, opened 2023-02-08
Add error case: 'the previous segment must be M/L/C'

yisibl:
This PR adds a test case to ensure that special path parsing does not throw an error. The following error can be reproduced in the <=0.28.0 version.

```
git checkout v0.28.0
cargo run --release -- tests/svg/a-path-001.svg tests/png/a-path-001.png
```

There will be an error:

```
thread 'main' panicked at 'the previous segment must be M/L/C', usvg/src/pathdata.rs:219:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

image


RazrFalcon:
Very strange. If I downgrade just `kurbo` - it still works just fine. Will investigate.

RazrFalcon:
See: https://github.com/RazrFalcon/resvg/tree/master/tests#adding-a-new-test

The PNG should be 300x300. And call it `e-path-053.svg`

yisibl:
I also tried to downgrade kurbo, and it did.

RazrFalcon:
Can confirm on v0.28. Very strange. Will have to investigate the cause.

yisibl:
Since our online service relies on new releases, I have prioritized the release of `resvg-js` 2.4.0. I will update ChangeLog as soon as you have new findings.

https://github.com/yisibl/resvg-js/releases/tag/v2.4.0

RazrFalcon:
Finally found to look into it and it's bizarre. I worked with SVG for more than 10 year and I didn't knew that MoveTo after ClosePath is optional...

>If a "closepath" is followed immediately by any other command, then the next subpath starts at the same initial point as the current subpath.

Meaning that `M L z A` is totally fine. And `resvg` doesn't handle this case properly. It got accidentally fixed, but there is still a bug.

RazrFalcon:
Yes, this was an SVG path data parsing bug. Not even `usvg` one. Somehow the bug slipped unnoticed. Now it's fine.

PR #580, opened 2023-02-13
store svg size in private data member

missdeer:
c-api forgets to store svg size in private data when it loads data from memory directly.


RazrFalcon:
Thanks.

PR #584, opened 2023-02-23
Fixed bug where clipping rect wasn't calculated properly for nested s…

LaurenzV:
Consider the following svg:

```svg
<svg
    xmlns="http://www.w3.org/2000/svg"
    viewBox="0 0 500 500"
    width="100%" height="100%"
>
    <rect width="100%" height="100%" fill="white"/>
    <svg x="0" y="0" width="50%" height="50%">
        <rect width="100%" height="100%" fill="red"/>>
    </svg>
</svg>
```

It should look like this:
nested_2

But instead, it looks like this:
nested_2

I did some debugging, and I believe that I found the culprit:

https://github.com/RazrFalcon/resvg/blob/f146503f84c48875f39a159a8ca4a41833abd090/usvg/src/use_node.rs#L141-L168

In this part of the code, the viewbox of the nested svg is correctly calculated as being 250x250. However, since the `state` variable now contains the new viewbox, when calculating the clipping rect in line 162, we _again_ apply the 50% width/height when calculating the dimensions of the clipping rect, meaning that the resulting dimension will be 125x125, which explains the result we get in the picture above.

I'm afraid I don't have a deep enough insight into library, so I applied the fix that seemed most intuitive to me (namely saving the new state in a new variable and using the old state for the calculation of the clipping rect), but I'm not sure if this is the right way to fix this bug and if this could have any unintended consequences (but at least all of the tests seem to pass).

Would be great if someone could comment on this. :)


RazrFalcon:
Interesting. It doesn't break any existing tests, which is good. But I would like you to [add a new test case](https://github.com/RazrFalcon/resvg/tree/master/tests#adding-a-new-test). It would be `e-svg-041.svg`. You can use `e-svg-024.svg` as a reference.

LaurenzV:
Should I add the test directly to this repo or `resvg-test-suite`? According to the README, `resvg-test-suite` is the source of truth, but it seems like this repo seems to be more up-to-date than the other one (for example the test case `e-path-053.svg` only exists here and not in `resvg-test-suite`.

LaurenzV:
I went ahead and just implemented the test here for now (and I also fixed two test cases which make `check.py` terminate with an error. If you want me to add the test to `resvg-test-suite` as well, just let me know.

RazrFalcon:
Don't worry. I will update `resvg-test-suite` myself.

PR #585, opened 2023-02-24
Extend C API to allow navigating the render tree

vonknio:
When working with the C API I needed to extract information from render tree nodes for the purpose of manual rendering. This PR extends the C API with functions that allow this, including functions that allow the navigation of the render tree, as well as functions that extract data from nodes such as transform, stroke colour etc.

Some similar functionality was already available, but it used string ids as node identifiers, and not all nodes have unique ids.

### DIFF
--- a/c-api/lib.rs
+++ b/c-api/lib.rs
@@ -13,7 +13,7 @@ use std::os::raw::c_char;
 use std::slice;
 
 use resvg::tiny_skia;
-use resvg::usvg::{self, NodeExt};
+use resvg::usvg::{self, NodeExt, NodeKind, PathCommand, PathSegment};
 #[cfg(feature = "text")]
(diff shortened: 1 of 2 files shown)


RazrFalcon:
You should have asked beforehand. There are reasons why it wasn't available...
I do not plan providing such API, because `usvg` changing non-stop and keeping C API up to date would be a nightmare.

Also, such API would be extremely slow. Almost every call in the new API is indexing + matching. For something like path segments it would absurdly slow. I do not plan adapting `usvg` API to C API needs. This is not the goal. C API is just for rendering.

You should create a separate project just for an extended API. I have no time supporting and testing it. It's a huge maintenance burden.

vonknio:
These are modifications I initially made for my own use. If you are against incorcorporating them, fair enough 🤷

PR #586, opened 2023-02-24
Fixed bug where applying text decoration with only one cluster resulted in a panic

### LINKED ISSUE #581: panic with SVG generated by PlantUML
Example SVG:
https://www.plantuml.com/plantuml/svg/SoWkIImgAStDuU9ApaaiBbPmoy_dujBcud98pKi12WC0

call stack:
```txt
# Child-SP          RetAddr               Call Site
00 00000042`c8ef9db0 00007ffd`c8846720     KERNELBASE!RaiseException+0x69
01 00000042`c8ef9e90 00007ff7`cdf63f48     VCRUNTIME140!_CxxThrowException(void * pExceptionObject = 0x00000042`c8ef9f10, struct _s__ThrowInfo * pThrowInfo = <Value unavailable error>)+0x90 [D:\a\_work\1\s\src\vctools\crt\vcruntime\src\eh\throw.cpp @ 75] 
02 (Inline Function) --------`--------     Yiili!panic_unwind::real_imp::panic(void)+0x77 [/rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library\panic_unwind\src\seh.rs @ 322] 
03 00000042`c8ef9ef0 00007ff7`cdebcfcb     Yiili!panic_unwind::__rust_start_panic(void)+0x88 [/rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library\panic_unwind\src\lib.rs @ 103] 
04 00000042`c8ef9f30 00007ff7`cdebcca7     Yiili!std::panicking::rust_panic(void)+0x1b [/rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library\std\src\panicking.rs @ 736] 
05 00000042`c8efa000 00007ff7`cdebc95b     Yiili!std::panicking::rust_panic_with_hook(void)+0x297 [/rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library\std\src\panicking.rs @ 706] 
06 00000042`c8efa0f0 00007ff7`cdeb8def     Yiili!std::panicking::begin_panic_handler::closure$0(void)+0x8b [/rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library\std\src\panicking.rs @ 577] 
07 00000042`c8efa170 00007ff7`cdebc5b0     Yiili!std::sys_common::backtrace::__rust_end_short_backtrace<std::panicking::begin_panic_handler::closure_env$0,never$>(void)+0x1f [/rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library\std\src\sys_common\backtrace.rs @ 137] 
08 00000042`c8efa1b0 00007ff7`ce339e35     Yiili!std::panicking::begin_panic_handler(void)+0x70 [/rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library\std\src\panicking.rs @ 575] 
09 00000042`c8efa200 00007ff7`ce339f5c     Yiili!core::panicking::panic_fmt(void)+0x35 [/rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library\core\src\panicking.rs @ 64] 
0a 00000042`c8efa250 00007ff7`cdfa43b8     Yiili!core::panicking::panic(void)+0x4c [/rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library\core\src\panicking.rs @ 111] 
0b 00000042`c8efa2c0 00007ff7`cdf9b5de     Yiili!ZN16usvg_text_layout13text_to_paths17h416cba2b0444cc32E+0xe1b8
0c 00000042`c8efa550 00007ff7`cdf92ae5     Yiili!ZN16usvg_text_layout13text_to_paths17h416cba2b0444cc32E+0x53de
0d 00000042`c8eface0 00007ff7`cdf95440     Yiili!ZN65_$LT$usvg..text..Text$u20$as$u20$usvg_text_layout..TextToPath$GT$7convert17hb74169065f07d82dE+0x75
0e 00000042`c8efb140 00007ff7`cde39ed4     Yiili!ZN16usvg_text_layout12convert_text17h3d42be6e397c8a03E+0x4b0
0f 00000042`c8efb270 00007ff7`ce250a6b     Yiili!resvg_parse_tree_from_data+0xa4
10 00000042`c8efb390 00007ff7`ce250e04     Yiili!ResvgRenderer::load(class QByteArray * data = 0x00000042`c8eff5e8, class ResvgOptions * opt = <Value unavailable error>)+0x4b [H:\Shareware\yiili\src\thirdparty\resvg\include\ResvgQt.h @ 356] 
```

LaurenzV:
This PR resolves #581.

I was able to reduce the panic to the following svg:

```svg
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1"
     viewBox="0 0 200 200" >
    <text fill="#33FF02" font-family="sans-serif" font-size="14" font-weight="bold" lengthAdjust="spacing"
          text-decoration="wavy underline" textLength="8" x="5" y="243.3594">+
    </text>
</svg>
```

The issue lies in the following piece of code:
https://github.com/RazrFalcon/resvg/blob/00005084c2544179a7e094d4efa979df1ec04e1b/usvg-text-layout/src/lib.rs#L1845-L1850

In our case, we only have one cluster (the +), meaning that `(cluster_indexes.len() - 1)` will evaluate to 0, and thus the factor will be infinity.

Because of this, the `Rect::new` call here
https://github.com/RazrFalcon/resvg/blob/00005084c2544179a7e094d4efa979df1ec04e1b/usvg-text-layout/src/lib.rs#L807-L816
will panic, since `dec_span.width` is (in this case) `-inf`, which is not a valid length. The way I fixed this is by setting `factor` to 0 in case there is just one cluster. I'm not sure if this is the correct approach, so I would appreciate some comments on that.

Also, while playing around with the example above, I realized that just by removing the `font-family` attribute, the example doesn't panic anymore:
```svg
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1"
     viewBox="0 0 200 200" >
    <text fill="#33FF02" font-size="14" font-weight="bold" lengthAdjust="spacing"
          text-decoration="wavy underline" textLength="8" x="5" y="243.3594">+
    </text>
</svg>
```
which didn't make sense to me, because the actual bug lies inside the handling of the text decoration. After debugging this, I realized the only reason this doesn't panic is that in this case the value will be `+inf`instead of `-inf`. When checking if a length is valid, we only check whether it is > 0.0:
https://github.com/RazrFalcon/resvg/blob/00005084c2544179a7e094d4efa979df1ec04e1b/usvg/src/geom.rs#L78-L83

So in this case, it will think the width is valid. Because of this, I added a check to make sure that the width is actually finite, I hope that this is okay.


RazrFalcon:
Thanks. That `unwrap()` should be removed completely. We should simply skip spans with malformed bboxes.

Also, please add a test case.

Also, `wavy` isn't part of `SVG`.

LaurenzV:
> Thanks. That unwrap() should be removed completely. We should simply skip spans with malformed bboxes.

Are you suggesting something like:

Changing
```rust
fn convert_decoration(
    dy: f64,
    span: &TextSpan,
    font: &ResolvedFont,
    mut decoration: TextDecorationStyle,
    decoration_spans: &[DecorationSpan],
    transform: Transform,
) -> Path {
    debug_assert!(!decoration_spans.is_empty());

    let thickness = font.underline_thickness(span.font_size.get());

    let mut path = PathData::new();
    for dec_span in decoration_spans {
        let rect = Rect::new(0.0, -thickness / 2.0, dec_span.width, thickness).unwrap();

        let start_idx = path.len();
        path.push_rect(rect);

        let mut ts = dec_span.transform;
        ts.translate(0.0, dy);
        path.transform_from(start_idx, ts);
    }

    path.transform(transform);

    Path {
        visibility: span.visibility,
        fill: decoration.fill.take(),
        stroke: decoration.stroke.take(),
        data: Rc::new(path),
        ..Path::default()
    }
}
```
to
```rust
fn convert_decoration(
    dy: f64,
    span: &TextSpan,
    font: &ResolvedFont,
    mut decoration: TextDecorationStyle,
    decoration_spans: &[DecorationSpan],
    transform: Transform,
) -> Option<Path> {
    debug_assert!(!decoration_spans.is_empty());

    let thickness = font.underline_thickness(span.font_size.get());

    let mut path = PathData::new();
    for dec_span in decoration_spans {
        let rect = Rect::new(0.0, -thickness / 2.0, dec_span.width, thickness)?;

        let start_idx = path.len();
        path.push_rect(rect);

        let mut ts = dec_span.transform;
        ts.translate(0.0, dy);
        path.transform_from(start_idx, ts);
    }

    path.transform(transform);

    Some(Path {
        visibility: span.visibility,
        fill: decoration.fill.take(),
        stroke: decoration.stroke.take(),
        data: Rc::new(path),
        ..Path::default()
    })
}
```

and then
```rust
            if let Some(decoration) = span.decoration.underline.clone() {
                // TODO: No idea what offset should be used for top-to-bottom layout.
                // There is
                // https://www.w3.org/TR/css-text-decor-3/#text-underline-position-property
                // but it doesn't go into details.
                let offset = match text_node.writing_mode {
                    WritingMode::LeftToRight => -font.underline_position(span.font_size.get()),
                    WritingMode::TopToBottom => font.height(span.font_size.get()) / 2.0,
                };

                let path =
                    convert_decoration(offset, span, font, decoration, &decoration_spans, span_ts);

                if let Some(r) = path.data.bbox() {
                    bbox = bbox.expand(r);
                }

                new_paths.push(path);
            }
```

to
```rust
            if let Some(decoration) = span.decoration.underline.clone() {
                // TODO: No idea what offset should be used for top-to-bottom layout.
                // There is
                // https://www.w3.org/TR/css-text-decor-3/#text-underline-position-property
                // but it doesn't go into details.
                let offset = match text_node.writing_mode {
                    WritingMode::LeftToRight => -font.underline_position(span.font_size.get()),
                    WritingMode::TopToBottom => font.height(span.font_size.get()) / 2.0,
                };

                if let Some(path) =
                    convert_decoration(offset, span, font, decoration, &decoration_spans, span_ts) {

                    if let Some(r) = path.data.bbox() {
                        bbox = bbox.expand(r);
                    }

                    new_paths.push(path);
                }
            }
```

RazrFalcon:
No, I would suggest:

```rust
let rect = match Rect::new(0.0, -thickness / 2.0, dec_span.width, thickness) {
    Some(v) => v,
    None => continue,
};
```

Error handling in SVG is tricky. We should try to preserve as much as possible.

LaurenzV:
Should be good now I hope!

RazrFalcon:
Thanks!

PR #587, opened 2023-02-26
Expose tiny-skia features

e00E:
tiny-skia has a couple of features which aren't required for resvg as a library but are nonetheless always enabled: simd, png-format. It is better to leave it up to the user to decide what features they want.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ members = ["c-api", "svgfilters", "rosvgtree", "usvg", "usvg-text-layout"]
 
 [[bin]]
 name = "resvg"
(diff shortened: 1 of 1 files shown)


RazrFalcon:
No. This is not necessary. All those features must be enabled.

e00E:
It compiles fine without the png feature so it doesn't look like it needs to be enabled. The reason I made this PR is that I was going through my dependency tree and was surprised to see the png crate there when I'm not doing anything with pngs.

RazrFalcon:
The `png` crate is a tiny dependency. It might not even be present in the final binary if you do not use it.

But each build feature is a maintenance burden. And I cannot afford it. It's already too complicated.

PR #601, opened 2023-03-28
usvg: add xml option to specify decimal places limit

flxzt:
This adds an Xml option to specify the number of decimal places when writing numbers. It replaces the hard-coded limit to 8 places. This is useful when one wants to reduce the size of the usvg-export Svg.

For usage with the Cli, I also added an optional `--num-n-decimal-places` arg, it falls back to `8` if not found.

### DIFF
--- a/usvg/src/main.rs
+++ b/usvg/src/main.rs
@@ -92,6 +92,8 @@ OPTIONS:
                                 option. [values: 1..2^32] [default: 100]
 
   --id-prefix                   Adds a prefix to each ID attribute
+  --num-n-decimal-places NUM    limits the decimal places when writing numbers
+                                [values: 0..2^8] [default: 8]
   --indent INDENT               Sets the XML nodes indent
                                 [values: none, 0, 1, 2, 3, 4, tabs] [default: 4]
   --attrs-indent INDENT         Sets the XML attributes indent
@@ -128,6 +130,7 @@ struct Args {
     default_height: u32,
 
     id_prefix: Option<String>,
+    num_n_decimal_places: Option<u8>,
     indent: xmlwriter::Indent,
     attrs_indent: xmlwriter::Indent,
 
@@ -189,6 +192,7 @@ fn collect_args() -> Result<Args, pico_args::Error> {
             .unwrap_or(100),
 
         id_prefix: input.opt_value_from_str("--id-prefix")?,
+        num_n_decimal_places: input.opt_value_from_str("--num-n-decimal-places")?,
         indent: input
             .opt_value_from_fn("--indent", parse_indent)?
             .unwrap_or(xmlwriter::Indent::Spaces(4)),
@@ -407,6 +411,7 @@ fn process(args: Args) -> Result<(), String> {
 
     let xml_opt = usvg::XmlOptions {
         id_prefix: args.id_prefix,
+        num_n_decimal_places: args.num_n_decimal_places.unwrap_or(8),
         writer_opts: xmlwriter::Options {
             use_single_quote: false,
             indent: args.indent,
--- a/usvg/src/main.rs
+++ b/usvg/src/main.rs
@@ -22,88 +22,91 @@ USAGE:
   usvg [OPTIONS] - -c                # from stdin to stdout
 
 OPTIONS:
-  -h, --help                    Prints help information
-  -V, --version                 Prints version information
-  -c                            Prints the output SVG to the stdout
+  -h, --help                        Prints help information
+  -V, --version                     Prints version information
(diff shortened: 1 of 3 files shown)


RazrFalcon:
I'm skeptical about this change. I had this feature in the [svgcleaner](https://github.com/RazrFalcon/svgcleaner/blob/master/src/task/round_numbers.rs) and it's a very tricky one.

First, it's not lossless. You can break SVG in unexpected ways. This should be clearly telegraphed.
Second, transforms precision almost always should be higher, therefore we need two options.
Third, your rounding method is extremely slow. See the one in `svgcleaner`.

Also, we should call it `coordinates_precision` instead of `num_n_decimal_places` and the range should be like 2..8

flxzt:
fair points, I'll do my best to address them.

Two questions:

- do you have a suggestion for a formulation for the new cli args / fields telling users that it is not lossless / that the Svg might break when the value is a low number?
- so the other option should be called `transformations_precision` ? And the range for should be 2..8 (I assume including 8 for both then) as well?

RazrFalcon:
As for help, I would suggest:

```
--coordinates-precision <NUM>           Set the coordinates numeric precision.
                                        Smaller precision can lead to a malformed output in some cases
                                        [values: 2..8] [default: 8]
--transforms-precision <NUM>            Set the transform values numeric precision
                                        Smaller precision can lead to a malformed output in some cases
                                        [values: 2..8] [default: 8]
```

And put it below `--attrs-indent`.

I guess we would have to shift help messages by 4 spaces as well, to fit longer options.

Yes, 8 is included.

usvg/src/main.rs
```diff
@@ -261,6 +270,16 @@ fn parse_length(s: &str) -> Result<u32, String> {
     }
 }
 
+fn parse_precision(s: &str) -> Result<u8, String> {
+    let n: u8 = s.parse().map_err(|_| "invalid precision NUM value")?;
+
+    if n <= 8 {
+        Ok(n)
+    } else {
+        Err("precision NUM cannot be larger than eight".to_string())
```
generalmimon, line 279:
```suggestion
        Err("precision NUM cannot be larger than 8".to_string())
```

RazrFalcon:
I don't see why we need `2^32 - 1`? It's suppose to be `u32::MAX`.

generalmimon:
@RazrFalcon:

> I don't see why we need `2^32 - 1`? It's suppose to be `u32::MAX`.

I can see @flxzt's point - `u32::MAX` **is** `2^32 - 1`:

```rust
fn main() {
    println!("{}", u32::MAX); // 4294967295
    println!("{}", u64::pow(2, 32) - 1); // 4294967295
}
```

Because the `u32` type is 32-bit, the _number_ of values it can hold is `2^32`, but you have to count `0` as well. So the range is (in Rust notation) `0..=(2^32 - 1)`.

RazrFalcon:
Hmm... My bad. The ranges syntax is a mess. Everyone do whatever they want. `..` is exclusive in Rust, but inclusive in Ruby.

Let's settle on `[values: 2..8 (inclusive)]` and `[values: 1..4294967295 (inclusive)]`. And add ` (inclusive)` to all ranges in the help. Including one in resvg CLI.

>it may also help to specify it in hex notation: 0x0000_0000_u32..=0xffff_ffff_u32

I don't think so. This tool is for non-developers.

generalmimon:
> > it may also help to specify it in hex notation: 0x0000_0000_u32..=0xffff_ffff_u32
> 
> I don't think so. This tool is for non-developers.

Of course, sorry for confusion, I didn't meant to put it in help text either. That was just to demonstrate that the `u32` range is the way it is.

RazrFalcon:
I see. No problem.

usvg/src/writer.rs
```diff
@@ -25,15 +25,34 @@ impl<T: Default + PartialEq + Copy> IsDefault for T {
 }
 
 /// XML writing options.
-#[derive(Clone, Default, Debug)]
+#[derive(Clone, Debug)]
 pub struct XmlOptions {
     /// Used to add a custom prefix to each element ID during writing.
     pub id_prefix: Option<String>,
 
+    /// Set the coordinates numeric precision.
```
RazrFalcon, line 33: Add a new line after the first sentence, so docs would be rendered correctly.

And add a period at the end of the second sentence.

flxzt: I always forget about that

RazrFalcon:
Do you consider this patch finished or do you plan more changes?

flxzt:
No, it ise ready from my side

Issue #241, opened 2020-02-27
Invalid blur clipping


JoKalliauer:
#### svg

![svg](https://upload.wikimedia.org/wikipedia/commons/0/05/Vector_saturn.svg)

#### png

|ink|rsvg|resvg|
|--|--|--|
|![Vector saturn](https://user-images.githubusercontent.com/15687581/75484392-010a0380-59a9-11ea-8f84-2cf7c0313407.png)|![Vector saturn_r](https://user-images.githubusercontent.com/15687581/75484398-01a29a00-59a9-11ea-9219-135906684973.png)|![Vector saturn_re](https://user-images.githubusercontent.com/15687581/75484400-023b3080-59a9-11ea-88ad-2deacdfccf8a.png)|

#### Expected
![PNG](https://upload.wikimedia.org/wikipedia/commons/8/85/Vector_saturn_%28Correct_render%29.png)

#### Copyright
source https://commons.wikimedia.org/wiki/File:Vector_saturn.svg

author Kelvinsong

license https://creativecommons.org/licenses/by-sa/3.0/deed.de

RazrFalcon:
Looks like everyone is wrong:

![Vector_saturn](https://user-images.githubusercontent.com/725494/75491693-040fee80-59bf-11ea-94b5-c66a47489cbf.png)

Will investigate further to find the source of the issue.

matpaul:
@RazrFalcon Hi! I try use your implementation of fastblur and saw same effect
fix for me was in `box_blur_vert` change initialization of fv, lv 
```rs
let fv: RGBA8 = backbuf[ti];
let lv: RGBA8 = backbuf[ti + width * (height - 1)]; 
```

in `box_blur_horz` change change initialization of fv, lv
```rs 
let fv: RGBA8 = backbuf[ti];
let lv: RGBA8 = backbuf[ti + width - 1];
```

ps. sorry don't have enough time to PR

RazrFalcon:
Thanks, will try it out.

yisibl:
Hi, @RazrFalcon 

I think the Blur filter might be a separate issue that we encountered in real business and got fixed by @matpaul 's code.

Can we fix the Blur filter issue https://github.com/RazrFalcon/resvg/issues/324 first?

Test case:

![blur-test](https://user-images.githubusercontent.com/2784308/164417284-92ecd11b-741e-44ee-9b79-eb171989a380.svg)

RazrFalcon:
[...]

PR #615, opened 2023-04-20
fix blur issue

m1guelpf:
@yisibl's fix for #241

### DIFF
--- a/svgfilters/src/box_blur.rs
+++ b/svgfilters/src/box_blur.rs
@@ -101,8 +101,8 @@ fn box_blur_vert(blur_radius: usize, backbuf: &ImageRefMut, frontbuf: &mut Image
         let mut li = ti;
         let mut ri = ti + blur_radius * width;
 
-        let fv = RGBA8::default();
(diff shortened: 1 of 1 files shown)


RazrFalcon:
Looks like it breaks the current output quite a bit. Will take a look.

RazrFalcon:
Doesn't seem to work at all. Try rendering `tests/svg/e-filter-002.svg`. It's has no blur anymore.

yisibl:
Yes, this PR will introduce new problems. We desperately need a better program.

m1guelpf:
hmm, I'm using [this fork](https://github.com/zimond/resvg/tree/resvg-31)(which includes the commit above, along with a bunch of other changes), and it fixes the issue for [my use-case](https://github.com/m1guelpf/world-id-marbles/blob/5ce76318814a1f2f96723149bbecc47e77e89e4d/packages/world-id-marbles/src/lib.rs#L126-L144).

RazrFalcon:
Well, in it's current form it break most resvg tests. I cannot accept it.

PR #620, opened 2023-05-17
Symlink resvg README

CosmicHorrorDev:
This symlinks the repo README into the `resvg` crate, so that it gets picked up by crates.io with the new structure


RazrFalcon:
Thanks!

PR #625, opened 2023-05-20
Fix simple case svg file in polygon

nnabeyang:
This PR fixes to draw a red triangle on the back.
Currently, a straight red line is drawn.


RazrFalcon:
Thanks. Good catch.

Issue #638, opened 2023-06-16
Crash when elements are translated offscreen


gmeeker:
This seems to have been introduced by 0.34.0. I realize this is not a normal use case (I only discovered it accidentally) but it's still causing resvg to crash where it didn't before. I believe this SVG is still valid.
This is macOS 13. I haven't checked elsewhere.

thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', crates/resvg/src/geom.rs:27:61
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

![crash](https://github.com/RazrFalcon/resvg/assets/6538028/12b4bec4-5de0-453a-8709-16e0ed093aa3)

jonmmease:
I opened a PR in https://github.com/RazrFalcon/resvg/pull/639 with this proposed fix

jonmmease:
Thanks for the quick fix @RazrFalcon! 

One question, I saw that the fix was published in 0.35.0 as a GitHub Release, but I don't see it on [crates.io](https://crates.io/crates/resvg) yet (It still shows 0.34.1 as the latest). Do you know when that will happen? Thanks again!

RazrFalcon:
Should be up now.

jonmmease:
perfect, thanks!

PR #639, opened 2023-06-27
Prevent crash when element is transformed out of the view box

jonmmease:
Proposed fix for #638, as suggested in https://github.com/RazrFalcon/resvg/issues/638#issuecomment-1607746136.

I also added a new test at `crates/resvg/tests/tests/structure/transform/translate-out-of-view.svg` that previously crashed with the error described in #638.

Thanks for the great library and for taking the time to consider this!


jonmmease:
I'm not sure why CI isn't running here, is it configured to not run on forks?

RazrFalcon:
Thanks, but this is not how it should be fixed. Pushed a fix.

jonmmease:
Thanks for the fix!

PR #646, opened 2023-07-15
Fix d attribute of the path tag in rotate-at-position.svg

nnabeyang:
This PR fixes the position of the red rectangle in rotate-at-position.svg.


RazrFalcon:
Thanks.

PR #649, opened 2023-07-24
add musl release target

gustofarbi:
this PR adds a musl release to the list of current releases

### DIFF
--- a/.github/workflows/tagged-release.yml
+++ b/.github/workflows/tagged-release.yml
@@ -38,23 +38,41 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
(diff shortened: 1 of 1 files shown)


RazrFalcon:
Why?

gustofarbi:
shouldve added a better description 😅 , we are using resvg inside an alpine container in two apps and it is pretty time consuming to have an extra layer to build resvg, would be nice to just download the prebuilt binary

RazrFalcon:
Sorry, but this is a very niche use case. You should host prebuilt versions yourself.

RazrFalcon:
Because you can easily build it yourself.

PR #652, opened 2023-08-04
Fix broken link to tests

mmoult:
Fix the broken link for SVG test cases. Previously, the link pointed to the test directory in resvg (which is now located at https://github.com/RazrFalcon/resvg/tree/master/crates/resvg/tests#resvg-tests-vs-resvg-test-suite-tests), but I think that pointing to the test suite repo is more appropriate. After all, the discussion at https://github.com/RazrFalcon/resvg/tree/master/crates/resvg/tests#resvg-tests-vs-resvg-test-suite-tests says that the repo is the source of truth.


RazrFalcon:
Thanks!

PR #661, opened 2023-09-14
Add `auto-start-reverse` variant to `orient`

EpicEricEE:
> If placed by [marker-start](https://svgwg.org/svg2-draft/painting.html#MarkerStartProperty), the marker is oriented 180° different from the orientation that would be used if 'auto' where specified. For all other markers, 'auto-start-reverse' means the same as 'auto'.

I hope I didn't miss anything, and that the test is alright.


RazrFalcon:
Oh wow, thanks! I wasn't expecting it to be that simple.

PR #665, opened 2023-10-11
Add support for `context-fill` and `context-stroke`

LaurenzV:
Here is the [corresponding specification](https://www.w3.org/TR/SVG2/painting.html#SpecifyingPaint).

This is still a WIP, in particular, the following I still haven't looked into more deeply (as I'm still unfamiliar with the codebase) and I think might be a bit tricky to implement:
> When the context paint layers include paint server references, then the coordinate space and the bounding box used to scale the paint server elements and content are those of the [context element](https://www.w3.org/TR/SVG2/painting.html#TermContextElement). In other words, any gradients and patterns referenced with these keywords should be continuous from the main shape to the markers, or from one element within a [use-element shadow tree](https://www.w3.org/TR/SVG2/struct.html#TermUseElementShadowTree) to another.

But I'm opening this as a draft PR so that people know that someone is working on it. 😄 

I also had to temporarily switch to my own version of `svgtypes` since it's not included in this repository.


crates/usvg-parser/Cargo.toml
```diff
@@ -22,5 +22,5 @@ log = "0.4"
 roxmltree = "0.18"
 simplecss = "0.2"
 siphasher = "0.3" # perfect hash implementation
-svgtypes = "0.12"
+svgtypes = {path = "../../../svgtypes"}
```
RazrFalcon, line 25: You could link git version here.

crates/usvg-parser/src/converter.rs
```diff
@@ -18,6 +18,8 @@ use crate::{Error, Options};
 pub struct State<'a> {
     pub(crate) parent_clip_path: Option<SvgNode<'a, 'a>>,
     pub(crate) parent_markers: Vec<SvgNode<'a, 'a>>,
+    pub(crate) context_stroke: Option<Stroke>,
```
RazrFalcon, line 21: We have to make sure it would work with nested `use` and `marker`.

crates/usvg-parser/src/converter.rs
```diff
@@ -18,6 +18,8 @@ use crate::{Error, Options};
...
+    pub(crate) context_fill: Option<Fill>,
```
RazrFalcon, line 22: Swap fill and stroke, to make fill first.

RazrFalcon:
For marker tests, let's use triangle and not a line for the base shape. This way we would be able to test `marker-mid` as well.
And let's remove opacity one markers as well.

RazrFalcon:
Looks ok overall. The feature itself is somewhat convoluted and niche. Even Chrome and Safari doesn't support it yet.

Nested `use` and `marker` as well as paint servers bbox preserving seems like the only tricky parts.
We should have tests for all of them. I can help with tests if needed.

LaurenzV:
No worries, I can do the tests! Might need a bit help with figuring out how to properly deal with paint servers, but I'll have a look at it myself later / in the next few days.

crates/usvg-parser/src/converter.rs
LaurenzV: I added some more tests.

LaurenzV:
> For marker tests, let's use triangle and not a line for the base shape. This way we would be able to test marker-mid as well.
And let's remove opacity one markers as well.

I used the start instead to make it more consistent with other tests, I hope that is okay.

LaurenzV:
I looked a bit more into the source code, and I just had two (somewhat) related questions:

1. Is there a reason why we need to keep the `transform` attribute in `usvg::Path`? Because as far as I can tell, if a `path` element in SVG has a `transform` attribute, it will automatically be converted into a `Group` with a corresponding transform, so the transform of the path itself always stays the default. At least I just tried what happened when I remove it and apart from the fact that the `writer` might need to be changed a bit, all test cases seem to pass. So I guess this would be another nice simplification of `usvg:Tree`. But I'm not sure if I'm missing something.

2. This one might be a bit more difficult to implement, but currently, `usvg:Tree` distinguishes between the units `ObjectBoundingBox` and `UserSpaceOnUse`. But whenever there is `ObjectBoundingBox` it mainly seems to be about adding another transform to the transform chain to account for the bbox of the object. So I was wondering whether this also is something that could be simplified at the `usvg:Tree` level to make it even simpler? Or is there a good reason to preserve this semantic difference at that stage?

RazrFalcon:
>Is there a reason why we need to keep the transform attribute in usvg::Path? 

Probably not. `usvg` was rewritten a billion times. Probably just a left-over.  Let's leave to a separate patch.

>Or is there a good reason to preserve this semantic difference at that stage?

Ugh... were do I even begin. I plan to write a whole chapter for [Notes on SVG parsing](https://razrfalcon.github.io/notes-on-svg-parsing/) about `ObjectBoundingBox`. Maybe one day.
In short - no, because of text. Text doesn't have bbox until we outline it, which we cannot do during parsing.

LaurenzV:
> Probably not. usvg was rewritten a billion times. Probably just a left-over. Let's leave to a separate patch.
Okay, I'll open a separate PR for this!

> In short - no, because of text. Text doesn't have bbox until we outline it, which we cannot do during parsing.
Ah, that's unfortunate. :(

RazrFalcon:
>Ah, that's unfortunate.

It is... `objectBoundingBox` is one of the worst SVG features. It's nice on paper, but it's ridiculously hard to implement. Not only a text bbox is a very complicated thing in SVG (no, it's not an outlined path bbox), but even without text there are tons of edge cases.

I've tried multiple times to get rid of `objectBoundingBox`, but it's not possible.
Same with transforms. It's technically impossible to flatten transforms in SVG, but people keep asking me to implement this. Because they view SVG a list of paths, which is like 1% of the spec.

LaurenzV:
So after doing some more digging into the source code, I'm honestly a bit clueless about how to implement proper support for gradients/patterns without introducing too much "ugly" code. Two approaches I could think of are:

1) In the `usvg:Tree`, all we change is that for patterns and gradients, we store some kind of additional flag that indicates whether the paint should be resolved relative to the bbox and transform of the current path, or whether the context element should be chosen for that. This would reduce the complexity in the tree itself (since we are only adding a simple flag) but leave it upon `resvg` or other renderers to deal with how exactly to implement this. 

However, in addition to that, we also could add another flag to groups to indicate whether this group is a context element. By doing this, the renderer can find the corresponding context element of a path simply by traversing its ancestors until it hits the first group that is a context element. Then, it simply needs to resolve the bbox and transform of that group and it can use this apply the appropriate bounding boxes and transforms to the paint.

2) Instead of doing the above, we try to somehow embed the bounding box and transform information directly into the paint. This way, we don't have to leave it to the renderer to do that. However, as far as I can tell while parsing we can't even really determine the bounding boxes, so I'm not even sure if this would work, let alone the fact that this would introduce a lot more complexity just to support a simple feature...

So yeah, I think for now I will attempt to implement the first approach and see how far I can get.

RazrFalcon:
Ideally, `usvg` should do all of this. Like we already have `text_bbox` in `Path`. And don't forget [this monstrosity](https://github.com/RazrFalcon/resvg/blob/master/crates/usvg-text-layout/src/lib.rs#L796-L879). Therefore I would suggest doing something similar.

LaurenzV:
No worries, I'll also give it some more thought!

LaurenzV:
So I think implementing proper support for gradients/patterns will be really tricky and probably require some bigger changes. Would it be okay for you if we merge this now (so that it at least works properly for solid colors) and create an issue so that we don't forget about adding support for gradients/patterns as well later on? I think for now I will try working on some other "low-hanging fruits" (i.e. some other SVG2 features) before getting back to this.

But if not, I it's fine as well, but not sure when I will revisit this PR. I need to gain some more familiarity with the library by working on some other easier issues before proceeding with this.

RazrFalcon:
I would be able to work on it maybe next month. I don't want to merge a half-backed solution. It's not very `resvg`-like.

As for `typst` - never, ever under any circumstances use SVG for anything. In SVG, you cannot even renderer paths reliably. The format is a joke.
But if you have to, try using only the most primitive SVG features possible. And markers are definitely not one of them. They are essentially unsupported by most libraries. You should not write them to begin with.
And obviously forget about SVG 2.

LaurenzV:
> I would be able to work on it maybe next month. I don't want to merge a half-backed solution. It's not very resvg-like.

I totally understand, let's leave it open then. :)

reknih:
A use-case I have seen for this is marker color. Inkscape seems to default to specifying marker fill as `context-stroke` for arrows now.

RazrFalcon:
@reknih as I've mentioned before, you should avoid using markers to begin with. They are [poorly supported](https://razrfalcon.github.io/resvg-test-suite/svg-support-table.html#painting-marker). And any SVG 2 features are even worse.
As for `context-fill`/`context-stroke` and any SVG2 features - it will be done when it will be done. I have no time working on this project in the near feature.

reknih:
> it will be done when it will be done. I have no time working on this project in the near feature.

No worries 🙌

LaurenzV:
Added some more test cases which should cover the most tricky parts when implementing this for gradients and patterns. In case anyone feels adventurous enough to actually try this...

RazrFalcon:
Thanks.

LaurenzV:
I wanted to ask again whether you really wouldn't consider merging this until it also works correctly for pattern and gradient fills? I understand the argument of not wanting a half-baked solution, but it seems like at least Inkscape, Firefox and librsvg already support `context-fill` and `context-stroke` for solid colors, while they face the exact same issues for gradient and pattern fills. To me, this seems to be more of an edge case.

And #707 seems to be just another example that this feature is actually used by some producers, so I think it would make sense to at least provide support for solid colors. And we can just create an issue for the other two so we don't forget about it.

RazrFalcon:
I will look into it on the weekends. Maybe we will merge it as is.

RazrFalcon:
As for commented out test - simply enable them. I doesn't matter if the current output is incorrect.

Does paint servers work when `user` units are used? 

As for object bbox resolving, if my brain understands it correctly, can't we simply override the bbox in `usvg::paint_server::node_to_user_coordinates`? We have a similar trick already for text bbox, which seems to act similar to context.
And since we always represent `use` as a `Group` we can add a flag to a group like `use_for_context`. In which case it would set `text_bbox` (should rename) to the group's bbox. Should work, I guess.

Markers are harder, but since markers can only be set on paths, we already know the bbox. So we just have to propagate it. Right? I can modify `paint_servers.rs convert()` to accept an optional bbox.

LaurenzV:
> As for commented out tests - simply enable them. It doesn't matter if the current output is incorrect.

I'll do that.

> Does paint servers work when user units are used?

Nope, not yet. I still haven't gotten to trying it out yet (I just made a commit to merge it with the new resvg), I'll give it a shot soon!

yisibl:
I asked [here](https://issues.chromium.org/issues/40362923#comment29) if Chrome has any plans to support gradients.

LaurenzV:
Ah, then we definitely should look into that... Sorry for the delay from my side, I will try to make some time for it in the next few days.

LaurenzV:
Making progress! :D

context-with-pattern-and-transform-in-use

LaurenzV:
@RazrFalcon context in `use` seems to work fine now (although I still need to clean up the code), but no idea how to deal with markers... Because for use node I could just go through all ancestors to find the corresponding context group, while for markers this doesn't work because markers aren't children of their corresponding paths.

RazrFalcon:
Aren't marker context is the current path fill/stroke? And we already resolved it.

LaurenzV:
I think I figured it out. :D Will clean up a bit and mark the PR as ready to review once I'm done.

LaurenzV:
I hope it's okay!

crates/usvg/src/parser/converter.rs
```diff
@@ -596,6 +608,7 @@ pub(crate) fn convert_group(
     force: bool,
     cache: &mut Cache,
     parent: &mut Group,
+    is_context_element: bool,
```
RazrFalcon, line 611: Since this flag doesn't affect `convert_group` logic, let's set it to `false` by default. And modify in `use_node.rs` when necessary. 

`convert_group` already has too many arguments and because Rust doesn't support named arguments - it looks like a mess.

crates/usvg/src/tree/mod.rs
```diff
@@ -628,12 +631,29 @@ impl Default for FillRule {
...
+
 /// A fill style.
 #[derive(Clone, Debug)]
 pub struct Fill {
     pub(crate) paint: Paint,
     pub(crate) opacity: Opacity,
     pub(crate) rule: FillRule,
+    // Whether the current fill needs to be resolved relative
+    // to a context element.
+    pub(crate) context_element: Option<ContextElement>,
```
RazrFalcon, line 656: The only reason we have to store it is because of paint bbox resolver?
I think we can avoid it for markers, since we already know the bbox beforehand, unlike with text.
But I guess I will try figuring it out by myself, since it would be a bit tricky.

RazrFalcon:
Paint resolver code looks a bit too complicated. Will see if I would be able to simplify it.
Otherwise, tests look good. And I assume the output matches the expected one (chrome fails to render correctly).
Not much else from me. Almost ready to merge.

LaurenzV:
> And I assume the output matches the expected one.

Yes, it does! All of them.

> Paint resolver code looks a bit too complicated. 

You mean `convert_paint`? Yes, the fact that we are returning a tuple here is kind of ugly. The only reason we need this is because we are storing the context element inside of the fill/stroke and not the paint. Since `convert_paint` only returns a paint, we need to preserve that information somehow so that we can insert it again once we construct the new fill/stroke. Maybe it's possible to store it inside the paint directly, not sure if this would require many changes.

I know the code isn't great. :( But it's also a tricky feature to implement...

crates/usvg/src/tree/mod.rs
LaurenzV: Yes, we mainly need it as a "marker" so that when iterating over all paint servers we know which strokes/fills need to be converted.

RazrFalcon:
Thanks again for looking into this! I will look if it can be simplified a bit and will publish a new release probably on weekends.

LaurenzV:
Thanks, I'm glad this feature is implemented now!

PR #666, opened 2023-10-13
A mixed bag of fixes :)

nyurik:
* Fixed Build badge link to CI
* A few grammar & spelling mistakes
* inline format args where appropriate - makes it a bit easier to read and follow
* Removed un-needed borrow (per clippy suggestion)
* Replaced `.rev().next()` with `next_back()` (per clippy)
* Ran `cargo fmt` to keep things a bit tidier


nyurik:
P.S. in all seriousness though - I was looking at some `unwrap()` - seems they may crash during execution (I am still investigating) - and not certain if it would make sense to change them to `Result`s instead?

RazrFalcon:
Thanks, but I don't think I'm ready to switch to inline format. Let's keep it as is.

nyurik:
@RazrFalcon any objections to the other changes?

RazrFalcon:
Hard to differentiate them, but seems ok. 

I just don't like the fact that inline format doesn't support expressions. Feels half-backed.

RazrFalcon:
Thanks. In general I do not accept code formatting and clippy fixes.

PR #667, opened 2023-10-13
inline format args where appropriate

nyurik:
This makes it a bit easier to read and follow.

It used to be a part of another PR, now moved here.
In case you decide to merge it later.


RazrFalcon:
Maybe one day.

PR #668, opened 2023-10-14
Remove `transform` attribute from images and paths

LaurenzV:
Since all transforms on paths and images will be added in the form as a group, we won't really need them on `Image` and `Path` anymore. This makes the tree a bit simpler.

I hope I didn't miss anything.

### DIFF
--- a/crates/usvg-tree/src/lib.rs
+++ b/crates/usvg-tree/src/lib.rs
@@ -737,8 +737,8 @@ impl NodeKind {
     pub fn transform(&self) -> Transform {
         match self {
             NodeKind::Group(ref e) => e.transform,
-            NodeKind::Path(ref e) => Transform::default(),
-            NodeKind::Image(ref e) => Transform::default(),
+            NodeKind::Path(_) => Transform::default(),
+            NodeKind::Image(_) => Transform::default(),
             NodeKind::Text(ref e) => e.transform,
         }
     }
--- a/crates/usvg-tree/src/lib.rs
+++ b/crates/usvg-tree/src/lib.rs
@@ -738,7 +738,7 @@ impl NodeKind {
         match self {
             NodeKind::Group(ref e) => e.transform,
             NodeKind::Path(ref e) => Transform::default(),
-            NodeKind::Image(ref e) => e.transform,
+            NodeKind::Image(ref e) => Transform::default(),
             NodeKind::Text(ref e) => e.transform,
         }
     }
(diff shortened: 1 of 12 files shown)


crates/usvg-tree/src/lib.rs
```diff
@@ -737,8 +737,8 @@ impl NodeKind {
     pub fn transform(&self) -> Transform {
         match self {
             NodeKind::Group(ref e) => e.transform,
-            NodeKind::Path(ref e) => e.transform,
-            NodeKind::Image(ref e) => e.transform,
+            NodeKind::Path(_) => Transform::default(),
+            NodeKind::Image(_) => Transform::default(),
             NodeKind::Text(ref e) => e.transform,
```
RazrFalcon, line 742: Shouldn't we remove it from text as well?

LaurenzV: Possibly, kind of forgot about it. Will try.

RazrFalcon:
I guess you can update [crates/usvg/docs/spec.adoc](https://github.com/RazrFalcon/resvg/blob/master/crates/usvg/docs/spec.adoc) as well. Remove it from `image` and note that only `path`s inside `clipPath` can have a transform.

RazrFalcon:
If I remember correctly, transform is set only on groups because `filter`, `mask` and `clip-path` can be set only on groups. And

```xml
<g mask="..." transform="...">
   <path/>
</g>
```

is not the same as:

```xml
<g mask="...">
   <path transform="..."/>
</g>
```

The confusing part is that fill and stroke would still be set on the path, despite the fact they are also affected by the transform. 

Groups in SVG is a very strange and counter-intuitive concept. I wrote about them a bit [here](https://razrfalcon.github.io/notes-on-svg-parsing/isolated-groups.html) and [here](https://razrfalcon.github.io/notes-on-svg-parsing/empty-groups.html).

crates/usvg-tree/src/lib.rs
LaurenzV: Should we remove the transform method on `Node` all-together now that only groups can have them anyway? Would make a bit more sense, on the other hand it's also convenient if you can just call it on the node itself instead of checking if its a group first.

RazrFalcon: I think we should. Otherwise it's just confusing.

RazrFalcon:
Thanks!

PR #675, opened 2023-10-29
Convert text in embedded svgs as well

### LINKED ISSUE #633: No text in referenced SVGs
We should convert text to paths in referenced SVGs as well.

LaurenzV:
Fixes #633


RazrFalcon:
Thanks!

Issue #674, opened 2023-10-29
text-decoration not inherited from tspan


LaurenzV:
Consider the following SVG:
```svg
<svg width="200" height="200" viewBox="0 0 200 200" xmlns="http://www.w3.org/2000/svg">
    <defs/>
    <text id="text1" font-family="Noto Sans" font-size="48" x="50" y="100">
        <tspan text-decoration="overline" fill="#000000" stroke="#FF0000">
            <tspan>Text</tspan>
        </tspan>
    </text>
</svg>
```

When rendering with resvg, the overline will not appear at all. If I put it into the `text` element instead, it will appear. Shouldn't it also be inherited when it appears in a t-span element? Especially since the fill and stroke do seem to be inherited. At least in Chrome and Safari it seems to show up that way. Is this a bug or am I missing something about how inheritance works in this case?

RazrFalcon:
#410 ?

RazrFalcon:
`text-decoration` is insanely hard to parse and apply: https://razrfalcon.github.io/notes-on-svg-parsing/text/text-decoration.html

Issue #410, opened 2021-07-09
Indirect `text-decoration` is not resolved


RazrFalcon:
https://github.com/RazrFalcon/resvg-test-suite/blob/master/svg/a-text-decoration-019.svg

PR #678, opened 2023-10-31
Add support for indirect text-decoration resolving

LaurenzV:
I hope I didn't mess up anything in the algorithm, but at least all the test cases pass and the `indirect.svg` one looks correct now as well.

Also, I ran `cargo fmt` which changed two other files a bit, I know you usually don't take formatting PRs but I hope it's okay in this case, otherwise I always need to manually undo the changes when running the command. But if you don't want to, I'll revert that change.


RazrFalcon:
Thanks! It's even shorter now.
Once again, many tasks in resvg/usvg are not that hard, I simply have no time working on them.

Does this PR fixes #674 and #410 ?

>I know you usually don't take formatting PRs but I hope it's okay in this case

I don't take stand-alone formatting patches. Otherwise it's fine.

LaurenzV:
Yes, it fixes both of them!

RazrFalcon:
Thanks!

PR #680, opened 2023-11-04
Add support for markers in-between fills/strokes

### LINKED ISSUE #550: Figure out how to implement custom markers paint order
Since `usvg` converts markers into regular nodes, we cannot render them between path's fill and stroke. Which is allowed by the spec. We can render them only below or above the path.

The only way to implement this is by rewriting markers support. Basically removing this simplification step from `usvg`, which is unfortunate.

LaurenzV:
Fixes #550.

Although you do mention in the issue that it's not possible to implement this without adding markers in usvg itself, so I'm not sure if I'm missing something obvious? 😅 But at least the test cases work as expected.


crates/usvg-parser/src/converter.rs
```diff
@@ -694,14 +694,54 @@ fn convert_path(
...
+                }
+                PaintOrderKind::Stroke => {
+                    if !path.stroke.is_none() {
+                        let mut stroke_path = path.clone();
+                        stroke_path.fill = None;
+                        stroke_path.id = String::new();
+                        parent.append(Node::new(NodeKind::Path(stroke_path)));
+                    }
+                }
+                _ => unreachable!(),
```
RazrFalcon, line 733: Let's avoid `unreachable` just in case.

RazrFalcon:
Good question. I guess I didn't wanted to duplicated paths or thought that stroke outlining would be required, but it should be fine now.

LaurenzV:
Yeah I mean the path duplication only happens for that specific case, so I guess it is worth to clone it just in this case so that the output is correct. ^^

RazrFalcon:
Thanks!

PR #682, opened 2023-11-10
Adding text writing for usvg CLI

LaurenzV:
# Summary 

This PR includes the following changes:
- Instead of having a single `writer.rs` file with all writing operations, I split them up into different files to make it more easy to navigate the code.
- I added the ability to directly write text using the usvg CLI instead of having to convert text into paths beforehand.
- I tested everything against the test suite to make sure that it works as intended. In order to do that, I basically changed

# Test Results

In order to test the writer, I changed

```rust
let mut tree = usvg::Tree::from_data(&svg_data, &opt).unwrap();
```

to

```rust
let mut tree = usvg::Tree::from_str(
            &usvg::Tree::from_data(&svg_data, &opt)
                .unwrap()
                .to_string(&XmlOptions::default()),
            &opt,
        )
        .unwrap();
```

in `tests/integration/main.rs`. Basically, I check against the exact same reference images but by passing the simplified SVG to resvg instead of the original ones. 

The results are as follows:
```
failures:

---- render::filters_feImage_svg stdout ----
thread 'render::filters_feImage_svg' panicked at crates/resvg/tests/integration/render.rs:193:36:
assertion `left == right` failed
  left: 26
 right: 0
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- render::filters_filter_functions_color_adjust_functions_2 stdout ----
thread 'render::filters_filter_functions_color_adjust_functions_2' panicked at crates/resvg/tests/integration/render.rs:353:66:
assertion `left == right` failed
  left: 3600
 right: 0

---- render::filters_filter_functions_color_adjust_functions_200percent stdout ----
thread 'render::filters_filter_functions_color_adjust_functions_200percent' panicked at crates/resvg/tests/integration/render.rs:354:75:
assertion `left == right` failed
  left: 3600
 right: 0

---- render::paint_servers_pattern_attributes_via_xlink_href stdout ----
thread 'render::paint_servers_pattern_attributes_via_xlink_href' panicked at crates/resvg/tests/integration/render.rs:527:64:
assertion `left == right` failed
  left: 5
 right: 0

---- render::paint_servers_pattern_everything_via_xlink_href stdout ----
thread 'render::paint_servers_pattern_everything_via_xlink_href' panicked at crates/resvg/tests/integration/render.rs:531:64:
assertion `left == right` failed
  left: 5
 right: 0

---- render::paint_servers_pattern_children_via_xlink_href stdout ----
thread 'render::paint_servers_pattern_children_via_xlink_href' panicked at crates/resvg/tests/integration/render.rs:529:62:
assertion `left == right` failed
  left: 5
 right: 0

---- render::painting_marker_marker_on_circle stdout ----
thread 'render::painting_marker_marker_on_circle' panicked at crates/resvg/tests/integration/render.rs:728:49:
assertion `left == right` failed
  left: 7
 right: 0

---- render::painting_marker_on_ArcTo stdout ----
thread 'render::painting_marker_on_ArcTo' panicked at crates/resvg/tests/integration/render.rs:738:41:
assertion `left == right` failed
  left: 1
 right: 0

---- render::painting_marker_orient_eq_30 stdout ----
thread 'render::painting_marker_orient_eq_30' panicked at crates/resvg/tests/integration/render.rs:745:45:
assertion `left == right` failed
  left: 3
 right: 0

---- render::shapes_circle_missing_cx_attribute stdout ----
thread 'render::shapes_circle_missing_cx_attribute' panicked at crates/resvg/tests/integration/render.rs:924:51:
assertion `left == right` failed
  left: 33
 right: 0

---- render::shapes_ellipse_missing_cx_attribute stdout ----
thread 'render::shapes_ellipse_missing_cx_attribute' panicked at crates/resvg/tests/integration/render.rs:930:52:
assertion `left == right` failed
  left: 31
 right: 0

---- render::structure_transform_origin_on_text_path stdout ----
thread 'render::structure_transform_origin_on_text_path' panicked at crates/resvg/tests/integration/render.rs:1248:56:
assertion `left == right` failed
  left: 1
 right: 0

---- render::text_text_anchor_on_tspan_with_arabic stdout ----
thread 'render::text_text_anchor_on_tspan_with_arabic' panicked at crates/usvg/src/writer/text.rs:227:38:
range end index 18 out of range for slice of length 17

---- render::text_text_complex_grapheme_split_by_tspan stdout ----
thread 'render::text_text_complex_grapheme_split_by_tspan' panicked at crates/usvg/src/writer/text.rs:227:38:
range end index 3 out of range for slice of length 2

---- render::text_tspan_bidi_reordering stdout ----
thread 'render::text_tspan_bidi_reordering' panicked at crates/usvg/src/writer/text.rs:227:38:
range end index 15 out of range for slice of length 12


failures:
    render::filters_feImage_svg
    render::filters_filter_functions_color_adjust_functions_2
    render::filters_filter_functions_color_adjust_functions_200percent
    render::paint_servers_pattern_attributes_via_xlink_href
    render::paint_servers_pattern_children_via_xlink_href
    render::paint_servers_pattern_everything_via_xlink_href
    render::painting_marker_marker_on_circle
    render::painting_marker_on_ArcTo
    render::painting_marker_orient_eq_30
    render::shapes_circle_missing_cx_attribute
    render::shapes_ellipse_missing_cx_attribute
    render::structure_transform_origin_on_text_path
    render::text_text_anchor_on_tspan_with_arabic
    render::text_text_complex_grapheme_split_by_tspan
    render::text_tspan_bidi_reordering

test result: FAILED. 1640 passed; 15 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.43s
```

- Most of these are differ only in a few pixels, these are most likely a result of float inaccuracies (for example since rotations are not represented accurately), so we can pretty much ignore them since there isn't much we can do about that.
- For the `filters` one, for some reason the color of the `saturate` one looks different. I have no idea why.
Input:
```svg
     font-family="Noto Sans" font-size="8">
    <title>Color adjust functions, 2 (SVG 2)</title>
    <!-- column 3 -->
    <rect id="rect7" x="140" y="20" width="40" height="40" fill="seagreen"
          filter="saturate(2)"/>
    <text id="text7" x="140" y="68">saturate</text>
</svg>
```

Output:
```svg
<svg width="200" height="200" viewBox="0 0 200 200" xmlns="http://www.w3.org/2000/svg">
    <defs>
        <filter id="f1" x="-0.1" y="-0.1" width="1.2" height="1.2">
            <feColorMatrix color-interpolation-filters="sRGB" in="SourceGraphic" result="result" type="saturate" values="2"/>
        </filter>
    </defs>
    <g id="g2" filter="url(#f1)">
        <path id="p1" fill="#2e8b57" stroke="none" d="M 140 20 L 180 20 L 180 60 L 140 60 Z"/>
    </g>
    <text id="t1" xml:space="preserve" text-rendering="optimizeLegibility"><tspan><tspan font-family="Noto Sans" font-size="8" fill="#000000" stroke="none" x="140" y="68">saturate</tspan></tspan></text>
</svg>
``` 
But for some reason, when rendering them with resvg the color looks different.

- Three text tests panic. See below for more information.

# Discussion points
This PR is not done yet because there are a couple of things I'm unsure about that probably need to be adapted, so I'm marking this as draft. I would appreciate if you could give me some pointers on those. In particular:

1) As mentioned above, three text test cases panic. My assumption was that the `positions/rotate` contain one element for each Unicode codepoint, that's why I used the `chars().count()` method to calculate the start/end:
 https://github.com/LaurenzV/resvg/blob/a5f07d1088dd1e886b6fafbe4c9b8c9108c684b2/crates/usvg/src/writer/text.rs#L146-L148
That seems to work fine for nearly all test cases, but looks like the assumption is not true for some Arabic text. I know text processing is complex so I had a hunch that I would encounter trouble with this, but since I don't know much about that, I would appreciate any pointers you can give me on how to correctly approach this.

2) It seems like `write_text_fmt` in `xml_writer` intentionally only escapes `<` but not `&`, which messes with text that contains such characters. Is there a reason for that? Is my approach of escaping it on the `usvg` side correct? (https://github.com/LaurenzV/resvg/blob/b4b9f9fe633e07b41e7ea5130c7fed27e1aa5440/crates/usvg/src/writer/text.rs#L152)

3) I encountered a fundamental problem with IDs while adding support for `text` elements. Firstly, because of the `text_to_user_space_on_use` method (https://github.com/RazrFalcon/resvg/blob/9101f30a70398cf7acd6def11f0743d3fc69757e/crates/usvg-text-layout/src/lib.rs#L824), not all paint servers might have an id, which is needed if we want to write them again. And secondly, in order to write the referenced paths of `textPath`elements, I needed to generate new IDs for them so that I can reference them (since the ID is not preserved in usvg). But this is obviously problematic because I don't know which IDs already exist in the document. 
This is why my idea was that whenever we call the `to_string` method, we create a deep copy (I'm not sure if I did this correctly though? If I see it right, calling `make_deep_copy` will only copy the node itself, but not the descendants?) of it and remap all of the IDs that occur in there. We also store what the last ID generated for each type of element was. This way, I can easily generate new IDs on the fly when needed. This obviously comes with a performance cost, and as I mentioned I believe that it currently might not work completely correctly, so not sure if this is the best approach. If you have other ideas, I'm all open for that.

4) I still need to update `docs/spec.adoc`, but I'll do that in the very end.


RazrFalcon:
>Instead of having a single writer.rs file with all writing operations, I split them up into different files to make it more easy to navigate the code.

No need to. One file is fine. As long as we're below 2000 LOC per file I'm fine.

>I added the ability to directly write text using the usvg CLI instead of having to convert text into paths beforehand.

The reason it wasn't supported before is because it's really hard to do losslessly, as I'm sure you have noticed. 
Also, it has to be heavily tested, which you're not doing right now. We need dedicated `usvg` tests for this. SVG to SVG one, but otherwise similar to resvg one. Required SVG files should be copied, not linked, to the usvg test suite.

>For the filters one, for some reason the color of the saturate one looks different. I have no idea why.
Input

Ignore it for now. As I've said before, usvg CLI output isn't tested right now. It is expected to be broken. Adding a usvg test suite is in my TODO list.

>My assumption was that the positions/rotate contain one element for each Unicode codepoint, that's why I used the chars().count() method to calculate the start/end

Kinda... If you look how `usvg-text-layout` works, you will notice that we use clusters for indexing, not Unicode codepoints. And clusters are graphemes, which can contain multiple codepoints.
Which in itself is technically incorrect, because we have to use UTF-16 and not UTF-32 or clusters for `x`, `y`, `dx`, `dy` and `rotate`.
Maybe [this](https://razrfalcon.github.io/notes-on-svg-parsing/text/character.html) would clarify things a bit.

In short, I would suggest using `chars().count()` + checked indexing everywhere for now and ignore edge cases. They can be fixed later.
I genuinely have no idea what to do here myself.

>It seems like write_text_fmt in xml_writer intentionally only escapes < but not &

Yes. For now escape it on the `usvg` side.

>I encountered a fundamental problem with IDs while adding support for text elements.

Yeah... Not sure what we should do here. Cloning `usvg::Tree` before writing is no-go. Not only it would be really hard due to `Rc` and shared everything. But also would be extremely expensive.
I think this should be somehow handled by `usvg-text-layout`. We already have `usvg_parser::Cache` which does it during parsing. But there is no way to pass it to `usvg-text-layout`.

There real fundamental problem here is that usvg/resvg do not care about IDs beyond the initial parsing phase. They rely on `Rc` pointers, not IDs. There are many places where IDs are omitted and as long as resvg is fine - we're fine as well.
The only reason usvg preserves IDs to begin with is to allow query methods, but it's barely working as well.

For now, I would suggest caching written IDs on the `usvg/writer` side. Basically remember IDs for each paint server and then generate a new one when there is one without it.

PS: As I said before multiple times: usvg CLI is just a debugging tool. It doesn't really "work". I'm aware of it. And it's pretty hard to make it functional.
Rendering SVG alone is an insanely hard task. I simply have no time to support everything else.

LaurenzV:
> No need to. One file is fine. As long as we're below 2000 LOC per file I'm fine.

Okay, so should I merge everything back together? I mean now I've split it up already, but if it's important for you to have everything in one file I'll revert this change. 

> The reason it wasn't supported before is because it's really hard to do losslessly, as I'm sure you have noticed.
Also, it has to be heavily tested, which you're not doing right now. We need dedicated usvg tests for this. SVG to SVG one, but otherwise similar to resvg one. Required SVG files should be copied, not linked, to the usvg test suite.

Yeah, sure. My goal with this PR isn't to add a test suite for usvg. Using the resvg test suite simply was a way for me to check whether it works _in general_, and this seems to be the case for now, apart from a few edge cases. 

> Ignore it for now. As I've said before, usvg CLI output isn't tested right now. It is expected to be broken. Adding a usvg test suite is in my TODO list.

Okay!

> Kinda... If you look how usvg-text-layout works, you will notice that we use clusters for indexing, not Unicode codepoints. And clusters are graphemes, which can contain multiple codepoints.
Which in itself is technically incorrect, because we have to use UTF-16 and not UTF-32 or clusters for x, y, dx, dy and rotate.
Maybe [this](https://razrfalcon.github.io/notes-on-svg-parsing/text/character.html) would clarify things a bit.
In short, I would suggest using chars().count() + checked indexing everywhere for now and ignore edge cases. They can be fixed later.
I genuinely have no idea what to do here myself.

Alright, I will do that for now.

> Yes. For now escape it on the usvg side.
The xmlwriter crate should be updated, but... you know, the same old story - I have zero free time.

Okay!

> For now, I would suggest caching written IDs on the usvg/writer side. Basically remember IDs for each paint server and then generate a new one when there is one without it.

Makes sense, I'll look into it.

> PS: As I said before multiple times: usvg CLI is just a debugging tool. It doesn't really "work". I'm aware of it. And it's pretty hard to make it functional.
Rendering SVG alone is an insanely hard task. I simply have no time to support everything else.

Well, considering that only 15 test cases fail (out of which a few are just pixel differences), it seems to work pretty well I think! :D

But if you feel like this PR would take up too much of your time, that's okay for me, I can close it. The main reason I made it is that I need the ability to preserve text in SVG for my own purposes, so I thought I might as well upstream it now that it's implemented and seems to work pretty well. But if you feel like it will take up too much of your time, I completely get it!

RazrFalcon:
>Okay, so should I merge everything back together?

Yes, just leave it as it was. And in general do not mix changes. It's hard to review otherwise. 

>But if you feel like this PR would take up too much of your time, that's okay for me, I can close it.

No, the PR is fine. By lack of time I've meant if I would write it myself.


Also, I haven't looked at the code yet, mainly due refactoring, but why do you need to calculate text length in Unicode codepoints to begin with? Afaik you can write `x`, `y`, `dx`, `dy` and `rotate` as is. They are already resolved.
You should technically dump the `usvg_tree::Text` structure as is. No post-processing needed. No?

LaurenzV:
> Also, I haven't looked at the code yet, mainly due refactoring, but why do you need to calculate text length in Unicode codepoints to begin with? Afaik you can write x, y, dx, dy and rotate as is. They are already resolved.
You should technically dump the usvg_tree::Text structure as is. No post-processing needed. No?

I think that didn't work for some reason, but I will try again. Maybe I just missed something.

crates/usvg/src/writer.rs
```diff
@@ -966,6 +1093,22 @@ fn has_xlink(node: &Node) -> bool {
...
+                    .chunks
+                    .iter()
+                    .map(|chunk| {
+                        if let TextFlow::Path(_) = chunk.text_flow {
+                            true
+                        } else {
+                            false
+                        }
+                    })
+                    .any(|b| b)
```
RazrFalcon, line 1107: Isn't `map` + `any` is just `find`?

crates/usvg/src/writer.rs
```diff
@@ -966,6 +1093,22 @@ fn has_xlink(node: &Node) -> bool {
             NodeKind::Image(_) => {
                 return true;
             }
+            NodeKind::Text(ref text) => {
+                if text
+                    .chunks
+                    .iter()
+                    .map(|chunk| {
+                        if let TextFlow::Path(_) = chunk.text_flow {
```
RazrFalcon, line 1101: You can use `matches!(chunk.text_flow, TextFlow::Path(_))` instead.

crates/usvg/src/writer.rs
```diff
@@ -668,12 +709,98 @@ fn conv_element(node: &Node, is_clip_path: bool, opt: &XmlOptions, xml: &mut Xml
...
-        NodeKind::Text(_) => {
-            log::warn!("Text must be converted into paths.");
+        NodeKind::Text(ref text) => {
+            xml.start_svg_element(EId::Text);
+
+            if !text.id.is_empty() {
+                xml.write_id_attribute(&text.id, writer_context.opt);
+            }
+
+            xml.write_attribute_raw("xml:space", |buf| {
```
RazrFalcon, line 723: Isn't this just `xml.write_attribute_raw("xml:space",  "preserve")`?

PS: yes, you got it right, usvg text should always be marked as `preserve` and no new lines must be written.

crates/usvg/src/writer.rs
```diff
@@ -1228,3 +1369,274 @@ fn write_num(num: f32, buf: &mut Vec<u8>, precision: u8) {
...
+    }
+}
+
+fn write_paint_order(xml: &mut XmlWriter, span: &TextSpan) {
+    if span.paint_order == PaintOrder::StrokeAndFill {
+        xml.write_svg_attribute(AId::PaintOrder, "stroke fill");
+    }
+}
+
+fn write_smallcaps(xml: &mut XmlWriter, span: &TextSpan) {
```
RazrFalcon, line 1435: I'm not sure we need such small functions, especially when they will be called only once. It only bloats the code.

crates/usvg/src/writer.rs
```diff
@@ -1228,3 +1369,274 @@ fn write_num(num: f32, buf: &mut Vec<u8>, precision: u8) {
...
+fn write_baseline_shift(xml: &mut XmlWriter, baseline_shift: &BaselineShift) {
+    match baseline_shift {
+        BaselineShift::Number(num) => xml.write_svg_attribute(AId::BaselineShift, num),
+        BaselineShift::Baseline => xml.write_svg_attribute(AId::BaselineShift, "baseline"),
+        BaselineShift::Subscript => xml.write_svg_attribute(AId::BaselineShift, "sub"),
+        BaselineShift::Superscript => xml.write_svg_attribute(AId::BaselineShift, "super"),
+    }
+}
+
+fn write_rotations(xml: &mut XmlWriter, text: &Text, actual_start: usize, actual_end: usize) {
```
RazrFalcon, line 1474: Rotations should be written as is. No post-processing is required. Parser already resolved everything, afaik.

crates/usvg/src/writer.rs
```diff
@@ -1228,3 +1369,274 @@ fn write_num(num: f32, buf: &mut Vec<u8>, precision: u8) {
...
+fn write_letter_spacing(xml: &mut XmlWriter, span: &TextSpan) {
+    if span.letter_spacing != 0.0 {
+        xml.write_svg_attribute(AId::LetterSpacing, &span.letter_spacing);
+    }
+}
+
+fn write_baseline_shift(xml: &mut XmlWriter, baseline_shift: &BaselineShift) {
+    match baseline_shift {
+        BaselineShift::Number(num) => xml.write_svg_attribute(AId::BaselineShift, num),
+        BaselineShift::Baseline => xml.write_svg_attribute(AId::BaselineShift, "baseline"),
```
RazrFalcon, line 1468: There is no `baseline` in SVG/CSS. It should be ignored completely in this case. `BaselineShift::Baseline` is basically `None`. But it's simpler to have `BaselineShift::Baseline` compared to `Option<BaselineShift>`.

LaurenzV:
> Also, I haven't looked at the code yet, mainly due refactoring, but why do you need to calculate text length in Unicode codepoints to begin with? Afaik you can write x, y, dx, dy and rotate as is. They are already resolved.
You should technically dump the usvg_tree::Text structure as is. No post-processing needed. No?

I'm not quite sure I get how exactly you mean to dump it. Let's say we have the following SVG:
```svg
     font-family="Noto Sans" font-size="32">
    <title>On `tspan`</title>

    <text id="text1" x="100" y="60">
        Start
        <tspan id="tspan1" x="100" y="100">middle</tspan>
    </text>

    <!-- image frame -->
</svg>
```

And the converted result (without x/y) looks like this:
```svg
<svg width="200" height="200" viewBox="0 0 200 200" xmlns="http://www.w3.org/2000/svg">
    <defs/>
    <path id="crosshair" fill="#000000" stroke="#808080" stroke-width="0.5" d="M 20 100 L 180 100 M 100 20 L 100 180"/>
    <text id="text1" xml:space="preserve" text-rendering="optimizeLegibility"><tspan><tspan font-family="Noto Sans" font-size="32" fill="#000000" stroke="none">Start </tspan></tspan><tspan><tspan font-family="Noto Sans" font-size="32" fill="#000000" stroke="none">middle</tspan></tspan></text>
    <path id="frame" fill="none" stroke="#000000" d="M 1 1 L 199 1 L 199 199 L 1 199 Z"/>
</svg>
 ```
Where/how would you define the x/y attributes so that the end result is the same?

**Comment by RazrFalcon** [comment]
Oh, I forgot that we merge the string into one... But you should simply use `usvg::TextSpan::start` + `usvg::TextSpan::end`. No need to count it manually.
You should still use checked indexing everywhere, just in case.

But `rotate` specifically can be set only on `text` and should be written as is.

As for `usvg::TextChunk` - it should be written as a `tspan` I guess, without actually writing the text. Aka:

```xml
<text> <!-- usvg::Text -->
    <tspan> <!-- usvg::TextChunk -->
        <tspan> <!-- usvg::TextSpan -->
        </tspan>
    </tspan>
</text>
```

LaurenzV:
You are right, for rotate this does seem to work! But for x/y it's not as easy I think.

LaurenzV:
> Oh, I forgot that we merge the string into one... But you should simply use usvg::TextSpan::start + usvg::TextSpan::end. No need to count it manually.

How, though? As far as I can tell, the start/end are reset for every TextChunk, but the x/y positions are global over all text chunks. For example, the following SVG:
```svg
     font-family="Noto Sans" font-size="64">
    <title>`x` and `y` with `dx` and `dy` lists</title>

    <text id="text1" x="30 40" dx="10 40" y="100 100" dy="5 20">Text</text>

    <!-- image frame -->
</svg>
```

The text element looks somewhat like this (irrelevant parts cut off). Note that for the first chunk, the range is 0-1 and for the second chunk it's 0-3 instead of 1-4. So in order to access the right element in the `positions` array, I need to count all characters from the previous text chunks. Am I missing something?:
```
Text {
    id: "text1",
    rendering_mode: OptimizeLegibility,
    positions: [
        CharacterPosition {
            x: Some(
                30.0,
            ),
            y: Some(
                100.0,
            ),
            dx: Some(
                10.0,
            ),
            dy: Some(
                5.0,
            ),
        },
        CharacterPosition {
            x: Some(
                40.0,
            ),
            y: Some(
                100.0,
            ),
            dx: Some(
                40.0,
            ),
            dy: Some(
                20.0,
            ),
        },
        CharacterPosition {
            x: None,
            y: None,
            dx: None,
            dy: None,
        },
        CharacterPosition {
            x: None,
            y: None,
            dx: None,
            dy: None,
        },
    ],
    rotate: [
        0.0,
        0.0,
        0.0,
        0.0,
    ],
    writing_mode: LeftToRight,
    chunks: [
        TextChunk {
            x: Some(
                30.0,
            ),
            y: Some(
                100.0,
            ),
            anchor: Start,
            spans: [
                TextSpan {
                    start: 0,
                    end: 1,
                },
            ],
            text_flow: Linear,
            text: "T",
        },
        TextChunk {
            x: Some(
                40.0,
            ),
            y: Some(
                100.0,
            ),
            anchor: Start,
            spans: [
                TextSpan {
                    start: 0,
                    end: 3,
                    // Other attributes
                },
            ],
            text_flow: Linear,
            text: "ext",
        },
    ],
}

> But rotate specifically can be set only on text and should be written as is.

That seems to work indeed, thanks!

**Comment by LaurenzV** [comment]
Alright, so:
- I applied the changes from your code review. Array index is now checked to prevent panic.
- I changed the way I deal with the IDs for text paths. Instead of remapping the whole tree, I basically just traverse the tree to get all existing IDs. And then whenever I need a new ID, I generate a new one (and it will make sure that each new ID is unique).

What's still missing:
- Revert changes from resvg test renderer once no other changes are needed.
- Update the docs.
- Resolve the discussion regarding the span offset, but as I've mentioned above there is no way around counting it manually?
- (Apply any other changes you suggest).

**Comment by RazrFalcon** [comment]
Well, I guess I do not remember how my own library works... SVG text is absurdly complex.
For one, `CharacterPosition::x` and `CharacterPosition::y` are not used outside the `usvg-parser`. We don't have to keep them public. The parser code should be restructured a bit. I will look into it later.
For now, you should ignore those properties completely. Write the one set on chunks.

As for dx/dy, I guess you can dump them as is on the root `text` element, same as `rotate`. Hopefully this would help.

Basically, you should use only `TextSpan::start` and `TextSpan::end` when you want to slice a string. Manually counting chars would always be incorrect. And usvg already done it for you.

Again, there are reasons why text writing wasn't implement. It's really hard to do. SVG text basically the hardest part of SVG in general.

Also, the SVG output is probably incorrect for complex scripts (like Arabic) anyway, because resvg uses UTF-32 for indexing instead of UTF-16. So rendering with resvg works, but it doesn't mean rendering with other SVG libraries would work as well.
That's why usvg test suite should use SVG-to-SVG tests and not resvg rendering tests.

>Update the docs.

I think I will do it myself. There are a lot of work.

File: crates/usvg/src/lib.rs [hunk 7]
```diff
@@ -69,6 +81,125 @@ pub trait TreeWriting {
...
+}
+
+fn get_ids(tree: &Tree) -> HashSet<String> {
+    let mut ids = HashSet::new();
+    get_ids_impl(&tree.root, &mut ids);
+    ids.remove("");
+    ids
+}
+
+fn get_ids_impl(node: &Node, ids: &mut HashSet<String>) {
```
RazrFalcon, line 143: Ugh... I guess I would have to revert to the old tree representation, the one with defs. This way there would be no need for such ugly code.
But that approach has it's own downsides as well.

crates/usvg/src/lib.rs
```diff
@@ -69,6 +81,125 @@ pub trait TreeWriting {
...
+
+        while ids.contains(&id) {
+            id = bump();
+        }
+
+        id
+    }
+}
+
+fn get_ids(tree: &Tree) -> HashSet<String> {
```
RazrFalcon, line 136: I would preferrer `collect_ids`

crates/usvg/src/lib.rs
```diff
@@ -69,6 +81,125 @@ pub trait TreeWriting {
...
+    }
+}
+
+/// A generator for new IDs.
+#[derive(Default)]
+pub(crate) struct IdGenerator {
+    /// A set of IDs that should never be generated by the generator.
+    ids: HashSet<String>,
+    /// The counter for the current path ID.
+    path: u64,
```
RazrFalcon, line 104: Why paths? Aren't you generating paint servers ids?

crates/usvg/src/writer.rs
```diff
@@ -1228,3 +1395,173 @@ fn write_num(num: f32, buf: &mut Vec<u8>, precision: u8) {
...
+    if span.small_caps {
+        xml.write_svg_attribute(AId::FontVariant, "small-caps");
+    }
+
+    if span.paint_order == PaintOrder::StrokeAndFill {
+        xml.write_svg_attribute(AId::PaintOrder, "stroke fill");
+    }
+
+    if !span.apply_kerning {
+        xml.write_attribute_raw("style", |buf| {
```
RazrFalcon, line 1456: You can use `xml.write_attribute("style", "font-kerning:none")` directly.

crates/usvg/src/writer.rs
```diff
@@ -1228,3 +1395,173 @@ fn write_num(num: f32, buf: &mut Vec<u8>, precision: u8) {
...
+            buf.extend_from_slice("font-kerning:none".as_bytes())
+        });
+    }
+
+    match span.dominant_baseline {
```
RazrFalcon, line 1461: Better check for Auto first and then `match` can returns just a string. Kinda like:

```rust
let name = match span.dominant_baseline {
    DominantBaseline::UseScript => "use-script"
    // ...
};
xml.write_svg_attribute(AId::DominantBaseline, name);
```

Same with `AlignmentBaseline`. 

This would make the code much cleaner.

crates/usvg/src/lib.rs
LaurenzV: Right now I'm only generating it for the paths of `textPaths`. Paint server IDs are currently not generated anymore.

This means that gradients with object bounding box will still work if text is not converted to paths (since then the `paint_server_to_user_space_on_use` is never called, so the IDs from the paint servers are preserved). But when converting it to paths, it won't work anymore (it didn't work before either, though).

I tried reimplemeting it for paint servers using the same approach as for text paths, but the problem is I can't use the same approach that easily because the order in which we process the paint servers might not be the same. So for example, if I use `loop_over_paint_servers` to find all paint servers without an ID it will loop over paths/texts as well as corresponding underlines/overlines in a specific order, and if I push the new IDs into a vector just like for text paths, I need to make sure that when writing the paint servers I use the exact same order which requires some more thinking on how to do it, so I ignored it for now.

crates/usvg/src/writer.rs
LaurenzV: So something like this? Is there a way to avoid the final unreachable?
```rust
if span.dominant_baseline != DominantBaseline::Auto {
        let name = match span.dominant_baseline {
            DominantBaseline::UseScript => "use-script",
            ...
            DominantBaseline::Auto => unreachable!()
        };
        xml.write_svg_attribute(AId::DominantBaseline, name);
    }
```

Or should I remove the if statement and just return "auto" for Auto?

LaurenzV:
> For one, CharacterPosition::x and CharacterPosition::y are not used outside the usvg-parser. We don't have to keep them public. The parser code should be restructured a bit. I will look into it later.

Ahh, makes sense... Yeah, I actually was wondering about why this is "duplicated".

> As for dx/dy, I guess you can dump them as is on the root text element, same as rotate. Hopefully this would help.

It does!

> Again, there are reasons why text writing wasn't implement. It's really hard to do. SVG text basically the hardest part of SVG in general.

Yeah, got it. That's also why I made the option to preserve texts opt-in instead of opt-out, so the default will still be converting it to paths.

> I think I will do it myself. There are a lot of work.

Okay!

crates/usvg/src/lib.rs
RazrFalcon: >Right now I'm only generating it for the paths of textPaths. 

Aren't they always have an ID?

>the paint servers I use the exact same order

Why do you care about the order in the first place? All you have to do is to have `HashMap<Node, String>`, no?

crates/usvg/src/writer.rs
RazrFalcon: Just keep the `unreachable!()`. It's fine.

crates/usvg/src/lib.rs
LaurenzV: > Aren't they always have an ID?

No, because text paths only store the data of the path they belong to, not their actual ID:

https://github.com/RazrFalcon/resvg/blob/9101f30a70398cf7acd6def11f0743d3fc69757e/crates/usvg-tree/src/text.rs#L240-L248

> Why do you care about the order in the first place? All you have to do is to have HashMap<Node, String>, no?

Well, for example the underline and overline of a text might have a different fill, right? But they belong to the same TextNode (not to mention all other tspans of that Text node who might have their own fills, so that wouldn't work I think?

RazrFalcon: >No, because text paths only store the data of the path they belong to, not their actual ID:

Meh... I guess this is because text layers were never preserved originally, so it was fine. And even if we preserve IDs, we still have to make sure to deduplicate those paths during writing. Because multiple text elements can reference the same path.

>Well, for example the underline and overline of a text might have a different fill, right?

Damn...

RazrFalcon:
Here is an idea. A year ago I've [changed](https://github.com/RazrFalcon/resvg/blob/master/CHANGELOG.md#changed-8) the `usvg::Tree` structure. Before it was just a tree of SVG nodes, including `svg` and `defs` nodes. This way `usvg` CLI was simply dumping a tree as is. `usvg` was guarantee to produce the correct tree structure. 

The problem with that approach is that the tree isn't really strongly typed, so to speak. Because there are no restrictions on the tree structure, unlike now. A path can have a gradient child for no reason, there could be multiple `svg` and `defs` nodes and so on.

The current `usvg::Tree` structure is more robust and easier to understand and query, but it has tons of downside as well.
For one, traversing the tree is pain (see `Node::subroots`), because it's not longer a tree, but a tree of trees.
Second, the current tree doesn't care about node IDs at all. It's just an `Rc`, while before `defs` were accessed by the ID/string. Slow, but guarantee to have properly resolved IDs.
Also, writing the new tree is pain, as I'm sure you have noticed. This was one of the reasons `usvg` CLI weren't available for a while.

So the solution would be to revert the old tree structure, except using `Rc` instead of `String` for defs referencing.
This way producing the correct tree with proper `defs` is the parser's job and the writer can simply dump it without caring much.
The downside is that generating a list of unique `defs` is kinda hard. We have to deduplicate them and remove an unused one. Which slows the parsing down a bit, which is not great when we're only rendering an SVG.
But it also makes debugging a bit easier, since `usvg` output is now 1:1 to the actual data in memory.
And all/most of the ID generation can/should be done during parsing now.
We can even make `defs` IDs to use `NonEmptyString` string or something, to guarantee that all IDs were set.

Basically, there is no silver bullet and both approaches have downsides, but I think the old behavior would actually be easier to use.
Now I just have to rewrite half of `usvg` before you can merge this patch... I have no idea how long it would take. Will try it out on the weekend.

LaurenzV:
Hmm, okay... Would that mean that the strong typedness would be lost again? I guess that would be kind of unfortunate as well.

But I mean, if having to completely rewrite `usvg` is required for this, I'm not sure if it's worth it, especially since you mentioned you don't have a lot of time I would feel a bit bad if you spend so much time on this just for this feature. :( 

As I mentioned, I can just use my own `resvg` fork for my personal use case, so it's not urgent or anything.

RazrFalcon:
The change boils down to having a dedicated `defs` node as the first child of the SVG. And this node would have all paint server, masks, etc. This way you can dump `usvg::Tree` as is without doing any post-processing and IDs resolving.
So it's not that major of a change. We're basically making `usvg::Tree` closer to the original SVG tree.

And this PR is just a yet another reminder that the current logic is bit flawed. I was planning to do this long ago anyway.

RazrFalcon:
There is no `usvg_tree::Text::positions` anymore, as we discussed above.

RazrFalcon:
Sorry for a delay. I did some experiments and it doesn't look that simple. 
I've already abandoned the idea of having a list of defs. Too much work with little benefits. Instead, we should focus on trying to guarantee that all IDs in the tree are unique, which is surprisingly hard.
We already have an ID generator that does this, but it is parser-only. Meaning that in `usvg-text-layout` we have to collect all IDs once again, which would probably be quite expensive.
Not to mention that a user can add a random element without any repercussions . Meaning that there is no way to guarantee that all IDs are unique during writing.

Therefore we have two options: 
1. Make the ID generator part of the `Tree`. This way each `Tree` would carry a list of its unique IDs. It still doesn't guarantee uniqueness, since each ID is a mutable string. But better than nothing.
2. Do not generate IDs during parsing to begin with. Preserve only the one existing in the SVG. And generate missing IDs only during writing.

The second option seems to be the more viable one. It would simplify the parser a bit and make it slightly faster, but the writer would have to do more work.
This is basically what you have already did, but more generic.
For starter, we have to add the `id` field to `TextPath` and move the ID generator from the parser to the writer. 
I think I figured out how to add my commits to this PR, so I will do the low level stuff on my own first.

LaurenzV:
Sounds good, keep me posted!

LaurenzV:
Cool, thanks! I'm a bit busy at the moment so not sure when I'll be able to keep working on it, might be a few days, but might also be a few weeks, but I'll definitely finish it.

RazrFalcon:
Sure, not problem. Take your time.

LaurenzV:
Nice! Some minor issues I still need to look into, but in general I like your new version! I will let you know as soon as it's ready for review.

Regarding the tests, is there anything in specific you want me to add? I will definitely add tests to make sure that text paths work properly, and I'll think about whether anything else comes to my mind. But I'd like to know what your "goal" for the usvg test suite is, should it eventually be similar to resvg in terms of comprehensiveness where very edge case is tested, or should its focus just be on quirks of usvg to make sure that edge cases work as expected? I guess the second would make more sense?

RazrFalcon:
Ideally, usvg tests should be as thorough as possible, but there isn't much to test, unlike resvg.
A common case + some edge cases are fine. For text you can borrow some resvg tests. We definitely want to test `rotate` attribute, `dx`/`dy` list attributes, complex `text-decoration`, text-on-path and so on. Not really edge cases, but rather the SVG Text complexity.

LaurenzV:
Alright. Will see what I can do!

LaurenzV:
Some remarks:

There is a small problem with the following test:

```svg
     font-family="Noto Sans" font-size="64">
    <title>Filter bounding box</title>
    <desc>Make sure filters use text bounding box and not path bounding box.</desc>

    <filter id="filter1" x="0" y="0" width="1" height="1">
        <feFlood flood-color="green"/>
    </filter>

    <text id="text1" x="100" y="100" text-anchor="middle" filter="url(#filter1)">Text</text>
    <text id="text2" x="100" y="100" text-anchor="middle">Text</text>

    <!-- image frame -->
</svg>
```

If I convert text to paths and rewrite it, the output will be the following:
```svg
<svg width="200" height="200" viewBox="0 0 200 200" xmlns="http://www.w3.org/2000/svg">
    <defs>
        <filter id="filter1" x="0" y="0" width="1" height="1">
            <feFlood color-interpolation-filters="linearRGB" flood-color="#008000" flood-opacity="1" result="result1"/>
        </filter>
    </defs>
    <path id="crosshair" fill="#000000" stroke="#808080" stroke-width="0.5" d="M 20 100 L 180 100 M 100 20 L 100 180"/>
    <g filter="url(#filter1)">
        <g id="text1">
            <path fill="#000000" stroke="none" d="M 59.231995 100 L 53.471996 100 L 53.471996 59.359993 L 39.199993 59.359993 L 39.199993 54.303993 L 73.439995 54.303993 L 73.439995 59.359993 L 59.231995 59.359993 L 59.231995 100 Z M 88.352 65.056 Q 92.768 65.056 95.936 66.976 Q 99.104 68.895996 100.8 72.384 Q 102.49599 75.872 102.49599 80.544 L 102.49599 83.936 L 79.007996 83.936 Q 79.135994 89.76 81.98399 92.8 Q 84.832 95.84 89.952 95.84 Q 93.216 95.84 95.744 95.232 Q 98.272 94.624 100.96 93.472 L 100.96 98.4 Q 98.336 99.552 95.776 100.096 Q 93.216 100.64 89.696 100.64 Q 84.832 100.64 81.088 98.656 Q 77.343994 96.672 75.26399 92.736 Q 73.184 88.8 73.184 83.104004 Q 73.184 77.472 75.072 73.44 Q 76.95999 69.408 80.383995 67.231995 Q 83.808 65.056 88.352 65.056 Z M 88.28799 69.664 Q 84.256 69.664 81.92 72.256 Q 79.584 74.848 79.135994 79.456 L 96.608 79.456 Q 96.544 75.104 94.56 72.384 Q 92.576 69.664 88.28799 69.664 Z M 118.048 82.464 L 106.208 65.696 L 112.608 65.696 L 121.44 78.624 L 130.208 65.696 L 136.544 65.696 L 124.70399 82.464 L 137.18399 100 L 130.784 100 L 121.44 86.304 L 111.968 100 L 105.632 100 L 118.048 82.464 Z M 155.232 96.032 Q 156.512 96.032 157.856 95.808 Q 159.2 95.584 160.032 95.328 L 160.032 99.616 Q 159.136 100.064 157.472 100.352 Q 155.808 100.64 154.272 100.64 Q 151.584 100.64 149.312 99.712 Q 147.04001 98.784 145.632 96.48 Q 144.224 94.176 144.224 90.016 L 144.224 70.048 L 139.36 70.048 L 139.36 67.36 L 144.288 65.119995 L 146.528 57.824 L 149.856 57.824 L 149.856 65.696 L 159.776 65.696 L 159.776 70.048 L 149.856 70.048 L 149.856 89.888 Q 149.856 93.024 151.36 94.528 Q 152.864 96.032 155.232 96.032 Z"/>
        </g>
    </g>
    <g id="text2">
        <path fill="#000000" stroke="none" d="M 59.231995 100 L 53.471996 100 L 53.471996 59.359993 L 39.199993 59.359993 L 39.199993 54.303993 L 73.439995 54.303993 L 73.439995 59.359993 L 59.231995 59.359993 L 59.231995 100 Z M 88.352 65.056 Q 92.768 65.056 95.936 66.976 Q 99.104 68.895996 100.8 72.384 Q 102.49599 75.872 102.49599 80.544 L 102.49599 83.936 L 79.007996 83.936 Q 79.135994 89.76 81.98399 92.8 Q 84.832 95.84 89.952 95.84 Q 93.216 95.84 95.744 95.232 Q 98.272 94.624 100.96 93.472 L 100.96 98.4 Q 98.336 99.552 95.776 100.096 Q 93.216 100.64 89.696 100.64 Q 84.832 100.64 81.088 98.656 Q 77.343994 96.672 75.26399 92.736 Q 73.184 88.8 73.184 83.104004 Q 73.184 77.472 75.072 73.44 Q 76.95999 69.408 80.383995 67.231995 Q 83.808 65.056 88.352 65.056 Z M 88.28799 69.664 Q 84.256 69.664 81.92 72.256 Q 79.584 74.848 79.135994 79.456 L 96.608 79.456 Q 96.544 75.104 94.56 72.384 Q 92.576 69.664 88.28799 69.664 Z M 118.048 82.464 L 106.208 65.696 L 112.608 65.696 L 121.44 78.624 L 130.208 65.696 L 136.544 65.696 L 124.70399 82.464 L 137.18399 100 L 130.784 100 L 121.44 86.304 L 111.968 100 L 105.632 100 L 118.048 82.464 Z M 155.232 96.032 Q 156.512 96.032 157.856 95.808 Q 159.2 95.584 160.032 95.328 L 160.032 99.616 Q 159.136 100.064 157.472 100.352 Q 155.808 100.64 154.272 100.64 Q 151.584 100.64 149.312 99.712 Q 147.04001 98.784 145.632 96.48 Q 144.224 94.176 144.224 90.016 L 144.224 70.048 L 139.36 70.048 L 139.36 67.36 L 144.288 65.119995 L 146.528 57.824 L 149.856 57.824 L 149.856 65.696 L 159.776 65.696 L 159.776 70.048 L 149.856 70.048 L 149.856 89.888 Q 149.856 93.024 151.36 94.528 Q 152.864 96.032 155.232 96.032 Z"/>
    </g>
    <path id="frame" fill="none" stroke="#000000" d="M 1 1 L 199 1 L 199 199 L 1 199 Z"/>
</svg>
```

Since now the filter will be applied to a path instead of the text itself, the bounding box is different:
test

vs.

test

The same applies to some other tests that rely on the text bbox. Do you know if there is anything we can do against that? Like a way to "force" a bigger bounding box? Or should we just ignore it for now? 

Next, there is a small problem with how we prepare the defs IDs. Consider the following SVG:
```svg
    <title>Recursive on child</title>

    <pattern id="patt1" patternUnits="userSpaceOnUse" width="40" height="25">
        <rect id="rect1" width="40" height="25" fill="url(#patt2)" stroke="green"/>
    </pattern>
    <pattern id="patt2" patternUnits="userSpaceOnUse" width="40" height="25">
        <rect id="rect2" width="40" height="25" fill="url(#patt1)" stroke="green"/>
    </pattern>

    <rect id="rect3" x="20" y="20" width="160" height="70" fill="url(#patt1)" stroke="darkblue"/>
    <rect id="rect4" x="20" y="110" width="160" height="70" fill="url(#patt2)" stroke="darkblue"/>

    <!-- image frame -->
</svg>
```

Currently we will process each def separately and generate its ID when it is its turn, but this is a problem if defs reference each other. For example for patterns we convert its elements first before generting the IDs of all other patterns, which is why the `get_def_ids` call will fail (which was caught thanks to the `debug_assert`statement):

https://github.com/RazrFalcon/resvg/blob/7e1f99c675ccb02c6c8418a8196a8a2af0709575/crates/usvg/src/writer.rs#L647-L665

I'll change it so that all `prepare_def_ids` call are made in the beginning, even if this will make the code uglier (because now we have to loop twice...)

And finally, for the following SVG:

```svg
     xmlns:xlink="http://www.w3.org/1999/xlink">
    <title>Link to an element with opacity</title>
    <desc>
        resvg specific test.
        When `opacity`, `filter`, `mask` and/or `clip-path` attributes are present,
        resvg will create a parent group and move them to it.
        This works fine for most cases, but in case of `feImage`
        we potentially can still point to the original element and not the group,
        which will lead to an invalid results.
    </desc>

    <defs id="defs1">
        <filter id="filter1">
            <feImage xlink:href="#rect3"/>
        </filter>
        <rect id="rect3" x="36" y="36" width="120" height="120" fill="green" opacity="0.5"/>
    </defs>
    <rect id="rect1" x="20" y="20" width="160" height="160" fill="red" filter="url(#filter1)"/>
    <rect id="rect2" x="40" y="40" width="120" height="120" fill="none" stroke="black"/>

    <!-- image frame -->
</svg>
```

It will be converted to 
```svg
<svg width="200" height="200" viewBox="0 0 200 200" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
    <defs>
        <g opacity="0.5">
            <path id="rect3" fill="#008000" stroke="none" d="M 36 36 L 156 36 L 156 156 L 36 156 Z"/>
        </g>
        <filter id="filter1" x="-0.1" y="-0.1" width="1.2" height="1.2">
            <feImage color-interpolation-filters="linearRGB" preserveAspectRatio="xMidYMid" image-rendering="optimizeQuality" xlink:href="#" result="result1"/>
        </filter>
    </defs>
    <g filter="url(#filter1)">
        <path id="rect1" fill="#ff0000" stroke="none" d="M 20 20 L 180 20 L 180 180 L 20 180 Z"/>
    </g>
    <path id="rect2" fill="none" stroke="#000000" d="M 40 40 L 160 40 L 160 160 L 40 160 Z"/>
    <path id="frame" fill="none" stroke="#000000" d="M 1 1 L 199 1 L 199 199 L 1 199 Z"/>
</svg>
```

As you can see, the ID of the group in defs is missing, I think it's because it's generated by usvg but it doesn't create a corresponding ID, and we don't create any either here:

https://github.com/RazrFalcon/resvg/blob/7e1f99c675ccb02c6c8418a8196a8a2af0709575/crates/usvg/src/writer.rs#L261-L266

I guess I can just add a check if it's a group, and if so generate a new ID for it (if not existing already).

LaurenzV:
I added a couple of tests, hope it is good now! I encountered the problem that since you already added tests where text would be converted to paths, I had no way of creating my tests since they would be converted as well, which would defeat the purpose. The way I solved it is that you need to add an explicit comment in the SVG to remark that text should be converted to paths, I hope this is okay.

crates/usvg/tests/write.rs
```diff
@@ -112,3 +119,33 @@ fn clip_path_with_complex_text() {
 fn text_with_generated_gradients() {
     resave("text-with-generated-gradients");
 }
+
+#[test]
+fn text_on_path() {
```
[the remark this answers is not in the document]
RazrFalcon: This way the old tests would not be affected at all.

And you can add `preserve-text-` prefix for those test files.

crates/usvg/src/lib.rs
```diff
@@ -49,7 +49,6 @@ and can focus just on the rendering part.
 
 #![forbid(unsafe_code)]
 #![warn(missing_docs)]
-#![warn(missing_debug_implementations)]
```
RazrFalcon, line 52: Why?

crates/usvg/src/main.rs
```diff
@@ -426,7 +428,9 @@ fn process(args: Args) -> Result<(), String> {
     }?;
 
     let mut tree = usvg_tree::Tree::from_data(&input_svg, &re_opt).map_err(|e| format!("{}", e))?;
-    tree.convert_text(&fontdb);
+    if !args.preserve_text {
+        tree.convert_text(&fontdb);
```
RazrFalcon, line 432: No need to load fontdb in this case either.

crates/usvg/src/writer.rs
```diff
@@ -1438,3 +1593,122 @@ fn write_num(num: f32, buf: &mut Vec<u8>, precision: u8) {
...
+
+    xml.write_svg_attribute(AId::FontFamily, &span.font.families.join(", "));
+
+    match span.font.style {
+        FontStyle::Normal => {}
+        FontStyle::Italic => xml.write_svg_attribute(AId::FontStyle, "italic"),
+        FontStyle::Oblique => xml.write_svg_attribute(AId::FontStyle, "oblique"),
+    }
+
+    if span.font.weight != 400 {
```
RazrFalcon, line 1615: `span.font.stretch` is missing.

RazrFalcon:
Very good. But text wouldn't give up that easily.

crates/usvg/src/writer.rs
```diff
@@ -863,7 +898,117 @@ fn conv_element(node: &Node, is_clip_path: bool, ctx: &mut WriterContext, xml: &
...
+                }
+
+                xml.write_attribute("xml:space", "preserve");
+
+                match text.writing_mode {
+                    WritingMode::LeftToRight => {}
+                    WritingMode::TopToBottom => xml.write_svg_attribute(AId::WritingMode, "tb"),
+                }
+
+                match text.rendering_mode {
```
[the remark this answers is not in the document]
LaurenzV: You sure? At least [here](https://developer.mozilla.org/en-US/docs/Web/CSS/text-rendering) it says it should be `auto`? Not sure though.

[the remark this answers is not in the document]
LaurenzV: You sure? At least [here](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/text-rendering) it says it should be `auto`? Although it does mention that in this case legibility should be preferred.

crates/usvg/src/writer.rs
```diff
@@ -722,6 +733,30 @@ fn conv_defs(tree: &Tree, ctx: &mut WriterContext, xml: &mut XmlWriter) {
 
         xml.end_element();
     }
+
+    if tree.has_text_nodes() {
+        // TODO: doesn't check for text in patterns and masks...
```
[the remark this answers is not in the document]
LaurenzV: I can add the implementation, but we still have the problem I outlined above, namely that having something that requires a def (path of text path here) nested inside of another def will lead to a panic.

crates/usvg/src/writer.rs
```diff
@@ -1438,3 +1593,122 @@ fn write_num(num: f32, buf: &mut Vec<u8>, precision: u8) {
...
+fn write_span(
+    is_clip_path: bool,
+    ctx: &mut WriterContext,
+    xml: &mut XmlWriter,
+    chunk: &TextChunk,
+    span: &TextSpan,
+) {
+    xml.start_svg_element(EId::Tspan);
+
+    xml.write_svg_attribute(AId::FontFamily, &span.font.families.join(", "));
```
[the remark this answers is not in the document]
LaurenzV: I added this and also added a test.

crates/usvg/src/writer.rs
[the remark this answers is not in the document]
RazrFalcon: Hmm... kinda. We do not represent `auto` at all, because it wasn't needed. 
`optimizeLegibility` is de-facto the default value, so let's just ignore it. Most libraries do not support this property to begin with, so it's fine.

And maybe later we add a proper `auto` support.

crates/usvg/src/main.rs
RazrFalcon: `--list-font` shouldn't be used with input svg to begin with. A known bug. 
Let's leave it for later. The current implementation is fine.

crates/usvg/src/writer.rs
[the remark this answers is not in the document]
RazrFalcon: Oh right... we already wrote them... 
Remove the TODO then. My bad.

But we should have tests for text in clipPath, mask and pattern. And maybe feImage. If they fail, simply comment them out for now and I will figure out what to do with them later.

crates/usvg/src/writer.rs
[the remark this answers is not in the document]
RazrFalcon: The current `font-family` parser is trash. Known't issue. Please open a new issue with an SVG file sample.

RazrFalcon:
Looks really good now. Thanks!

And as the final touch, let's rename `preserve-text-text-simple-case` to just `preserve-text-simple-case`. Same with other tests. No need for `text-text` duplication.

And I will test it more thoroughly after the merge.

LaurenzV:
Cool, will try to address the remaining points in the evening.

LaurenzV:
I renamed them and added tests for text in patterns, clip paths as well as masks, and also a test for feImage.

RazrFalcon:
All good! Thanks again.

I will take a look into nested defs in the next couple of weeks and we're ready for the release.

LaurenzV:
Cool, thanks! And sorry that this ended up taking much longer.

RazrFalcon:
No-no, all good. Most downtime was because of me.

RazrFalcon:
Fixed usvg tests and published a new release. 

usvg defs writing is still a mess. We have to rewrite it. Ideally, we should collect all defs first, sort them from less used to most used and only then write them. But here are two problems:

1. How to actually sort them. We somehow have to make sure that each defs child node has all of its references before it. So it's not just the number of references. If a pattern references a clip path, we must put this clip path before pattern, somehow.
2. We have to prevent/detect recursive references. Technically, `usvg-parser` should do this for us, but it still can have bugs and no one stops you to create your own malformed tree.

Issue #521, opened 2022-07-06
Control over image fetching


kornelski:
I'd like to have more control over fetching and decoding of images referenced in SVG files, in order to use existing decoders that I have, and to enforce size and memory usage constraints.

Would you be interested in a PR that makes image decoding "pluggable"? I image it could be something like an `ImageDecoder` trait and the library could allow user to configure their own implementation.

RazrFalcon:
Sure, but can you clarify the use case? I mean, raster images are relatively rare in SVG. Therefore you're optimizing for an edge case.

Despite how it may look, I'm trying my best to simplify resvg. This is one of the reasons resvg is pure Rust.
So I don't really see a benefit from using custom raster image decoders, since they will only increase the complexity even more. Especially if you account building resvg as a C library. Linking would become a nightmare.

But again, as long as you will be able to introduce this feature with minimal changes - it might be ok. I have to see the code first.

PS: just in case, "pluggable" rendering is out of scope. resvg already had 4 backends and it's just not worth it. Or even technically possible, to begin with. Luckily, one can easy write their own "renderer" on top of usvg, as many do.

yisibl:
@RazrFalcon  We have a very wide range of use cases. For example, we use SVG for our ad templates, which contain a large number of images of products that are bitmaps.

![image](https://user-images.githubusercontent.com/2784308/177612967-6fae1165-1880-4a9f-9e40-0e5bcca0bb8d.png)

So, in [resvg-js v2.1.0](https://github.com/yisibl/resvg-js/releases/tag/v2.1.0), we forked resvg and made some [adjustments](https://github.com/zimond/resvg/commit/ab02d7e76481129b298b9223b0beec9e3d932186) to image loading to load `<image>` that contain http URLs. It would be nice if resvg could support this natively.  
@kornelski Are your needs and ideas similar to ours?

 ![load image URL Demo](https://user-images.githubusercontent.com/2784308/168537647-6787dfc3-49b6-42bb-be5e-35ddba8072d9.png)

TOGoS:
Hey there.  I, too, would like to use `resvg` to render SVGs that reference images by HTTP URL.  e.g.
```
	<image x="0" y="0" width="32" height="32"
		style="image-rendering: optimizeSpeed"
[...]

PR #689, opened 2023-12-24
Add support for more image formats

LaurenzV:
This PR adds support for more image formats, including WEBP, TIFF, and BMP (bitmap).

I ran some tests, and the following image formats embedded in SVG are supported by different viewers:
WEBP: Chrome, Firefox, Safari, Qt
TIFF: Safari, Inkscaepe, librsvg, Qt
BMP: Chrome, Firefox, Safari, Inkscape, librsvg, Qt
AVIF: Chrome, Firefox, Safari, Qt

I initially wanted to add all 4 of those, but I had trouble building image with the `avif-decode` feature flag, so I ended up leaving that one out. I'm not sure how common it is, anyway.

A few discussion points remain:
- One advantage of using the `image` crate is that we don't need separate code paths for different image formats, since it will take care of converting it into rgba representation for us. Since this is the case, maybe we could even consider simplifying the image representation by only distinguishing between SVG images and raster images (which are already decoded)? Not sure if there is a good reason to leave the decoding to the caller.
- The binary size does increase a bit. It is now 3.86MB with all features enabled. If I only enable the "jpg" and "gif" features (so basically only support what was supported before already), it gets reduced to 3.34MB. Originally, it was 3.29MB, so I think this is a negligible difference? We could also do it so that by default, only "jpg" and "gif" are supported and add another feature flag to enable the remaining image formats, which are not mandated by the SVG specification.
- I want to check whether it has any impact on performannce for the already existing formats by benching it. But before I spend time on that, I want to know whether you would be willing to switch to the `image` crate in the first place.

Let me know what you think!


LaurenzV:
That's weird, the tests run fine locally. 😅 I hope that doesn't mean that JPEG decoding is OS-dependent...

RazrFalcon:
Im aware of the `image` create and intentionally do not use it. It's too heavy for our use. We only need bitmap data decoders - nothing else.

>Not sure if there is a good reason to leave the decoding to the caller.

1. Some people prefer to use their own decoders. This is intentional. #521 
1. By using low-level decoders API we can avoid unnecessary transformations.
1. I don't want to decode images on the `usvg` side. This would be expensive. 
1. Some people might prefer decoding 16bit data, while resvg, currently, is 8bit only. 
1. The current code isn't that complicated to begin with.

Either way, we can try adding TIFF and BMP, if you really want to, but this is a very low priority task.
WEBP and AVIF would require pure Rust decoders.

LaurenzV:
Gotcha! Since WEBP is the only one I actually saw being requested, I might open a new PR once the decoder create is published.

PR #690, opened 2023-12-26
Fix  #685

### LINKED ISSUE #685: When `feImage` contains `x` or `y`, the rendering is incorrect.
```xml
<svg width="200" height="200" viewBox="0 0 200 200" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
  <rect width="200" height="200" fill="#ffdd88"/>

  <defs>
    <rect fill="#ff0000" opacity="1" id="rect" width="100" height="100"/>
    <filter id="colorful-image" >
      <feImage xlink:href="#rect" href="#rect" x="30" y="50" 
        result="feImageResultnode" />
      <feMerge>
        <feMergeNode in="feImageResultnode" />
      </feMerge>
    </filter>
  </defs>
  <rect x="30" y="50" width="100" height="100" fill="#1e1705" filter="url(#colorful-image)"/>
  <rect x="30" y="50" width="100" height="100" stroke="#00f" fill="none"/>
</svg>
```

i get left

LaurenzV:
I spent some time in the debugger and here is my understanding of what is going on / why the issue exists:

- First, we determine the filter bbox, since this is the area where the filter should be contained. Nothing should exceed this.
- We then create a pixmap with the size of the filter bbox, which will contain all of the graphics that are part of the filter. Since the pixmap starts at (0, 0), we need to shift everything we render to account for the x/y of the filter area.

If we take the following SVG:
```svg
<svg width="200" height="200" viewBox="0 0 200 200" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
  <rect width="200" height="200" fill="#ffdd88"/>

  <defs>
    <rect fill="#ff0000" opacity="1" id="rect" width="100" height="100"/>
    <filter id="colorful-image" >
      <feImage xlink:href="#rect" href="#rect" x="40" y="60" 
        result="feImageResultnode" />
      <feMerge>
        <feMergeNode in="feImageResultnode" />
      </feMerge>
    </filter>
  </defs>
  <rect x="30" y="50" width="100" height="100" fill="#1e1705" filter="url(#colorful-image)"/>
  <rect x="30" y="50" width="100" height="100" stroke="#00f" fill="none"/>
</svg>
```

The filter bbox would be `x=20, y=40, width=120, height=120`, so we translate everything by (-20, -40). As a result, the actual x and y of the `feImage` in the filter pixmap _should be_ 20, 20. However, when rendering the filter onto the pixmap, we aren't accounting for this shift anywhere, meaning that the image will start at (0, 0) instead of (20, 20). As a result of this, the rect won't be positioned accordingly, and as another side effect the clipping algorithm will clip parts of the rectangle that shouldn't be clipped.

All I did was add a transformation that accounts for the x/y position of the subregion. I also added some test cases to reflect that it works as intended now.

However, I have to admit that even after spending hours in the debugger, I haven't completely understood the whole logic of the filter rendering part. 😅 In particular, I'm unsure whether the translation should be preconcatenated or postconcatenated to the scaling (I haven't even figured out why that's needed), so I would appreciate your input on that.


RazrFalcon:
>I'm unsure whether the translation should be preconcatenated or postconcatenated to the scaling 

I don't blame you. Affine transforms are not very intuitive. 
And yes, you code is correct here. You can easily check it by adding `--zoom 2` to resvg.

> I haven't completely understood the whole logic of the filter rendering part

I don't blame you. x2 
SVG filters are ridiculously complicated and badly documented. That's why no one really supports them. It took me months to someone understand them.
I've tried my best to make the code as simple as possible, but it's hard. Especially if we remember that filters is a very performance sensitive part of the renderer. 

PS: please make sure all new SVG tests do not have lines longer than 100 characters. And make sure all elements have an `id` attribute.

crates/resvg/src/filter/mod.rs
```diff
@@ -972,7 +972,10 @@ fn apply_image(
         }
         usvg::filter::ImageKind::Use(ref node) => {
             let (sx, sy) = ts.get_scale();
-            let transform = tiny_skia::Transform::from_scale(sx, sy);
+
+            let transform =
```
RazrFalcon, line 976:
```rust
let transform = tiny_skia::Transform::from_row(sx, 0.0, 0.0, sy, subregion.x() as f32, subregion.y() as f32);
```

should be enough.

LaurenzV:
> I've tried my best to make the code as simple as possible, but it's hard. Especially if we remember that filters is a very performance sensitive part of the renderer.

I can imagine, no worries! I'm still in awe that you managed to write this library (including its dependencies) all by yourself.

RazrFalcon:
Thanks!

Yes, it was and still is an interesting project. It took around 3 years full time to get to this point. And back then I had a lot of free time, but not anymore.

Issue #647, opened 2023-07-18
Add optional parameters to limit the maximum dimensions ---max-width, --max-height


totoestcontent:
Some svgs have a huge viewbox and I would need to limit the size of the png produced.
The current settings (-w and -h) allow this if the image is larger than the given values.
If the image is smaller then it will be enlarged.
The idea is therefore not to exceed a given dimension (optional).

I made the changes and created a git patch.

Be indulgent, it's the first time I do Rust ;-)
So you can do much better, but if it can make your job easier, I'd be happy ;-)

[add_max_width_height.patch](https://github.com/RazrFalcon/resvg/files/12081836/add_max_width_height.patch)

totoestcontent:
I need to preview all svg images with a maximum dimension of 200x200.
If the svgs are smaller than 200x200 then the previews must be the size of the svg.
If the svgs are larger, then the preview should fit into a 200x200 rectangle.
So I don't want to reject some images because they're too big.

RazrFalcon:
Isn't this exactly what the current `--width` + `--height` attributes are doing?

totoestcontent:
Not exactly.

With image 900_600.svg if i call: `resvg --width 200 900_600.svg 900_600_width_200.png` , png has dimension: 200x134
With image 32x32.svg if i call: `resvg --width 200 32_32.svg 32_32_width_200.png`, png has dimension: 200x200,
with --max-width, i want: 200x134 and 32x32
![900_600](https://github.com/RazrFalcon/resvg/assets/55127910/8500cb85-a64e-4288-a351-4fe02453cec0)
![32_32](https://github.com/RazrFalcon/resvg/assets/55127910/213fd21d-9ba1-480e-a821-72e9e8e3e4a1)
![900_600_width_200 svg](https://github.com/RazrFalcon/resvg/assets/55127910/5a78df61-3a88-4473-b218-b2a964712442)
![32_32_width_200 svg](https://github.com/RazrFalcon/resvg/assets/55127910/969c94a4-8e21-40a9-a5d3-37734e8c1126)

RazrFalcon:
Oh, so you do not what to upscale the SVG, only downscale it. I get it now.
Yes, this is not supported. Will think how it could be done.
To me, the main issue is naming, since when I see `--max-width` I think the max limit, a cut off, which is not what you want. I do plan an option that would return an error when the limit is reached.
It's more like `--width-limit` to me. Or maybe even `--downscale-only`.

[...]

PR #693, opened 2024-01-03
issue #647: Add optional parameters to limit the maximum dimensions -…

totoestcontent:
@RazrFalcon, As we discussed in "[#647]", here are the modifications to add the two optional parameters [--max-width, --max-height].

If there is the slightest problem with what I deliver, you should tell me :-).

Thanks you

### DIFF
--- a/crates/resvg/src/main.rs
+++ b/crates/resvg/src/main.rs
@@ -160,6 +160,13 @@ OPTIONS:
   -z, --zoom FACTOR             Zooms the image by a factor
       --dpi DPI                 Sets the resolution
                                 [default: 96] [possible values: 10..4000 (inclusive)]
+                                
+  --max-width LENGTH            Sets the maximum width in pixels (default: no limit)
+                                Used during normal rendering and not during --export-id
+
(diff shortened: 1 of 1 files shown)


RazrFalcon:
Thanks, but that's definitely not the way I would write it.

For one, couldn't we just overwrite `fit_to` in `parse_args`? No need for a separate property. This would simply the code quite a lot.

And please run `cargo fmt` before committing.

crates/resvg/src/main.rs
```diff
@@ -160,6 +160,13 @@ OPTIONS:
   -z, --zoom FACTOR             Zooms the image by a factor
       --dpi DPI                 Sets the resolution
                                 [default: 96] [possible values: 10..4000 (inclusive)]
+                                
+  --max-width LENGTH            Sets the maximum width in pixels (default: no limit)
```
RazrFalcon, line 164: Put new options right after `--height`. No need for empty lines either.

RazrFalcon:
Closing after no response.

PR #698, opened 2024-01-17
Add copyright notice to LICENSE.txt

### LINKED ISSUE #697: Add Copyright Notice
The Volkswagen Group is using resvg as a component of Rhino (https://www.rhino3d.com). As part of their open source validation process, Volkswagen Group requests that a copyright notice be added to the license file. For example:

"Copyright 2017-2024 Yevhenii Reizner. All Rights Reserved."

An acceptable example of copyright + license is here:
https://github.com/Stykka/glTF-Bin/blob/master/LICENSE.txt

Thank you.

golsby:
Fixes #697

A valid copyright notice is required to pass the Volkswagen Group open source validation so that they can use this project as part of Rhino software (www.rhino3d.com).

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,3 +1,5 @@
+Copyright (C) 2017-2024 Yevhenii Reizner. All Rights Reserved.
+
 Mozilla Public License Version 2.0
 ==================================
 


golsby:
Thank you for considering this. I also don't know the officially correct way to do this; it seems to vary a lot. I found this clause in the MPF FAQ: https://www.mozilla.org/en-US/MPL/2.0/FAQ/

"To apply the Mozilla Public License to software that you have written, add the header from Exhibit A of the license to each source code file in your project. Sample headers for various commenting styles are available [here.](https://www.mozilla.org/en-US/MPL/headers/) You may also add additional accurate notices of copyright ownership, such as the name of the copyright holder, but this is not necessary."

This seems to apply to source files themselves and not the license file. 

It seems like the NOTICE.txt would be adequate. Thanks!

RazrFalcon:
Or we could use the whole text from: https://www.apache.org/licenses/LICENSE-2.0#apply It's seems generic enough.

golsby:
Now that I look deeper, this format seems very common (at least for MIT). Having the copyright with the license seems easier than the noticefile.txt; but I'll let you decide.

https://github.com/psf/black/blob/main/LICENSE

RazrFalcon:
The MIT license does include the copyright line, yes, but not all licenses do. We cannot randomly modify the license file template. Therefore something like NOTICE.txt is the only other choice I know of. 

See https://choosealicense.com/licenses/mpl-2.0/ vs https://choosealicense.com/licenses/mit/

Many open source licenses do not include the `Copyright` line, therefore you shouldn't be expecting in the first place.

golsby:
Thanks for continuing to do this. I'm happy with the NOTICEFILE.txt solution; Honestly, I don't need this - it's the massive legal machinery at Volkwagen that seems to want it - I'm just the messenger.

RazrFalcon:
No problem. I understand.

Issue #714, opened 2024-02-11
Only width/height on image


LaurenzV:
Hello,

I was investigating https://github.com/typst/svg2pdf/issues/58 and found out that the issue is that we don't consider the case where only one of width/height is set. After fixing it, I realized that this does seem to be intentional though? Because two test cases fail now, one of which being:

```svg
<svg id="svg1" viewBox="0 0 80 80" xmlns="http://www.w3.org/2000/svg"
     xmlns:xlink="http://www.w3.org/1999/xlink">
    <title>No `width` (SVG 2)</title>
    <desc>Nothing should be rendered in a SVG 1 conforming renderer.</desc>

    <image id="image1" x="8" y="8" height="64" xlink:href="../../../resources/image.jpg"/>

    <rect id="frame" x="1" y="1" width="78" height="78" fill="none" stroke="black"/>
</svg>
```

Does this mean we don't plan to support this at all? Or is this just a remnant from when we didn't support SVG2 yet? If so, I can open a PR that fixes this.

RazrFalcon:
I'm not sure I understand the issue. The current implementation works as expected.

LaurenzV:
Consider the following SVG (the image is a 100x100 pixel blue image):

```svg
<svg id="svg1" viewBox="0 0 200 200" xmlns="http://www.w3.org/2000/svg"
     xmlns:xlink="http://www.w3.org/1999/xlink">

    <rect id="frame" x="0" y="0" width="200" height="500" fill="red"/>

[...]

PR #715, opened 2024-02-11
Add support for no `width` and no `height` on images

LaurenzV:


LaurenzV:
Closes #714

LaurenzV:
The images are linked by relative paths because that's how it was before and I didn't know if I should change it. Should I embed them as base64 strings instead?

RazrFalcon:
References are fine.

As for the code, I would suggest using `tiny_skia_path::Size::scale_to_width` and `scale_to_height`. They are basically the same, but might make the code nicer. The current implementation is too verbose. 
I would also suggest calling `convert_user_length` first and then matching `Option<T>` pair.

LaurenzV:
> As for the code, I would suggest using tiny_skia_path::Size::scale_to_width and scale_to_height.

Seems like this API is only available for `IntSize` and not `Size`?

RazrFalcon:
Much better now.

>Seems like this API is only available for IntSize and not Size?

Weird, ok.

Also about the PR name. We do support optional width/height. We do not support aspect preserving. 
So I guess it should be called: "Preserve image aspect ratio when `width` or `height` are missing.". Does this sound ok?

RazrFalcon:
Also, I think we need tests with non-rectangular images. Otherwise the aspect is always 1:1 and we're not testing anything. No?

RazrFalcon:
I guess current tests are fine, because you do scale the image, but the aspect is always 1:1 right now. So I guess a single test with a non-rectangular image would be nice.

Let's use this image:

image32x16

LaurenzV:
> Also, I think we need tests with non-rectangular images. Otherwise the aspect is always 1:1 and we're not testing anything. No?

It's not necessarily about the aspect.

Before: If we have an image with size 100x100, but set only the height to 200, then the image would be assumed to be 100x200 -> It will get shifted instead.

Now:  If we have an image with size 100x100, but set only the height to 200, then the width will be increased by the same ratio that the height would increase/decrease, so in this case 200 -> the result is 200x200.

LaurenzV:
> I guess current tests are fine, because you do scale the image, but the aspect is always 1:1 right now. So I guess a single test with a non-rectangular image would be nice.

Okay, will add it.

RazrFalcon:
Thanks!

Issue #381, opened 2021-05-03
Transforms ignored after filters


afdw:
Consider the following file:
```svg
<?xml version="1.0" encoding="UTF-8"?>
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="512" height="512" viewBox="0 0 512 512">
    <defs>
        <filter id="filter-remove-color-and-invert-alpha" x="0%" y="0%" width="100%" height="100%">
            <feColorMatrix values="0 0 0 0 1 0 0 0 0 1 0 0 0 0 1 0 0 0 -1 1"/>
        </filter>
        <filter id="filter-0" x="0%" y="0%" width="100%" height="100%">
            <feImage xlink:href="#compositing-group-1" result="source" x="0" y="0" width="737.28" height="737.28"/>
            <feImage xlink:href="#compositing-group-2" result="destination" x="0" y="0" width="737.28" height="737.28"/>
            <feComposite in="source" in2="destination" operator="xor" color-interpolation-filters="sRGB"/>
        </filter>
        <filter id="filter-1" x="0%" y="0%" width="100%" height="100%">
            <feImage xlink:href="#compositing-group-3" result="source" x="0" y="0" width="737.28" height="737.28"/>
            <feImage xlink:href="#compositing-group-4" result="destination" x="0" y="0" width="737.28" height="737.28"/>
            <feComposite in="source" in2="destination" operator="arithmetic" k1="0" k2="1" k3="1" k4="0" color-interpolation-filters="sRGB"/>
        </filter>
        <g id="compositing-group-0" transform="translate(142.64, 142.64)">
            <rect x="-142.64" y="-142.64" width="737.28" height="737.28" fill-opacity="0"/>
            <rect x="-142.64" y="-142.64" width="737.28" height="737.28" fill="rgb(100%, 100%, 100%)"/>
        </g>
        <mask id="mask-0">
            <use xlink:href="#compositing-group-0"/>
        </mask>
        <mask id="mask-1">
            <use xlink:href="#compositing-group-0" filter="url(#filter-remove-color-and-invert-alpha)"/>
        </mask>
        <mask id="mask-2">
            <g transform="translate(142.64, 142.64)">
                <path fill="rgb(100%, 100%, 100%)" d="M 10 0 L 130 0 L 130 90 L 10 90 Z M 10 0 "/>
            </g>
        </mask>
        <g id="compositing-group-1" mask="url(#mask-2)">
[...]

PR #720, opened 2024-02-20
Add filter test cases

### LINKED ISSUE #380: Different results with feImage
Consider the following file:
```svg
<?xml version="1.0" encoding="UTF-8"?>
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="100" height="100" viewBox="0 0 100 100">
    <defs>
        <filter id="filter-0" x="0%" y="0%" width="100%" height="100%">
            <feImage xlink:href="#compositing-group-0"/>
        </filter>
        <g id="compositing-group-0">
            <circle cx="10" cy="10" r="30"/>
        </g>
    </defs>
    <g filter="url(#filter-0)">
        <rect x="-25" y="-10" width="100" height="30"/>
    </g>
</svg>
```

The results are:
| resvg | rsvg | Chrome, Inkscape |
| - | - | - |
| re | r | ink |

Who is right here? I do not seem to understand the SVG specification on how all these coordinates work.

Also, removing the `x="0%" y="0%" width="100%" height="100%"` part changes the result in Chrome and Inkscape, but not in resvg and rsvg.

LaurenzV:
Closes #380 and #381.


RazrFalcon:
Thanks!

PR #725, opened 2024-03-18
Fix context paint on shape with zero-size bbox

LaurenzV:
Consider the following SVG:

```svg
    <title>`context-stroke` on shape with zero-size bbox</title>

    <marker id="marker1" refX="10" refY="10" markerWidth="20" markerHeight="20">
        <path id="path-marker" d="M 10 0 16 20 H 4 Z" fill="context-stroke"/>
    </marker>
    <path id="path1" stroke="green" d="M 100 25 L 100 175"
          marker-end="url(#marker1)"/>
    <!-- image frame -->
</svg>
```

Currently, it renders like this:
test

The reason being that in my previous implementation, I wrongly assumed that we always need to have a bbox for `context-fill`/`context-stroke`, which is not true for solid paints. Since a line doesn't have a bbox, this example would fail to render correctly previously. This PR fixes this:
test


RazrFalcon:
Thanks. Zero width/height bbox are tricky to handle in SVG.

PR #728, opened 2024-03-27
Split text layouting and text outlining into two separate steps

LaurenzV:
I had to update one reference image, but it was only a one pixel difference. Probably due to some float rounding issue.


LaurenzV:
The feature thing is probably not solved very nicely... I'm open for suggestions on how else to do it.

crates/usvg/src/tree/text.rs
```diff
@@ -562,12 +567,20 @@ impl Text {
...
     pub fn flattened(&self) -> &Group {
         &self.flattened
     }
 
+    /// The positioned glyphs and decoration spans of the text.
+    ///
+    /// This should only be used if you need more low-level access
+    /// to the glyphs that make up the text. If you just need the
+    /// outlines of the text, you should use `flattened` instead.
+    #[cfg(feature = "text")]
```
RazrFalcon, line 579: Aren't text nodes are disabled completely when there are no `text` feature? Or have you changed that?

crates/usvg/src/tree/text.rs
```diff
@@ -562,12 +567,20 @@ impl Text {
...
+    pub fn layouted(&self) -> &Vec<TextFragment> {
```
RazrFalcon, line 580: Shouldn't we return just `[TextFragment]`?

crates/usvg/src/tree/text.rs
LaurenzV: I don't think they are. Otherwise this wouldn't even compile without a cfg attribute, no? (The same for all other appearences of `Node::Text`)

https://github.com/RazrFalcon/resvg/blob/5b97397af1bc995f849b274118039afa6f4d1ee3/crates/usvg/src/tree/mod.rs#L882-L889

From what I gathered only text to path conversion is locked behind the feature gate.

crates/usvg/src/tree/text.rs
LaurenzV: Makes sense. 😄

crates/usvg/src/tree/text.rs
LaurenzV: Should I change this behaviour maybe?

RazrFalcon: Yes, Text nodes themselves are enabled, but never parsed without the `text` feature. 
I changes this behaviour multiple times and already forgot which one is the current.

I think we should change it so calling `layouted()` without the `text` feature would simply print a warning. Meaning `TextFragment` should be public as well.

The `flattened()` comment is also outdated...

```rust
    /// Text converted into paths, ready to render.
    ///
    /// Returns `None` when the `text` build feature was disabled.
    pub fn flattened(&self) -> &Group {
        &self.flattened
    }
``` 

It will never return `None` anymore. The text node would not be parsed.

In a nutshell, no `text` feature - no `text` elements parsing. This is because we _must_ know bboxes of all nodes, which is impossible for text without layout/flattening.

crates/usvg/src/text/layout.rs
```diff
@@ -2,364 +2,465 @@
...
+    AlignmentBaseline, ApproxZeroUlps, BaselineShift, DominantBaseline, Fill, FillRule, Font,
+    FontStretch, FontStyle, LengthAdjust, PaintOrder, Path, ShapeRendering, Stroke, Text,
+    TextAnchor, TextChunk, TextDecorationStyle, TextFlow, TextPath, TextSpan, Visibility,
+    WritingMode,
+};
 use fontdb::{Database, ID};
 use kurbo::{ParamCurve, ParamCurveArclen, ParamCurveDeriv};
 use rustybuzz::ttf_parser;
+use rustybuzz::ttf_parser::GlyphId;
+use std::collections::HashMap;
```
RazrFalcon, line 16: I would prefer the old style: std imports, empty line, external imports, empty line, crate imports.

crates/usvg/src/text/mod.rs
```diff
@@ -0,0 +1,29 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use crate::Text;
+
+mod layout;
+mod outline;
+
+pub use layout::{PositionedGlyph, Span, TextFragment};
```
RazrFalcon, line 10: I would suggest exporting just the `layout` module. Just like we do with filters. To have a cleaner global namespace.

So the caller would use `layout::Span` instead of `Span`, which we already have in `Text`.

crates/usvg/src/text/layout.rs
```diff
@@ -2,364 +2,465 @@
...
+/// A text fragment is a component of a text element.
+///
+/// It can either be a span (which contains a number of layouted glyphs that share
+/// some properties) or a path (which is a layouted decoration span).
+#[derive(Clone, Debug)]
+pub enum TextFragment {
+    /// A text span.
+    Span(Span),
+    /// A path.
+    Path(Path),
```
RazrFalcon, line 66: What is `Path` here? The `Path` node? Why?

crates/usvg/src/text/outline.rs
```diff
@@ -0,0 +1,125 @@
...
+
+fn resolve_rendering_mode(text: &Text) -> ShapeRendering {
+    match text.rendering_mode {
+        TextRendering::OptimizeSpeed => ShapeRendering::CrispEdges,
+        TextRendering::OptimizeLegibility => ShapeRendering::GeometricPrecision,
+        TextRendering::GeometricPrecision => ShapeRendering::GeometricPrecision,
+    }
+}
+
+pub(crate) fn outline_text(
```
RazrFalcon, line 22: I think this one should be `flatten` and not outline. Same with the file name.

crates/usvg/src/tree/text.rs
LaurenzV: > Meaning TextFragment should be public as well.

I actually did that initially. The problem is that a positioned glyph contains a glyph ID (which is from `ttf-parser`, which is only imported when the text feature is enabled). I could probably replace this by u16, but the other problem is that each glyph also holds the ID of the font, which is part of the `fontdb` crate. Not sure whether this can be replaced easily?

crates/usvg/src/text/layout.rs
LaurenzV: Yes, it's a usvg path. Right now it's basically just a decoration span. I guess I should rename the enum variant.

RazrFalcon:
Looks good to me. I'm still confused why we need `Path` in `TextFragment`, but other than that it's fine.

The current limitation is that there is no way to get the original text/character position from the layout, which will be useful when generation a PDF with `ToUnicode` object.

crates/usvg/src/tree/text.rs
RazrFalcon: Ugh... Then let's keep it as is.

crates/usvg/src/text/layout.rs
RazrFalcon: Oh, so those are for underlines and stuff? Hm... good question about we suppose to store them.
I guess we need just a list of rectangles. Maybe something like:

```rust
struct Layout {
    fragments: Vec<TextFragment>,
    underlines: Vec<NonZeroRect>,
    overlines: Vec<TextFragment>,
    strikethrougs: Vec<TextFragment>,
}
```

LaurenzV: I'm not sure I understand, are you suggesting to store text fragments that have different decoration spans in different vectors? But that would mess up the order, no? What problem do you see with the current approach? By storing decorations as paths and text spans in the same vector, we can simply iterate over them in the order they are defined in the vector and render them separately so we don't need to worry about that.

LaurenzV: And it's also nice if I as a consumer of the API don't need to worry about resolving the text decorations myself but can simply render the resulting paths.

crates/usvg/src/tree/text.rs
LaurenzV: Okay! We can always come up with a better solution later on.

> The flattened() comment is also outdated...

Yes, I noticed this as well and already updated the comment.

crates/usvg/src/text/layout.rs
LaurenzV: Actually... What I could do is that we store the paths of underlines, line-throughs and overlines in the span itself. And then we basically only have a list of spans. This might be a bit cleaner. Do you want me to change this?

LaurenzV:
> Looks good to me. I'm still confused why we need Path in TextFragment, but other than that it's fine.

I removed text fragment now. What do you think?

> The current limitation is that there is no way to get the original text/character position from the layout, which will be useful when generation a PDF with ToUnicode object.

I think it should be okay because I will still load the requested font and then save the cmap into the PDF, which should be enough (at least that's what Typst does and it seems to work fine). But I guess I will find out soon.

crates/usvg/src/text/layout.rs
RazrFalcon: My problem is that text decoration should not be part of `TextFragment`. Because it's kinda isn't.
If we have a underline over the whole text - then we have a single line/path. Why should it be part of `TextFragment `? It doesn't make any sense to me.

In case of text-on-path decorations _can_ be rotated, meaning just a Rect wouldn't be enough. So Path is fine for now.

And no, no one would be resolving text decoration manually, mainly because it would be impossible.

RazrFalcon:
You cannot release get a Unicode character from TrueType `cmap`. You're suppose to use `ToUnicode`. 
PS: I had a "pleasure" of working with PDF as well...

crates/usvg/src/text/layout.rs
LaurenzV: As mentioned below, I've removed `TextFragment` and store text decorations with the span now, I hope that's better!

LaurenzV:
Oh damn, yeah, you are right, I need to somehow store the corresponding string for the glyphs... Will see if I can figure out how.

LaurenzV:
Actually, sorry, there is one more small thing I would like to implement, so it would be good if we could hold off merging for now. I'll write here once I think it's done.

LaurenzV:
Okay well, I've managed to produce the same visual output for all test cases but embedded text instead, so from my side I have everything I need in that regard. I still want to get copy-pasting to work, so still want to try to figure out a way to map the glyphs to their respective character, but I guess this can happen in a separate PR in the next few days, if I find the time. I'm already pretty happy with what I have for now. :D

RazrFalcon:
Sure, sounds good.

I'm still trying to understand how do you handle decorations. I'm not sure they should be part of `layout::Span`.
Let's say we have "Hello world" where the second word is green, but the whole text has the same style otherwise and we have underline for the whole text. In this case we would have two spans but one underline. How do you handle it right now?
Because the old implementation would join continuous decorations.

LaurenzV:
So, I just checked with the following SVG:

```svg
     font-family="Noto Sans" font-size="64">
    <title>Simple case</title>

    <text id="text1" x="32" y="100" text-decoration="underline">Te<tspan fill="green">x</tspan>t</text>

    <!-- image frame -->
</svg>
```

In this case, there will be 3 spans and each of them has a separate underline (but each underline only covers the range of the glyphs they contain, so they are not overlapping)... But I think it should be fine like that, right? I mean it can't be noticed that they are three different underlines, anyway. Or is that a problem?

RazrFalcon:
Great, the current implementation also breaks underlines. I don't know how my own code works anymore... I wrote it like 4 years ago.

Either way, let's leave decoration as a part of a span. Not ideal, but not breaking either.

LaurenzV:
Thanks!

PR #729, opened 2024-03-29
Add text corresponding to layouted glyph

LaurenzV:
What do you think of this? I basically adapted it from https://github.com/typst/typst/blob/105d7156f8f9d95e16b3eefdf0fa97e5be7fbe5b/crates/typst/src/layout/inline/shaping.rs#L729-L733

I also checked and I've managed to get copyable text to work (in `svg2pdf`), I also tried it with other scripts (like Arabic) and it seems to work!

### DIFF
--- a/crates/usvg/src/text/layout.rs
+++ b/crates/usvg/src/text/layout.rs
@@ -27,13 +27,15 @@ use crate::{
 ///
 /// Note that the transform already takes the font size into consideration, so applying the
 /// transform to the outline of the glyphs is all that is necessary to display it correctly.
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Debug)]
 pub struct PositionedGlyph {
     /// The transform of the glyph. This transform should be applied to the _glyph outlines_, meaning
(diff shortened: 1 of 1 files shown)


LaurenzV:
Umm... I have no idea why Qt is failing?

RazrFalcon:
Yep, it's unrelated. I guess it fails to download the Qt version I need. Ignore it.

As for the patch, yes, Arabic/Hebrew is the first thing you should test. [Zalgo](https://zalgo.org/) as well.

I think the old code was able to get `char` by `byte_idx`, but not the whole string/cluster. Allocating the whole string for each span might be a bit expensive, but I guess it's fine for now. We could use `Box<str>` as an optimization here.

LaurenzV:
Yup, Hebrew seems to work as well. Not so much for zalgo (as in, it renders correctly, but copy-pasting results in more or less garbage), but it also doesn't work when exporting the SVG to PDF via Chrome, so I don't think this is worth investigaging...

RazrFalcon:
Thanks!

PR #735, opened 2024-04-10
Add support for color fonts

LaurenzV:
Not polished code yet, but it should be enough to test it.

Oh, and top-to-bottom text also doesn't quite work yet. 😅


LaurenzV:
So I just tried adding support for SVG and the positioning seems to work out-of-the-box (for horizontal writing), so the issue definitely seems limited to bitmap emojis...

RazrFalcon:
Thanks, will take a look.

LaurenzV:
COLRv0 works now, too!

```svg
      font-size="25">
    <title>Emojis</title>

    <text id="text1" x="0" y="50" font-family="Segoe UI Emoji">Hii😀 😁how </text>
    <text id="text1" x="0" y="90" font-family="Twitter Color Emoji">Hii😀😁how </text>
    <text id="text1" x="0" y="130" font-family="Apple Color Emoji">Hii😀😁how </text>
    <text id="text1" x="0" y="160" font-family="Noto Color Emoji">Hii😀😁how </text>

    <!-- image frame -->
</svg>
```

Our output:
test

Chrome:
image

LaurenzV:
Here is a toy sbix font we can use to debug: https://github.com/simoncozens/test-fonts?tab=readme-ov-file#cff-and-sbixotf

If I install it locally and open the following SVG:

```svg
      font-size="25">
    <title>Emojis</title>

    <text id="text1" x="0" y="100" font-family="CFF Outlines and SBIX">AAA</text>

    <!-- image frame -->
</svg>
```

I get the same behaviour on Firefox and Chrome. I'm trying to change different glyph metrics and recompile the font with `fonttools`, but none of them have an impact on the y offset. So if you feel like looking into it, this might help.

RazrFalcon:
Thanks for looking into it.

LaurenzV:
Looks like we are not the only ones troubled by this: https://github.com/harfbuzz/harfbuzz/issues/2679#issuecomment-1345595425

So can we ignore it for now?

RazrFalcon:
We should at least try to use the same hack everyone else is using.

yisibl:
> COLRv0 works now, too!

So Cool! Thank you for all you have done for this.

LaurenzV:
> We should at least try to use the same hack everyone else is using.

Well, guess what:

Consider the following SVG with the font I mentioned above:
```svg
      font-size="64">
    <title>Emojis</title>

    <text id="text1" x="0" y="100" font-family="CFF Outlines and SBIX">AAA</text>

    <!-- image frame -->
</svg>
```

Here is how it renders on MacOS in Chrome:
image

Here is how it renders on Windows 11 in Chrome:
image

So what I think is happening here is that Chrome, Firefox etc, use CoreText for rendering text in SVG, which is why on MacOS, they have the same behavior as all other Apple programs (like TextEdit, as mentioned in the issue linked above). However, on Windows they use something else, meaning that we don't have this same behaviour there... So I still think it's safe if we just ignore this for now, right?

RazrFalcon:
>So I still think it's safe if we just ignore this for now, right?

Sounds good to me.

LaurenzV:
I think it mostly works now, but there is still something wrong with ttb text that I need to investigate once I find the time to do so...

LaurenzV:
Also I'm not sure if 14MB is too big for a font (Twitter Color Emoji), should I use the SVG font from Simon Cozen's `test-font` repository instead?

RazrFalcon:
A smaller font is preferable. You could try subsetting the Twitter font with `pyftsubset`. Just not forget the license file and some readme explaining how you got this file.

LaurenzV:
Regarding subsetting, I think I tried that but iirc fonttools didn't support subsetting SVG tables. But I will try again.

So should I disregard TTB writing for now?

yisibl:
> Regarding subsetting, I think I tried that but iirc fonttools didn't support subsetting SVG tables. But I will try again.

I remember that `fonttools` already support subsetting OT-SVG fonts, see https://github.com/fonttools/fonttools/issues/534#issuecomment-970496354

Alternatively, you can use [nanoemoji](https://github.com/googlefonts/nanoemoji) to convert SVG files into color fonts in a variety of formats, which will produce test fonts that are small enough to use.

LaurenzV:
> I remember that fonttools already support subsetting OT-SVG fonts, see https://github.com/fonttools/fonttools/issues/534#issuecomment-970496354

Huh okay, I'll try it again then, thanks!

LaurenzV:
I think I'm getting close now. I'll try to investigate the tb writing mode issue again, then clean up and then it should be ready. Hopefully will be done within the next few days.

LaurenzV:
There's still something off, but not sure I will be able to fix it for now, this would require me to make a deep dive into how `writing-mode` works exactly, which I might do at some point, but I don't think it's worth it to hold up emoji support for this... This is how it looks currently:

test

versus in other browsers:

image

For some reason there is more space in the vertical writing mode... But no idea why. I will add a test case regardless so we don't forget about it.

LaurenzV:
I've added a few small TODOs where I wasn't sure what I should change it to.

crates/usvg/src/text/flatten.rs
```diff
@@ -19,58 +24,134 @@ fn resolve_rendering_mode(text: &Text) -> ShapeRendering {
...
+        // to create a new path for every glyph if we have many consecutive glyphs
+        // with just outlines (which is the most common case).
         let mut span_builder = tiny_skia_path::PathBuilder::new();
 
         for glyph in &span.positioned_glyphs {
-            if let Some(outline) = fontdb.outline(glyph.font, glyph.glyph_id) {
-                if let Some(outline) = outline.transform(glyph.transform) {
-                    span_builder.push_path(&outline);
+            // A COLRv0 glyph. Will return a vector of paths that make up the glyph description.
+            // TODO: Don't use black for foreground color? But not sure whether to use fill or stroke
```
RazrFalcon, line 79: I don't think this affects SVG much, but I do think we should prefer a transparent background here.

crates/usvg/src/text/flatten.rs
```diff
@@ -4,12 +4,17 @@
...
+use tiny_skia_path::{NonZeroRect, Size, Transform};
+
+use std::mem;
 use std::sync::Arc;
-use tiny_skia_path::{NonZeroRect, Transform};
 
-use crate::tree::BBox;
-use crate::{Group, Node, Path, ShapeRendering, Text, TextRendering};
+use crate::layout::Span;
+use crate::{
```
RazrFalcon, line 14: Let's just import all. Aka `use crate::*;`

crates/usvg/src/text/flatten.rs
```diff
@@ -4,12 +4,17 @@
 
 use fontdb::{Database, ID};
 use rustybuzz::ttf_parser;
-use rustybuzz::ttf_parser::GlyphId;
+use rustybuzz::ttf_parser::{GlyphId, RasterImageFormat};
+use tiny_skia_path::{NonZeroRect, Size, Transform};
+
+use std::mem;
```
RazrFalcon, line 10: `std` imports should be first. Then external crates. Then internal stuff.

crates/usvg/src/text/flatten.rs
```diff
@@ -19,58 +24,134 @@ fn resolve_rendering_mode(text: &Text) -> ShapeRendering {
...
-                Transform::default(),
-            )
-        }) {
-            stroke_bbox = stroke_bbox.expand(path.stroke_bounding_box());
-            new_paths.push(path);
+                let mut group = Group {
+                    transform: glyph.svg_transform(),
+                    ..Group::empty()
+                };
+                // TODO: Probably need to update abs_transform of children?
```
RazrFalcon, line 105: Yeah, I'm still thinking about a better `abs_transform` implementation. The current one is pretty meh.

crates/usvg/src/text/flatten.rs
```diff
@@ -126,8 +220,129 @@ impl DatabaseExt for Database {
...
+            if image.format == RasterImageFormat::PNG {
+                let bitmap_image = BitmapImage {
+                    image: Image {
+                        id: String::new(),
+                        visibility: Visibility::Visible,
+                        size: Size::from_wh(image.width as f32, image.height as f32)?,
+                        rendering_mode: ImageRendering::OptimizeQuality,
+                        kind: ImageKind::PNG(Arc::new(image.data.into())),
+                        abs_transform: Transform::default(),
+                        // TODO: What to change to?
```
RazrFalcon, line 243: Just an `image` size at 0x0 should be fine, afaik. After all, the flattened text has the identity transform now.

crates/usvg/src/text/flatten.rs
```diff
@@ -19,58 +24,134 @@ fn resolve_rendering_mode(text: &Text) -> ShapeRendering {
...
-            )
-        }) {
-            stroke_bbox = stroke_bbox.expand(path.stroke_bounding_box());
-            new_paths.push(path);
+                let mut group = Group {
+                    transform: glyph.svg_transform(),
+                    ..Group::empty()
+                };
+                // TODO: Probably need to update abs_transform of children?
+                group.children.push(Node::Group(Box::new(tree.root)));
```
RazrFalcon, line 106: I'm not sure if we should keep it as `Image::SVG`. But I guess it can be changed later.

crates/usvg/src/text/flatten.rs
LaurenzV: IIRC I can only choose the color but not the transparency... But I will check again when I'm on my computer.

RazrFalcon:
Thanks! Looks good to me. 
I will think about how we can simplify `abs_transform` handling.

crates/usvg/src/text/flatten.rs
RazrFalcon: You're passing RGBA color here, so 0,0,0,255 should work just fine.

LaurenzV: Sorry, but I'm still not following... I'm passing `Color::black()`, which is defined as:

```rs
    /// Constructs a new `Color` set to black.
    #[inline]
    pub fn black() -> Color {
        Color::new_rgb(0, 0, 0)
    }
```

And then I'm doing:

```rs
fn paint_foreground(&mut self) {
        self.paint_color(ttf_parser::RgbaColor::new(
            self.foreground.red,
            self.foreground.green,
            self.foreground.blue,
            255,
        ));
    }
```

So isn't that already 0,0,0,255?

crates/usvg/src/text/flatten.rs
LaurenzV: Ah true... Yeah I guess both would work.

RazrFalcon:
Ready to merge?

LaurenzV:
I think so!

PR #736, opened 2024-04-12
Use ascent and descent insead of x-height for shift in tb writing mode

LaurenzV:
This seems to be more in line with that Chrome and Firefox do.

Before:
test

After:
test

Chrome:
image


RazrFalcon:
Thanks!

PR #738, opened 2024-04-14
Allow enabling GIF, JPEG, and PNG independently

IvanUkhov:
It is good to keep the dependencies to a minimum, and in this PR, `raster-images` is split into three: `gif`, `jpeg`, and `png`.

### DIFF
--- a/crates/resvg/Cargo.toml
+++ b/crates/resvg/Cargo.toml
@@ -22,7 +22,7 @@ pico-args = { version = "0.5", features = ["eq-separator"] }
 png = { version = "0.17", optional = true }
 rgb = "0.8"
 svgtypes = "0.15.0"
-tiny-skia = "0.11.4"
+tiny-skia = { version = "0.11.4", default-features = false, features = ["std", "simd"] }
 usvg = { path = "../usvg", version = "0.41.0", default-features = false }
 
 [dev-dependencies]
@@ -40,4 +40,13 @@ memmap-fonts = ["usvg/memmap-fonts"]
 # Enables decoding and rendering of raster images.
 # When disabled, `image` elements with SVG data will still be rendered.
 # Adds around 200KiB to your binary.
-raster-images = ["gif", "jpeg-decoder", "png"]
+raster-images = ["gif", "jpeg", "png"]
+
+# Enables rasterization.
+raster = []
+# Enables rasterization in GIF.
+gif = ["raster", "dep:gif"]
+# Enables rasterization in JPEG.
+jpeg = ["raster", "jpeg-decoder"]
+# Enables rasterization in PNG.
+png = ["raster", "dep:png", "tiny-skia/png-format"]
--- a/crates/resvg/src/image.rs
+++ b/crates/resvg/src/image.rs
@@ -24,11 +24,11 @@ pub fn render_inner(
         usvg::ImageKind::SVG(ref tree) => {
             render_vector(tree, transform, pixmap);
         }
-        #[cfg(feature = "raster-images")]
+        #[cfg(feature = "raster")]
         _ => {
-            raster_images::render_raster(image_kind, transform, rendering_mode, pixmap);
+            raster::render(image_kind, transform, rendering_mode, pixmap);
         }
-        #[cfg(not(feature = "raster-images"))]
+        #[cfg(not(feature = "raster"))]
         _ => {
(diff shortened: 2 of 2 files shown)


crates/resvg/Cargo.toml
```diff
@@ -22,7 +22,7 @@ pico-args = { version = "0.5", features = ["eq-separator"] }
 png = { version = "0.17", optional = true }
 rgb = "0.8"
 svgtypes = "0.15.0"
-tiny-skia = "0.11.4"
+tiny-skia = { version = "0.11.4", default-features = false, features = ["std", "simd"] }
```
IvanUkhov, line 25: Those two are enabled by default. So I am not touching them. Just removing `png-format`.

crates/resvg/src/image.rs
```diff
@@ -54,29 +54,65 @@ fn render_vector(
...
 }
 
-#[cfg(feature = "raster-images")]
-mod raster_images {
+#[cfg(feature = "raster")]
+mod raster {
     use crate::OptionLog;
 
-    fn decode_raster(image: &usvg::ImageKind) -> Option<tiny_skia::Pixmap> {
+    pub fn render(
```
IvanUkhov, line 61: Just moving it up, as it is the only public in the module.

RazrFalcon:
I think this is too extreme. More over, PNG support cannot be optional. This is kinda a bug to begin with.

I can accept GIF being optional. Maybe JPEG. But that's it.

IvanUkhov:
Could you please explain why PNG cannot be optional? `raster-images` implies that rasterization is not always necessary and can be disabled. But then it does not fully disable it due to the default features of `tiny-skia`. Moreover, I am actually coming with a use case. In my case, I do not need rasterization but only the RGBA buffer of the produced PixelMap. So it is unfortunate to see PNG being pulled in. Perhaps there are more like me.

RazrFalcon:
The `png` crate is like 3% of the final binary. It doesn't matter.

Also, text flattening requires PNG support for emojis.
So no, PNG is here to stay. 

>raster-images implies that encoding is not always necessary 

`raster-images` affects only decoding of images inside the `image` element.

Here is a `cargo bloat` output per crate:

```
12.1%  21.0% 362.0KiB std
11.3%  19.5% 337.5KiB usvg
 7.4%  12.7% 220.0KiB rustybuzz
 5.4%   9.4% 161.6KiB tiny_skia
 4.9%   8.4% 145.1KiB resvg
 3.1%   5.3%  92.3KiB ttf_parser
 1.9%   3.3%  57.4KiB jpeg_decoder
 1.6%   2.9%  49.2KiB svgtypes
 1.4%   2.5%  43.2KiB roxmltree
 1.3%   2.3%  40.2KiB tiny_skia_path
 1.3%   2.2%  38.6KiB png
 1.0%   1.7%  28.7KiB simplecss
 0.9%   1.6%  27.3KiB miniz_oxide
 0.8%   1.3%  22.6KiB fontdb
 0.6%   1.0%  17.2KiB fdeflate
 0.6%   1.0%  17.0KiB [Unknown]
 0.5%   0.9%  15.7KiB imagesize
 0.4%   0.7%  12.2KiB data_url
 0.4%   0.7%  11.8KiB weezl
 0.3%   0.5%   9.0KiB pico_args
 0.3%   0.5%   8.6KiB gif
 0.3%   0.5%   8.6KiB unicode_bidi
 0.2%   0.3%   4.9KiB kurbo
 0.2%   0.3%   4.5KiB unicode_ccc
 0.1%   0.2%   4.1KiB flate2
 0.1%   0.1%   1.9KiB unicode_vo
 0.0%   0.1%   1.2KiB crc32fast
 0.0%   0.0%     440B memmap2
 0.0%   0.0%     440B smallvec
 0.0%   0.0%     380B slotmap
 0.0%   0.0%     336B simd_adler32
 0.0%   0.0%     336B log
 0.0%   0.0%     128B bytemuck
 0.0%   0.0%     128B unicode_bidi_mirroring
 0.0%   0.0%     100B unicode_script
 0.0%   0.0%      96B unicode_properties
 0.0%   0.0%      28B arrayvec
```

As you can see, raster images support is the least of your concerns.

IvanUkhov:
It is not so much about the size for me. It is more about not having to have the dependencies I do not need. This is a crate about SVG, not PNG, after all. But I understand. Thank you for explaining.

PR #758, opened 2024-05-11
Add best-effort support for COLRv1

LaurenzV:


RazrFalcon:
Ready to merge?

crates/usvg/src/tree/text.rs
```diff
@@ -565,7 +565,20 @@ impl Text {
         self.abs_stroke_bounding_box
     }
 
-    /// Text converted into paths, ready to render.
+    /// Text converted into paths, ready to render. Note that this is only a
```
RazrFalcon, line 568: Can you format it a bit better? Check out `cargo docs` output. It would be formatted as a one long line.

LaurenzV: Not sure I follow, this is how it looks for me, what exactly is wrong with that? Should I add a new line after the first sentence?

image

RazrFalcon: >Should I add a new line after the first sentence?

Yes.

LaurenzV: Done.

LaurenzV:
> Ready to merge?

Yep.

LaurenzV:
> I plan to make a resvg release soon. Do you have any more changes planned?

Well, ideally it would've been nice to figure out the font selection problem (#754) before the next release, but it it looks like it will still take some time and I think @laurmaedje is currently busy with other stuff right now, so I guess there is nothing holding us off from making a new release right now.

RazrFalcon:
Ok, I will take a look at #754 myself this week and if it would not work will release resvg as is.

PR #769, opened 2024-06-01
Support for user-provided font loading callbacks

laurmaedje:
This PR adds support for user-provided font loading callbacks. 

Here is small overview of the changes:
- The `Options` now contain an `Arc<Database>` and a new `FontResolver<'a>`. The `fontdb` arguments on other top-level functions were removed.
- The `ImageHrefResolver` is now `ImageHrefResolver<'a>` (since it might as well). With the default resolvers, you just get `Options<'static>`.
- The resolver functions take `&mut Arc<Database>`, so they can mutate the database with `Arc::make_mut` if necessary, but if they don't need to add a font, we don't force the database to be cloned.
- For each tree conversion, we have an `Arc<Database>` in the `Cache`, which is populated over the course of conversion. It is only cloned if fonts are actually added.
- The database from the `Cache` is moved into the `Tree` after conversion and users can access it with `tree.fontdb()` to make sense of the `ID`s in their `Span`s.
- Sub-SVGs start with a fresh database from the `Options`, but share the same resolver.
- The database in the `Options` is never mutated (it's `&Options` after all)

One small change in `fontdb` is necessary to make writing a custom `FontResolver` possible, or at least convenient: The best function to add a new font is `push_face_info`, but it doesn't return the new `ID` it assigned. I've opened https://github.com/RazrFalcon/fontdb/pull/66 to change that. As it is a breaking change, resvg would need to update to a new release of fontdb. (Alternatively, we could make it non-breaking with a new function.)

Supersedes https://github.com/RazrFalcon/resvg/pull/754

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -107,9 +107,9 @@ dependencies = [
 
 [[package]]
 name = "fontdb"
-version = "0.17.0"
+version = "0.18.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6d2894eb653f564384ae8da32b78d9c8db0394715525d917328e435b9babd902"
+checksum = "e32eac81c1135c1df01d4e6d4233c47ba11f6a6d07f33e0bba09d18797077770"
 dependencies = [
  "fontconfig-parser",
  "log",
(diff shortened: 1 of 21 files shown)


RazrFalcon:
Bumped `fontdb`

laurmaedje:
> Bumped fontdb

Thanks, that was quick! Since this PR doesn't directly depend on the fontdb changes (only users need it), the dependency can be updated separately from this PR, but if you want I can also update it here.

RazrFalcon:
Let update fontdb as well.

laurmaedje:
Done

RazrFalcon:
I guess we're ready for a new resvg release?

LaurenzV:
More than ready! 😄

PR #770, opened 2024-06-02
Improve GitHub releases 

ssssota:
I want to track changes from github releases.
So, I think that release notes should be include a link to the changelog.

At the same time, the create-release action is changed to one that is also maintained.

### DIFF
--- a/.github/workflows/tagged-release.yml
+++ b/.github/workflows/tagged-release.yml
@@ -15,16 +15,19 @@ jobs:
     steps:
       - name: Create Release
         id: create_release
-        uses: actions/create-release@v1
+        uses: softprops/action-gh-release@v2
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           tag_name: ${{ github.ref }}
-          release_name: ${{ github.ref }}
+          name: ${{ github.ref }}
           body: |
             - `viewsvg` is a simple application that showcases *resvg* capabilities
             - `resvg-0.*.0.tar.xz` is a sources archive with vendored Rust dependencies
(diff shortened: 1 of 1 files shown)


.github/workflows/tagged-release.yml
```diff
@@ -15,16 +15,19 @@ jobs:
     steps:
       - name: Create Release
         id: create_release
-        uses: actions/create-release@v1
```
ssssota, line 18: Archived at 3 years ago.

https://github.com/softprops/action-gh-release has almost the same interface(inputs/outputs).

.github/workflows/tagged-release.yml
```diff
@@ -15,16 +15,19 @@ jobs:
...
           tag_name: ${{ github.ref }}
-          release_name: ${{ github.ref }}
+          name: ${{ github.ref }}
           body: |
             - `viewsvg` is a simple application that showcases *resvg* capabilities
             - `resvg-0.*.0.tar.xz` is a sources archive with vendored Rust dependencies
             - `resvg-explorer-extension.exe` is an SVG thumbnailer for Windows Explorer
+            
+            Check [CHANGELOG](https://github.com/RazrFalcon/resvg/blob/${{ github.ref }}/CHANGELOG.md).
+          generate_release_notes: true
```
ssssota, line 30: append release notes automatically generated.
https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes

You can remove it if you feel it is unnecessary.

```suggestion
```

RazrFalcon:
Sure, makes sense.

As for autogenerated changelog - I'm strongly against it.

RazrFalcon:
And it got commited anyway...

Issue #777, opened 2024-06-10
Some Microsoft Segoe UI Emoji emojis look off


LaurenzV:
Here is a rather extereme example:
![test](https://github.com/RazrFalcon/resvg/assets/47084093/7fdb6b6f-0c9d-44e4-8a9a-ecf6fc3ee949)

For most other cases it's not there / less extreme, but something is still off. Here is how it's supposed to look:
<img width="502" alt="image" src="https://github.com/RazrFalcon/resvg/assets/47084093/080158ed-f87c-449a-8430-50f8403bf1b0">

I still have to figure out whether it's a `ttf-parser` or a `font2svg` bug. My plan is to try to implement `font2svg` for `skrifa` and see whether the issue exists there as well, in which case it's most likely due to the gradient limitations in SVG.

RazrFalcon:
Yes, we have to check for Sweep gradients first.

LaurenzV:
Can be closed now. 😄

PR #778, opened 2024-06-10
Fix bug in colr conversion

LaurenzV:
Turns out I just forgot to copy paste the gradient transform code from the linear gradient... Now they look as expected :)

test

`hb-view` for comparison:
image


LaurenzV:
Fixes #777.

yisibl:
I don't seem to see the font you screenshot in the test, can you add the Segoe UI Emoji font to the test?

LaurenzV:
Those are by Microsoft, I'm pretty sure we can't just add those to the repo.

Issue #760, opened 2024-05-12
PNG output rendering issue


aquaductape:
The input svg file is a circle with gradients. The issue is that on the output png file, on the very right of the circle the gradient is cut off showing a sharp line.

Input svg file.
<img width="83" alt="Screenshot 2024-05-01 at 3 12 20 PM" src="https://github.com/yisibl/resvg-js/assets/29286430/e1fdc96f-5c6b-49a2-8d12-62d12721aa15">
Output png file.
<img width="80" alt="Screenshot 2024-05-01 at 3 13 08 PM" src="https://github.com/yisibl/resvg-js/assets/29286430/c42ae2a5-f65f-492b-91c7-5ec9d2af6b99">

Input svg code.
```svg
<svg
  data-hk="undefinedNaN-0"
  viewBox="0 0 80 80"
  fill="none"
  role="img"
  xmlns="http://www.w3.org/2000/svg"
  width="80"
  height="80"
>
  <!--$--><!--/-->
  <defs>
    <filter
      id="filter_0"
      filterUnits="userSpaceOnUse"
      color-interpolation-filters="sRGB"
    >
      <feFlood flood-opacity="0" result="BackgroundImageFix"></feFlood>
      <feBlend
        in="SourceGraphic"
        in2="BackgroundImageFix"
        result="shape"
      ></feBlend>
      <feGaussianBlur
        stdDeviation="7"
        result="effect1_foregroundBlur"
      ></feGaussianBlur>
    </filter>
  </defs>
  <mask id="0" maskUnits="userSpaceOnUse" x="0" y="0" width="80" height="80">
    <rect width="80" height="80" rx="160" fill="#FFFFFF"></rect>
  </mask>
  <g mask="url(#0)">
    <rect width="80" height="80" fill="#49007E"></rect>
    <path
      filter="url(#filter_0)"
      d="M32.414 59.35L50.376 70.5H72.5v-71H33.728L26.5 13.381l19.057 27.08L32.414 59.35z"
      fill="#FF005B"
      transform="translate(-2 2) rotate(-282 40 40) scale(1.5)"
    ></path>
    <path
      filter="url(#filter_0)"
      style="mix-blend-mode: overlay"
      d="M22.216 24L0 46.75l14.108 38.129L78 86l-3.081-59.276-22.378 4.005 12.972 20.186-23.35 27.395L22.215 24z"
      fill="#FF7D10"
[...]

PR #784, opened 2024-06-27
Make FontResolver a trait and use to provide access to a fontdb

dhardy:
# Summary

`usvg::FontResolver` is changed from a struct to a trait, with a default implementation `DefaultFontResolver`.

`usvg::Options` and `usvg::Tree` no longer contain `fontdb`; instead then `FontResolver` provides access as required.

# Motivation

The existing model:

- Requires that the `fontdb` be shared under an `Arc`, which shouldn't be necessary
- Clones the `fontdb` before modifying
- Makes the modified `fontdb` accessible through `usvg::Tree`, which doesn't appear to be the right place for it

Instead, the user is expected to provide a `&dyn FontResolver` instance in order to process any SVG involving text. As a consequence:

- The user may use an existing `fontdb` instance (without cloning) and may even insert additional fonts directly into this DB (via `RefCell`, `Mutex` or similar)
- The user may retrieve the original `fontdb` instance used without `Arc`
- The minimal set-up with working text processing is slightly more complex (and unfortunately runs into a borrow-checker error if `font_resolver` is not declared before `usvg::Options`).

### DIFF
--- a/crates/resvg/examples/draw_bboxes.rs
+++ b/crates/resvg/examples/draw_bboxes.rs
@@ -15,13 +15,14 @@ fn main() {
         1.0
     };
 
+    let font_resolver = usvg::DefaultFontResolver::with_system_fonts();
+
     let mut opt = usvg::Options::default();
     // Get file's absolute directory.
     opt.resources_dir = std::fs::canonicalize(&args[1])
         .ok()
         .and_then(|p| p.parent().map(|p| p.to_path_buf()));
-
-    opt.fontdb_mut().load_system_fonts();
+    opt.font_resolver = Some(&font_resolver);
 
     let svg_data = std::fs::read(&args[1]).unwrap();
     let tree = usvg::Tree::from_data(&svg_data, &opt).unwrap();
--- a/crates/resvg/examples/minimal.rs
+++ b/crates/resvg/examples/minimal.rs
@@ -6,13 +6,14 @@ fn main() {
     }
 
     let tree = {
+        let font_resolver = usvg::DefaultFontResolver::with_system_fonts();
+
         let mut opt = usvg::Options::default();
         // Get file's absolute directory.
         opt.resources_dir = std::fs::canonicalize(&args[1])
             .ok()
             .and_then(|p| p.parent().map(|p| p.to_path_buf()));
-
-        opt.fontdb_mut().load_system_fonts();
+        opt.font_resolver = Some(&font_resolver);
 
         let svg_data = std::fs::read(&args[1]).unwrap();
         usvg::Tree::from_data(&svg_data, &opt).unwrap()
--- a/crates/resvg/src/main.rs
+++ b/crates/resvg/src/main.rs
@@ -562,7 +562,7 @@ fn parse_args() -> Result<Args, String> {
         image_rendering: args.image_rendering,
         default_size,
         image_href_resolver: usvg::ImageHrefResolver::default(),
-        font_resolver: usvg::FontResolver::default(),
+        font_resolver: Some(usvg::FontResolver::default()),
         fontdb: Arc::new(fontdb::Database::new()),
     };
 
--- a/crates/resvg/src/main.rs
+++ b/crates/resvg/src/main.rs
@@ -562,7 +562,7 @@ fn parse_args() -> Result<Args, String> {
         image_rendering: args.image_rendering,
         default_size,
         image_href_resolver: usvg::ImageHrefResolver::default(),
-        font_resolver: Some(usvg::FontResolver::default()),
+        font_resolver: Some(&usvg::DefaultFontResolver),
         fontdb: Arc::new(fontdb::Database::new()),
     };
 
--- a/crates/resvg/src/main.rs
+++ b/crates/resvg/src/main.rs
@@ -5,9 +5,8 @@
 #![allow(clippy::uninlined_format_args)]
 
 use std::path;
-use std::sync::Arc;
 
-use usvg::fontdb;
+use usvg::{fontdb, FontResolver};
 
 fn main() {
     if let Err(e) = process() {
@@ -31,7 +30,7 @@ where
 }
 
 fn process() -> Result<(), String> {
-    let mut args = match parse_args() {
+    let args = match parse_args() {
         Ok(args) => args,
         Err(e) => {
             println!("{}", HELP);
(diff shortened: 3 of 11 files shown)


RazrFalcon:
@LaurenzV

RazrFalcon:
Do I understand correctly that your approach doesn't allow sharing fontdb between threads? That's one of the requirements.

LaurenzV:
I'll delegate to @laurmaedje. :p

dhardy:
> Do I understand correctly that your approach doesn't allow sharing fontdb between threads? That's one of the requirements.

I think it should but haven't tested. Specifically, there is a `Sync` bound on `FontResolver` (to make`&dyn FontResolver: Send`) and the user may implement any type of lock they wish.

`DefaultFontResolver` is not really suitable for this, but it doesn't need to be.

RazrFalcon:
>DefaultFontResolver is not really suitable for this, but it doesn't need to be.

Both the current and the old implementations allowed sharing `fontdb` between threads.
But in this patch you're simply cloning fontdb in `crates/resvg/tests/integration/main.rs` This is far from ideal.

LaurenzV:
Just so you know I briefly talked to @laurmaedje and he said he would comment why he implemented it the way he did, although I'm not sure when he will get to it. But might be good to wait a bit before merging.

RazrFalcon:
Ok, will wait then. I do not want to break typst again.

laurmaedje:
For context, the current implementation comes from https://github.com/RazrFalcon/resvg/pull/769. This implementation was preceded by the earlier PR https://github.com/RazrFalcon/resvg/pull/754.

The design in this PR is closer to #754 than #760. The discussion of that PR provides some insights into problems, but I'll try to summarize things here.

The main conceptual problem, from my understanding, is [how support for fonts in embedded in SVGs would work](https://github.com/RazrFalcon/resvg/pull/754#issuecomment-2115946216). In the current model, since usvg manages the database, it can mutate it itself. The `Arc` optimizes for the common case, but conceptually each SVG has its own database and that's also why it ends up in the tree.

On another note, if I'm not mistaken the implementation in this PR doesn't really allow for lazy font loading. Since both `select_font` and `select_fallback` are `&self`, we'd have to resort to interior mutability on the `Database`, but that's incompatible with the method `fontdb(&self) -> &Database`. There is some discussion about this problem in #754, but all workarounds I found are ugly.

If I may ask: Which concrete use case is the motivation for changing things again? The `Arc` is very cheap and the deep-clone of the `Database` only happens if you _both_ provide fonts up-front and on-demand. And even then, cloning a `Database` isn't _that_ costly since the font data itself is reference counted internally or memory-mapped on-demand. Moreover, while having the fontdb in the `usvg::Tree` seems a bit weird perhaps, it makes the user-facing API much simpler and harder to use in the wrong way. 

The one somewhat costly thing I can see is that if you parse a bunch of SVGs and lazy load, the fontdb is populated over and over again. But you can cache things in the background and keep that relatively cheap (in the grand scheme of what usvg does I would be surprised if this was a large factor). Moreover, it helps a bit with correctness because in the current design it's harder to mess things up in a way where parsing of one SVG affects font selection of another SVG (because of an already populated database).

RazrFalcon:
>Moreover, while having the fontdb in the usvg::Tree seems a bit weird perhaps, it makes the user-facing API much simpler and harder to use in the wrong way.

Another reason for this is that in the future we will support fonts embedded in CSS, which must not be added to the "global" fontdb. That's why we need an ability to deeply clone the database on-demand.

dhardy:
> On another note, if I'm not mistaken the implementation in this PR doesn't really allow for lazy font loading. Since both `select_font` and `select_fallback` are `&self`, we'd have to resort to interior mutability on the `Database`,

The intention is that a wrapper type be used to manage that lock... which also requires an associated type (something I forgot to add previously)...
```rust
pub trait FontResolver: Debug + Sync {
    type DbGuard<'a>: std::ops::Deref<Target = Database>;
    fn fontdb(&self) -> Self::DbGuard<'_>;
    // ...
}

#[derive(Debug)]
struct LockingFontResolver {
    db: std::sync::Mutex<Database>,
}
impl FontResolver for LockingFontResolver {
    type DbGuard<'a> = std::sync::MutexGuard<'a, Database>;
    fn fontdb(&self) -> Self::DbGuard<'_> {
        self.db.lock().unwrap()
    }
}
```
... and thus usage of `dyn FontResolver` becomes problematic. (We probably don't want to add generic parameters to all usages.)

A better option might be to do this:
```rust
pub trait FontResolver: Debug + Sync {
    fn with_fontdb_dyn(&self, f: Box<dyn FnOnce(&Database) + '_>);
    // ...
}

pub trait FontResolverExt: FontResolver {
    fn with_fontdb<R>(&self, f: impl (FnOnce(&Database) -> R)) -> Option<R> {
        let mut result = None;
        let out = &mut result;
        self.with_fontdb_dyn(Box::new(|db| *out = Some(f(db))));
        result
    }
}
impl<FR: FontResolver + ?Sized> FontResolverExt for FR {}
```
Then we can still use `dyn FontResolver`:
```rust
pub(crate) fn convert(text: &mut Text, resolver: &dyn FontResolver) -> Option<()> {
    let (text_fragments, bbox) = layout::layout_text(text, resolver)?;
    text.layouted = text_fragments;
    text.bounding_box = bbox.to_rect();
    text.abs_bounding_box = bbox.transform(text.abs_transform)?.to_rect();

    let (group, stroke_bbox) = resolver.with_fontdb(|fontdb| flatten::flatten(text, fontdb))??;
    text.flattened = Box::new(group);
    text.stroke_bounding_box = stroke_bbox.to_rect();
    text.abs_stroke_bounding_box = stroke_bbox.transform(text.abs_transform)?.to_rect();

    Some(())
}
```

dhardy:
> Another reason for this is that in the future we will support fonts embedded in CSS, which must not be added to the "global" fontdb. That's why we need an ability to deeply clone the database on-demand.

Admittedly, this is a big motivation to keep the existing design (passing `Arc<Database>` in). Especially if you have another mechanism for caching loaded fonts.

laurmaedje:
> The intention is that a wrapper type be used to manage that lock... which also requires an associated type (something I forgot to add previously)...

> ... and thus usage of dyn FontResolver becomes problematic. (We probably don't want to add generic parameters to all usages.)

I went down the same path of thoughts during #754 and ended up with a very similar approach [here](https://github.com/laurmaedje/resvg/commit/c11f38e116dacf2cc7972b52e65c368fdb14313d#diff-71defd6fedd63f26d8897212299f074809606b4149146221668ee4584d682145), but I think it's far from pretty.

dhardy:
What if `Options` contains an `Arc<dyn FontResolver>` instead (where `FontResolver` is some trait impl'd by `fontdb::Database`)? An `Arc<fontdb::Database>` will upcast appropriately, as will a user-defined type `T` implementing that trait.

I'd also be satisfied with `Arc<T>` where `T: Borrow<fontdb::Database>`, but I understand not wanting to use generics.

laurmaedje:
Could you elaborate more on your motivation? The three bullet points from the original post all sound like conceptual concerns rather than motivating a specific use case that isn't currently achievable.

dhardy:
To further explain my **motivation**: [`kas_text::fonts::Database`](https://docs.rs/kas-text/latest/kas_text/fonts/struct.Database.html) is a struct wrapping `fontdb::Database` with a set of font aliases. This is used to implement [`FontSelector::select`](https://github.com/kas-gui/kas-text/blob/master/src/fonts/selector.rs#L402), where `FontSelector` is roughly akin to `usvg::Font` (i.e. font family name, weight, style).

Ideally therefore I would implement [`usvg::FontSelectionFn`](https://docs.rs/usvg/latest/usvg/type.FontSelectionFn.html) as a wrapper around my `FontSelector::select` which requires access to the `aliases`, which, in my case, are stored alongside the `fontdb::Database`.

dhardy:
It seems this approach is viable. Work on the font management side is here: https://github.com/kas-gui/kas-text/pull/87. (The SVG font resolver is WIP.)

Issue #781, opened 2024-06-16
Try to resize large JPEGs during decoding


RazrFalcon:
JPEG supports decoding an image at a smaller resolution, which is faster and uses less memory.
This can be useful when we're trying to decode a very large JPEG just to downscale it using `tiny-skia` anyway.

PR #789, opened 2024-07-08
Use zune-jpeg for JPEG decoding

### LINKED ISSUE #782: Switch to zune-jpeg
Apparently `jpeg-decoder` is somewhat deprecated.

Additionally:

- [ ] try to decode JPEG directly to `tiny_skia::Pixmap` to avoid needless allocations
- [ ] try to use JPEG-native resizing when possible #781

mattfbacon:
This is a draft because I haven't implemented #781 yet.

Closes #782.


mattfbacon:
Actually I don't think zune-jpeg supports that anyway.

RazrFalcon:
#781 is way out of scope. It would take a lot of work to implement. Just switching to zune-jpeg is more than enough.

PR #797, opened 2024-07-31
Simplify transform logic for glyphs

LaurenzV:
This PR adds a method to get the transform of a glyph independent of the actual table that will be used to render the glyph. This allows user to perform their own logic for glyph selection (which is the case for me when rendering glyphs via PDF). The specific glyph transform methods use this new method as the base.

### DIFF
--- a/crates/usvg/src/text/layout.rs
+++ b/crates/usvg/src/text/layout.rs
@@ -49,34 +49,31 @@ pub struct PositionedGlyph {
 }
 
 impl PositionedGlyph {
+    /// Returns the transform of glyph.
+    pub fn transform(&self) -> Transform {
+        let sx = self.font_size / self.units_per_em as f32;
+
+        self.span_ts
+            .pre_concat(self.cluster_ts)
+            .pre_concat(Transform::from_scale(sx, sx))
+            .pre_concat(self.glyph_ts)
+    }
+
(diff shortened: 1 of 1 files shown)


RazrFalcon:
So you have changed the code you wrote before? Because I do not recognize it. 
Can you explain why we had multiple transform methods before and now just one is ok?

LaurenzV:
Yes, that's the code I wrote before. We still have multiple transform methods but the difference is that I extracted the transforms common to all methods (i.e. span transform + cluster transform + glyph transform, you need to apply those regardless of whether you draw an outline or a bitmap glyph) into a separate method, and then each specific transform method calls those and applies additional transforms on top of that.

The reason I didn't do this the first time is that I was unable to disentangle the `ts.pre_scale(1.0, -1.0)` transform for outline methods from the other parts, as I got a test failure. But yesterday, after looking it again, I realized that this was because in the layout method, there was a transform that was applied wrongly (namely `let ts = Transform::from_translate(x + glyph.dx as f32, glyph.dy as f32);`, since in SVG coordinates and glyph coordinates the y axis is flipped, this should be `-glyph.dy`. And after making this change, I was able to disentangle it with all test cases still passing.

RazrFalcon:
Got it. Thanks!

PR #799, opened 2024-08-04
feat: add raw image to allow flexible resolvers

### LINKED ISSUE #772: usvg: allow custom image decoding in FormatKind
Currently, only JPEG, PNG and GIF are supported. A `FormatKind::Custom` or `FormatKind::Pixmap` would allow for greater flexibility and the clients can provide their own image decoding.

I'll be happy to make a PR if this seems like a good idea.

amhndu:
Adds a RawImage image kind to usvg and a transparent decode_raster implementation in resvg. This allows writing flexible ImageHrefResolver that can support additional image formats.  Another use case would be to allow alternate implementations, but with zunejpeg merged, I find it less important.

Closes #772 

I'm unsure about the xml writer implementation, but I've added a hack for now

### DIFF
--- a/crates/resvg/src/image.rs
+++ b/crates/resvg/src/image.rs
@@ -56,6 +56,8 @@ fn render_vector(
 
 #[cfg(feature = "raster-images")]
 mod raster_images {
+    use usvg::RawImage;
+
     use crate::OptionLog;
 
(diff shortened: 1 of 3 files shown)


amhndu:
I'm happy to refactor and follow whatever approach you have in mind

PR #812, opened 2024-09-01
Add the cargo-c metadata

### LINKED ISSUE #592: Provide pkg-config file with the C library
Providing a [pkg-config file](https://en.wikipedia.org/wiki/Pkg-config) makes it easier for users of the library to actually find it, check its version and get the correct compiler and linker flags.

[cargo-c](https://github.com/lu-zero/cargo-c) could be helpful here for generating the pkg-config file and also to provide a mechanism for installing it, the library itself and the header. It also ensures that the library has a valid and correct soname set, which cargo doesn't set at all when building a `cdylib`.

`cargo-c` also allows running cbindgen automatically but AFAIU you're manually editing the output of it so that's maybe not very useful here.

lu-zero:
Fixes #592

The default `cargo cinstall --destdir some/path` produces on macos:
```
└── usr
    └── local
        ├── include
        │   └── resvg
        │       └── resvg.h
        └── lib
            ├── libresvg.0.43.0.dylib
            ├── libresvg.0.43.dylib -> libresvg.0.43.0.dylib
            ├── libresvg.a
            ├── libresvg.dylib -> libresvg.0.43.0.dylib
            └── pkgconfig
                └── resvg.pc
```

### DIFF
--- a/crates/c-api/Cargo.toml
+++ b/crates/c-api/Cargo.toml
@@ -26,3 +26,10 @@ system-fonts = ["resvg/system-fonts"]
 # enables font files memmaping for faster loading (only for `text`)
 memmap-fonts = ["resvg/memmap-fonts"]
 raster-images = ["resvg/raster-images"]
+capi = []
+
+[package.metadata.capi.header]
+generation = false
+
+[package.metadata.capi.install.include]
+asset = [{ from="resvg.h" }]


RazrFalcon:
Can you elaborate what `cinstall` is and how it should be used?

lu-zero:
It is part of [cargo-c](https://crates.io/crates/cargo-c), it builds and install the C-API crate as a normal C-library (both static and shared) including the pkg-config file. 

It uses the metadata I added to Cargo.toml to find the crate(s) that should be made libraries and optionally generates the C header or install pre-made ones.

RazrFalcon:
Can you add a note about it to `crates/c-api/README.md`? So people would know how to use it.
I don't want to add a random stuff to Cargo.toml. Since it will only confuse people.

lu-zero:
Here an update with another commit to convert the example to use the uninstalled pkg-config.

RazrFalcon:
Thanks.

PR #813, opened 2024-09-02
impl From for FontStretch, FontStyle

### LINKED ISSUE #808: Deduplicate fontdb types
`resvg` optionally depends on `fontdb`. Meanwhile,

- `usvg::FontStyle` is identical to `fontdb::Style`
- `usvg::FontStretch` is identical to `fontdb::Stretch`

Worse, there isn't even a conversion method between the two. This is a pain when implementing a `usvg::FontResolver`.

dhardy:
Closes #808.

I added a CI test for `usvg` with all features since this was not present.

Note also:

- `fontdb::Weight` could include `From<u16>`, but this is relatively unimportant
- `svgtypes::FontFamily` could implement `From<fontdb::FontFamily>`; the reverse would need to be `fontdb::FontFamily<'a>: From<&'a svgtypes::FontFamily>` (omitted since this is a different repo)
- `fontdb::Query<'a>` cannot implement `From<usvg::Font>` (or `&'a usvg::Font`) due to lifetime of the list of `families`.

### DIFF
--- a/crates/usvg/src/tree/text.rs
+++ b/crates/usvg/src/tree/text.rs
@@ -33,6 +33,40 @@ impl Default for FontStretch {
     }
 }
 
+#[cfg(feature = "text")]
+impl From<fontdb::Stretch> for FontStretch {
+    fn from(stretch: fontdb::Stretch) -> Self {
+        match stretch {
+            fontdb::Stretch::UltraCondensed => FontStretch::UltraCondensed,
+            fontdb::Stretch::ExtraCondensed => FontStretch::ExtraCondensed,
+            fontdb::Stretch::Condensed => FontStretch::Condensed,
+            fontdb::Stretch::SemiCondensed => FontStretch::SemiCondensed,
+            fontdb::Stretch::Normal => FontStretch::Normal,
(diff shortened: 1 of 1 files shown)


dhardy:
Also note: the two enum conversions included here are the only bits I particularly wanted anyway.

.github/workflows/main.yml
```diff
@@ -32,6 +32,10 @@ jobs:
       working-directory: crates/usvg
       run: cargo check --no-default-features
 
+    - name: Build usvg with all features
```
RazrFalcon, line 35: `usvg` will all features will be build during `resvg` testing.

crates/usvg/src/tree/text.rs
```diff
@@ -33,6 +33,40 @@ impl Default for FontStretch {
     }
 }
 
+#[cfg(feature = "fontdb")]
```
RazrFalcon, line 36: This should be `text`, not `fontdb`.

.github/workflows/main.yml
dhardy: Removed

PR #820, opened 2024-09-13
Add support for the `background-color` attribute

LaurenzV:


RazrFalcon:
Is this task finished?

LaurenzV:
I think so? 😄

RazrFalcon:
Good. Just wanted to clarify.

crates/resvg/tests/integration/main.rs
```diff
@@ -55,10 +57,14 @@ pub fn render(name: &str) -> usize {
     );
     resvg::render(&tree, render_ts, &mut pixmap.as_mut());
 
-    // pixmap.save_png(&format!("tests/{}.png", name)).unwrap();
-    // Command::new("oxipng")
-    //     .args(["-o".to_owned(), "6".to_owned(), "-Z".to_owned(), format!("tests/{}.png", name)])
-    //     .output().unwrap();
+    if !Path::new(&png_path).exists() {
```
RazrFalcon, line 60: Let's revert to the old implementation. I want this behavior to be manually trigger-able.

crates/usvg/src/parser/converter.rs
```diff
@@ -7,15 +7,15 @@ use std::hash::{Hash, Hasher};
...
-#[cfg(feature = "text")]
-use fontdb::Database;
-use svgtypes::{Length, LengthUnit as Unit, PaintOrderKind, TransformOrigin};
-
 use super::svgtree::{self, AId, EId, FromValue, SvgNode};
 use super::units::{self, convert_length};
 use super::{marker, Error, Options};
 use crate::parser::paint_server::process_paint;
 use crate::*;
+#[cfg(feature = "text")]
```
RazrFalcon, line 15: Keep at the old position.

Are you manually moving import or is it IDE/formatter?

crates/usvg/src/parser/converter.rs
```diff
@@ -389,6 +404,29 @@ pub(crate) fn convert_doc(svg_doc: &svgtree::Document, opt: &Options) -> Result<
     Ok(tree)
 }
 
+fn background_path(background_color: svgtypes::Color, area: Rect) -> Option<Path> {
+    let path = {
+        let mut builder = PathBuilder::new();
```
RazrFalcon, line 409: We have `PathBuilder::from_rect`

crates/usvg/src/parser/svgtree/names.rs
```diff
@@ -665,15 +661,16 @@ impl std::fmt::Display for AId {
 struct Map<V: 'static> {
     pub key: u64,
     pub disps: &'static [(u32, u32)],
-    pub entries: &'static [(&'static str, V)],
+    pub entries: &'static[(&'static str, V)],
```
RazrFalcon, line 664: We should update `crates/usvg/codegen/main.rs` as well then.

crates/usvg/src/parser/converter.rs
LaurenzV: Must be the formatter. :(

crates/resvg/tests/integration/main.rs
LaurenzV: What do you think about what I did now? Then you can simply do `REPLACE="" cargo test` to indiciate that you want to replace reference images. This way, you don't always have to comment/uncomment it.

Or should I still revert completely?

RazrFalcon: It's fine.

RazrFalcon:
Thanks!

yisibl:
This is a bit strange, there is also a shorthand `background` property in CSS. Should we support it?

PR #832, opened 2024-10-13
Fix issue with relative units in symbols

### LINKED ISSUE #829: Weird black bar in rendered SVG
Originally reported at https://github.com/typst/typst/issues/5174.

With the following SVG:

test

`resvg` renders it as:
test

One problem is obviously the different font, but this is a known issue since we don't support `font-face` yet. However, another issue is that there is this weird black box for one path. In Chrome, it looks like that instead:
image

I'll try to get a MRE when I find the time to do so.

LaurenzV:
Fixes #829

I'm not a 100% sure about this fix, so some scrutiny would be appreciated. With that said, I think there were two things going on:

- When resolving relative sizes (like percentages), we only use the width/height of the viewbox, but I think when using a use node, we should use the width/height of that as the reference, if available.
- We did not set the `use_size` attribute for use nodes linked to symbols. 

Let me know what you think.

Given the following SVG:
```svg
<svg version="1.1" viewBox="0 0 200 200" xmlns="http://www.w3.org/2000/svg">
    <title>With size on `use` and relative units</title>
    <symbol id="image1">
        <image width="100%" height="100%"
               href="data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAMgAAADICAYAAACtWK6eAAAFKUlEQVR4Xu3VsRWDQAwFQa5/6nWKGzAONv1DrkCj28d5Pvdz+QgQ+ClwBOJlEHgXEIjXQeCPgEA8DwIC8QYINAF/kOZmakRAICOHtmYTEEhzMzUiIJCRQ1uzCQikuZkaERDIyKGt2QQE0txMjQgIZOTQ1mwCAmlupkYEBDJyaGs2AYE0N1MjAgIZObQ1m4BAmpupEQGBjBzamk1AIM3N1IiAQEYObc0mIJDmZmpEQCAjh7ZmExBIczM1IiCQkUNbswkIpLmZGhEQyMihrdkEBNLcTI0ICGTk0NZsAgJpbqZGBAQycmhrNgGBNDdTIwICGTm0NZuAQJqbqREBgYwc2ppNQCDNzdSIgEBGDm3NJiCQ5mZqREAgI4e2ZhMQSHMzNSIgkJFDW7MJCKS5mRoREMjIoa3ZBATS3EyNCAhk5NDWbAICaW6mRgQEMnJoazYBgTQ3UyMCAhk5tDWbgECam6kRAYGMHNqaTUAgzc3UiIBARg5tzSYgkOZmakRAICOHtmYTEEhzMzUiIJCRQ1uzCQikuZkaERDIyKGt2QQE0txMjQgIZOTQ1mwCAmlupkYEBDJyaGs2AYE0N1MjAgIZObQ1m4BAmpupEQGBjBzamk1AIM3N1IiAQEYObc0mIJDmZmpEQCAjh7ZmExBIczM1IiCQkUNbswkIpLmZGhEQyMihrdkEBNLcTI0ICGTk0NZsAgJpbqZGBAQycmhrNgGBNDdTIwICGTm0NZuAQJqbqREBgYwc2ppNQCDNzdSIgEBGDm3NJiCQ5mZqREAgI4e2ZhMQSHMzNSIgkJFDW7MJCKS5mRoREMjIoa3ZBATS3EyNCAhk5NDWbAICaW6mRgQEMnJoazYBgTQ3UyMCAhk5tDWbgECam6kRAYGMHNqaTUAgzc3UiIBARg5tzSYgkOZmakRAICOHtmYTEEhzMzUiIJCRQ1uzCQikuZkaERDIyKGt2QQE0txMjQgIZOTQ1mwCAmlupkYEBDJyaGs2AYE0N1MjAgIZObQ1m4BAmpupEQGBjBzamk1AIM3N1IiAQEYObc0mIJDmZmpEQCAjh7ZmExBIczM1IiCQkUNbswkIpLmZGhEQyMihrdkEBNLcTI0ICGTk0NZsAgJpbqZGBAQycmhrNgGBNDdTIwICGTm0NZuAQJqbqREBgYwc2ppNQCDNzdSIgEBGDm3NJiCQ5mZqREAgI4e2ZhMQSHMzNSIgkJFDW7MJCKS5mRoREMjIoa3ZBATS3EyNCAhk5NDWbAICaW6mRgQEMnJoazYBgTQ3UyMCAhk5tDWbgECam6kRAYGMHNqaTUAgzc3UiIBARg5tzSYgkOZmakRAICOHtmYTEEhzMzUiIJCRQ1uzCQikuZkaERDIyKGt2QQE0txMjQgIZOTQ1mwCAmlupkYEBDJyaGs2AYE0N1MjAgIZObQ1m4BAmpupEQGBjBzamk1AIM3N1IiAQEYObc0mIJDmZmpEQCAjh7ZmExBIczM1IiCQkUNbswkIpLmZGhEQyMihrdkEBNLcTI0ICGTk0NZsAgJpbqZGBAQycmhrNgGBNDdTIwICGTm0NZuAQJqbqREBgYwc2ppNQCDNzdSIgEBGDm3NJiCQ5mZqREAgI4e2ZhMQSHMzNSIgkJFDW7MJCKS5mRoREMjIoa3ZBATS3EyNCAhk5NDWbAICaW6mRgQEMnJoazYBgTQ3UyMCAhk5tDWbgECam6kRAYGMHNqaTUAgzc3UiIBARg5tzSYgkOZmakRAICOHtmYTEEhzMzUi8AVw2rznM39f5AAAAABJRU5ErkJggg=="></image>
    </symbol>

    <g transform="translate(25 50)">
        <use href="#image1" width="150" height="100"></use>
    </g>

    <!-- image frame -->
</svg>
```

Before:
test

After, which also matches Chrome:
test


RazrFalcon:
A yet another `symbol` bug? What a surprise. One the of the worst SVG features.

As for a solution, hard to say. As long as it doesn't break existing tests it should be fine, but it feels like a hack. No idea how it should be handled properly.

crates/usvg/src/parser/units.rs
```diff
@@ -31,7 +31,8 @@ pub(crate) fn convert_length(
             if object_units == Units::ObjectBoundingBox {
                 n / 100.0
             } else {
-                let view_box = state.view_box;
+                let width = state.use_size.0.unwrap_or(state.view_box.width());
```
RazrFalcon, line 34: Let's keep calling those values as `viewbox_width` and `viewbox_height`.

crates/usvg/src/parser/use_node.rs
```diff
@@ -53,7 +54,33 @@ pub(crate) fn convert(
...
+        if node.has_attribute(AId::Width) {
+            use_state.use_size.0 = Some(node.convert_user_length(AId::Width, &use_state, def));
+        }
+        if node.has_attribute(AId::Height) {
+            use_state.use_size.1 = Some(node.convert_user_length(AId::Height, &use_state, def));
+        }
+    };
+
     if linked_to_symbol {
+        set_use_size(&mut use_state);
```
RazrFalcon, line 82: Let's add a comment like with a call below, explaining why we need it.

RazrFalcon:
Can we simply override the viewbox size like we do in `convert_svg`? Then we would not need any changed to the units code. Or does it break something?

LaurenzV:
Hmm, I'm not sure though, because a view box conceptually has a x/y coordinate, too, does it really make sense to just replace the width/height of the view box in this case?

LaurenzV:
Better now?

RazrFalcon:
All good. Thanks!

PR #833, opened 2024-10-14
Add support for additional image-rendering attributes

### LINKED ISSUE #828: Support more values for imageRendering
As a standard SVG attribute, only `optimizeSpeed` and `optimizeQuality` are supported (see https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/image-rendering), however, as mentioned in the note there, it can also be used as a CSS property with the following possible values:

[smooth](https://developer.mozilla.org/en-US/docs/Web/CSS/image-rendering#smooth)

    The image should be scaled with an algorithm that maximizes the appearance of the image. In particular, scaling algorithms that "smooth" colors are acceptable, such as bilinear interpolation. This is intended for images such as photos.
[high-quality](https://developer.mozilla.org/en-US/docs/Web/CSS/image-rendering#high-quality)

    Identical to smooth, but with a preference for higher-quality scaling. If system resources are constrained, images with high-quality should be prioritized over those with any other value, when considering which images to degrade the quality of and to what degree.
[crisp-edges](https://developer.mozilla.org/en-US/docs/Web/CSS/image-rendering#crisp-edges)

    The image is scaled with an algorithm such as "nearest neighbor" that preserves contrast and edges in the image. Generally intended for images such as pixel art or line drawings, no blurring or color smoothing occurs.
[pixelated](https://developer.mozilla.org/en-US/docs/Web/CSS/image-rendering#pixelated)

    The image is scaled with the "nearest neighbor" or similar algorithm to the nearest integer multiple of the original image size, then uses smooth interpolation to bring the image to the final desired size. This is intended to preserve a "pixelated" look without introducing scaling artifacts when the upscaled resolution isn't an integer multiple of the original.

Here is an example SVG that uses `pixelated`:
test

LaurenzV:
Fixes #828


RazrFalcon:
Not sure if we should update `docs/svg2-changelog.md`, Painting section, as well. Does it clarify as SVG 2 feature? No one knows... But I guess so.

We should also update CLI help for resvg/usvg. The `--image-rendering` flag.

RazrFalcon:
Also, aren't `pixelated` requires integer scaling?
https://developer.mozilla.org/en-US/docs/Web/CSS/image-rendering#pixelated

LaurenzV:
> Also, aren't pixelated requires integer scaling?

Looks like it, but `tiny-skia` doesn't support that, right? I basically just used the closest match we have for each attribute, it's possible that some are not 100% as required by the spec. Should I document that somewhere or?

RazrFalcon:
I don't think `tiny-skia` should support this. This is purely a resvg/usvg task. We have to readjust/reset `image` size. At least this is my understanding. I haven't looked deep into it.

The only way to "document" it is to create a test that would be failing.

I personally thought the whole point of this feature is `pixelated`, otherwise you could achieve the same with `imageRendering`. Or is it purely to support CSS variants?

LaurenzV:
Yes, the main intention was to support the CSS attributes in the first place, so that you don’t end up with bicubic rendering if “pixelated” was requested.

But I will see whether I can figure it out.

LaurenzV:
Sent you an e-mail with the content "animation-charred-annoying-otter".

LaurenzV:
I think getting pixelated 100% correct is going to be very difficult. I think what we would have to do is

1. Determine the actual size the image will take up in the resulting image. 
2. Find the best integer that scales to that size. Draw the image to the pixmap with a transform to scale to that size, using "Nearest".
3. Turn the resulting pixmap into an image.
4. Proceed as with all other rendering modes, but now we use bilinear rendering for the "final smoothing".

But this is pretty hard because it requires us to change the transform in the parent group of the image that scales the raster image to the intended size. 

Would it be okay for you if I simply remark in the source code that pixelated doesn't work completely correctly yet?

RazrFalcon:
Yes, we can "skip" `pixelated` as part of this commit.

As for parent transforms, yes, I'm not sure what is the expected output here. We need a bunch of tests to figure it out. 
If the spec expects the image to be pixelated in canvas coordinates and not object coordinates (not transforms) then we're screwed. That would be a serious change.

And I'm not even sure what we're expected to do when rotation and/or skew transforms are present.
Preserving integer scaling in all situation would be hard.

RazrFalcon:
Let's add `image-rendering` to `docs/svg2-changelog.md` and note that `pixelated` is partially supported there.

RazrFalcon:
Thanks!

PR #835, opened 2024-10-19
Upgrade image-webp to 0.2.x get bugfixes and major optimizations

Shnatsel:
I know that PRs for updates to dependencies usually aren't welcome, but this cannot be taken care of by a `cargo update` because 0.1.x to 0.2.x is a semver-incompatible change, and needs to be done in resvg's `Cargo.toml`

Changelog: https://github.com/image-rs/image-webp/blob/main/CHANGES.md

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -146,9 +146,9 @@ dependencies = [
 
 [[package]]
 name = "image-webp"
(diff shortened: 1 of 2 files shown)


RazrFalcon:
Ok, I'll accept.

Issue #838, opened 2024-10-29
Relicense resvg under dual MIT/Apache-2.0


DJMcNab:
> [!NOTE]
>
> resvg has recently moved into the Linebender organisation. See #834 for details.
>
> We would also like to thank @RazrFalcon for all their previous work on resvg, and wish them luck in their future endeavours.

## What

As part of the move into Linebender, we are planning on relicensing resvg under the MIT/Apache 2.0 licenses.
These match the standard licenses used by Linebender, and it gives [maximum compatibility](https://rust-lang.github.io/api-guidelines/necessities.html#crate-and-its-dependencies-have-a-permissive-license-c-permissive) with the rest of the Rust ecosystem.

**If you are mentioned in this issue, we need your help to make this happen**

To agree to this relicense, please read the details in this issue, then leave a comment with the following message:

```
I license past and future contributions under the dual MIT/Apache-2.0 license, allowing licensees to choose either at their option.
```

This also follows a similar move done by wgpu. See [their issue](https://github.com/gfx-rs/wgpu/issues/1473) for their motivations, which broadly also apply to this project.

## MPL 2.0

For clarity, we are planning to remove the MPL 2.0 license requirement for new contributions, and so will no longer be releasing new code under the MPL.
This follows the precedent set by wgpu.
This will allow using code released under the standard MIT/Apache 2.0 license in this project.

For existing users, the Apache license is [documented as compatible](https://www.mozilla.org/en-US/MPL/2.0/FAQ/#mpl-bsd-and-apache) with the MPL,
so we expect that this change will only make resvg easier to use.
Additionally, this change does not (and cannot) remove the licenses for previous releases and commits, and so you can continue to use release 0.44.0 and earlier
under the MPL 2.0.

## Contributor checklist

- [x] @RazrFalcon
- [x] @LaurenzV
- [x] @gentoo90
- [x] @laurmaedje
- [x] `@ahaoboy`
- [x] `@akindle`
- [x] `@AnthonyMikh`
- [x] `@antmelnyk`
- [x] `@benoit-pierre` - reached by email by @DJMcNab 
- [x] `@CGMossa`
- [x] `@chubei-oppen`
- [x] `@conorgolden1`
- [x] `@CosmicHorrorDev`
- [x] `@dabreegster`
- [x] `@dhardy`
- [x] `@e00E`
- [x] `@enkore`
- [x] `@EpicEricEE`
- [ ] `@ferdnyc`
- [x] `@flxzt`
[...]

PR #842, opened 2024-11-07
Set `rust-version` for workspace members

sorairolake:
The edition of all crates in this repository are Rust 2021 (requires Rust 1.56.0 or later). So we can set the [`rust-version`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field) field for the `Cargo.toml` file. This makes it easier to check the crate's MSRV.

I didn't feel the need to set this field for crates which aren't published on [crates.io](https://crates.io/), so I didn't set the field for `codegen` and `explorer-thumbnailer` crates.

### DIFF
--- a/crates/c-api/Cargo.toml
+++ b/crates/c-api/Cargo.toml
@@ -5,6 +5,7 @@ authors = ["Yevhenii Reizner <razrfalcon@gmail.com>"]
 keywords = ["svg", "render", "raster", "c-api"]
 license = "MPL-2.0"
 edition = "2021"
+rust-version = "1.67.1"
 workspace = "../.."
 
 [lib]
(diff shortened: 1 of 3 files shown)


DJMcNab:
Welcome @sorairolake! 

This PR has arrived at an interesting time for the project, as we're going through a relicensing. To be able to accept this contribution, we're going to need you to agree to that - see #838

One other note, the badge in the README lists the MSRV as 1.67.1. I'd probably set the MSRV to that to start with.

Thank you

sorairolake:
https://github.com/linebender/resvg/issues/838#issuecomment-2464391910

sorairolake:
@DJMcNab The MSRV of `usvg` is Rust 1.65.0 (checked with [cargo-msrv](https://github.com/foresterre/cargo-msrv)), should this also be changed to Rust 1.67.1?

DJMcNab:
When we move to Linebender CI (see https://github.com/linebender/rfcs/pull/5), we will probably want to have a unified MSRV. But since this isn't being checked, just specifying the actual reality is fine.

Thank you!

Issue #252, opened 2020-02-28
Support !important CSS rule


JoKalliauer:
#### What happened?
|lib|pic|
|-|-|
|svg|![svg](https://upload.wikimedia.org/wikipedia/commons/5/58/Test_only.svg)
|ink|![Test only](https://user-images.githubusercontent.com/15687581/75588974-02573100-5a7a-11ea-8308-e0dac1f30c1f.png)
|resvg|![Test only_resvg](https://user-images.githubusercontent.com/15687581/75588983-05eab800-5a7a-11ea-8771-f1d4a8e8e538.png)
|rsvg|![Test only_rsvg](https://user-images.githubusercontent.com/15687581/75588987-08e5a880-5a7a-11ea-9c78-017064e28032.png)

#### What should have happened?
 - I think case 6 the box should be blue

#### Copyright

source https://commons.wikimedia.org/wiki/File:Test_only.svg

author H2g2bob

license  public Domain

#### Version and Operating System:

- Inkscape Version: 1.1dev
- resvg version: 0.9.0
- rsvg Version: 2.40.21
- Operating System: Ubuntu 18.04

RazrFalcon:
Yes, `!important` is not supported.

PR #843, opened 2024-11-07
Add support for the !important attribute

LaurenzV:
This solves the problem I had where I wanted to achieve that I can override _all_ styles in the SVG via stylesheet injection. I tested it with `rsvg-convert` and it works in the same way, so it would be great to have this merged. :)


RazrFalcon:
Are you sure `!important` shouldn't be handled on the `simplecss` side? I genuinely have no idea how it suppose to work.

crates/usvg/src/parser/svgtree/parse.rs
```diff
@@ -284,40 +292,40 @@ pub(crate) fn parse_svg_element<'input>(
     let mut write_declaration = |declaration: &Declaration| {
         // TODO: perform XML attribute normalization
         if declaration.name == "marker" {
-            insert_attribute(AId::MarkerStart, declaration.value);
-            insert_attribute(AId::MarkerMid, declaration.value);
-            insert_attribute(AId::MarkerEnd, declaration.value);
+            insert_attribute(AId::MarkerStart, declaration.value, declaration.important);
```
RazrFalcon, line 295: Too much copy-paste. Can we pass `declaration` directly? Or at least shorthand variables instead of typing `declaration.value, declaration.important` over and over again?

LaurenzV:
> Are you sure !important shouldn't be handled on the simplecss side? I genuinely have no idea how it suppose to work.

I think `simplecss` is low-level enough that it's enough if it provides the information whether an attribute has the `important` value or not, which is what it currently does. And since style deduplication is handled on the `resvg` side right now, we have to do it there anyway.

crates/usvg/src/parser/svgtree/parse.rs
LaurenzV: Don't think it's possible to avoid since we are overriding the value in some cases, but I created a shorthand variable now.

RazrFalcon:
Looks good to me.

RazrFalcon:
#252

LaurenzV:
Will try it.

LaurenzV:
Seems to work as intended. :)

Before:
test
sbahn2

After:
test
sbahn

LaurenzV:
I just noticed this, though:

```
When a declaration is important, the order of precedence is reversed. Declarations marked as important in the user-agent style sheets override all important declarations in the user style sheets. Similarly, all important declarations in the user style sheets override all important declarations in the author's style sheets. Finally, all important declarations take precedence over all animations.
```

So I think I need to adapt the swapping condition a bit. Will do so a bit later.

LaurenzV:
Should be good now.

LaurenzV:
Is this okay to merge? Not sure who to ping, @waywardmonkeys maybe?

waywardmonkeys:
@LaurenzV Well, I think @RazrFalcon said it looks good, so I think I can approve it and you can merge it?

DJMcNab:
It's Linebender policy to not merge without an approval. https://github.com/linebender/resvg/pull/843#issuecomment-2466621536 is close enough to approval that I'm happy to formalise it, although I'll just give the actual change a quick scan.

waywardmonkeys (approved):

DJMcNab (approved):
Thanks!

LaurenzV:
Thanks!

DJMcNab:
> where I wanted to achieve that I can override all styles in the SVG via stylesheet injection

Incidentally, does the new swapping behaviour not break this? This injection is basically equivalent to a user agent/user stylesheet, right?

DJMcNab:
Ooh, I see. That makes so much more sense.
I had the condition inverted in my head.

Because my intuition was that the user stylesheet would override the document, since that seems obviously correct. But that's not how the spec is written (the document overwriting the user agent makes sense of course, which I missed)

PR #846, opened 2024-11-12
Update fontdb and rustybuzz dependencies

alerque:
No changes required, all tests pass. I set out to to confirm whether an updated RustyBuzz version would work without breaking anything (prior to releasing 0.20.1 upstream) and all was well. The fontdb bump is to keep two versions of ttf-parser out of the dependency tree.


xStrom (approved):
Seems reasonable, thanks!

PR #850, opened 2024-11-26
Add support for embedded luma JPEG images

LaurenzV:


crates/resvg/src/image.rs
```diff
@@ -87,7 +87,22 @@ mod raster_images {
...
         let mut decoder = zune_jpeg::JpegDecoder::new_with_options(data, options);
-        let img_data = decoder.decode().ok()?;
+        decoder.decode_headers().ok()?;
+        let output_cs = decoder.get_output_colorspace()?;
+
+        let img_data = {
+            let data = decoder.decode().ok()?;
+            match output_cs {
+                ColorSpace::RGBA => data,
+                // For some reason zune jpeg won't coerce luma JPEGs into RGBA, so we do it manually.
```
DJMcNab, line 97: This ideally needs a link to an upstream issue.

crates/resvg/src/image.rs
```diff
@@ -87,7 +87,22 @@ mod raster_images {
...
+                ColorSpace::Luma => data
+                    .into_iter()
+                    .flat_map(|p| [p, p, p, 255])
+                    .collect::<Vec<_>>(),
+                _ => return None,
```
DJMcNab, line 102: Would it make sense to warn here? (Do we even have warning infrastructure available?)

crates/resvg/tests/integration/render.rs
```diff
@@ -1098,7 +1098,9 @@ use crate::render;
 #[test] fn structure_image_embedded_gif() { assert_eq!(render("tests/structure/image/embedded-gif"), 0); }
 #[test] fn structure_image_embedded_jpeg_as_image_jpeg() { assert_eq!(render("tests/structure/image/embedded-jpeg-as-image-jpeg"), 0); }
 #[test] fn structure_image_embedded_jpeg_as_image_jpg() { assert_eq!(render("tests/structure/image/embedded-jpeg-as-image-jpg"), 0); }
+#[test] fn structure_image_embedded_jpeg_luma() { assert_eq!(render("tests/structure/image/embedded-jpeg-luma"), 0); }
 #[test] fn structure_image_embedded_jpeg_without_mime() { assert_eq!(render("tests/structure/image/embedded-jpeg-without-mime"), 0); }
+#[test] fn structure_image_embedded_png_luma() { assert_eq!(render("tests/structure/image/embedded-png-luma"), 0); }
```
DJMcNab, line 1103: Unrelated to this PR, but this is a really nice way of doing tests.

DJMcNab (approved):
Is this workaround as preferred by upstream? I think it would be fine to land this, but it would be better to not have to...

crates/resvg/src/image.rs
LaurenzV: From what I understand it's not really an issue, it's just that zune_jpeg doesn't guarantee that all input types can be converted to RGBA. See: https://docs.rs/zune-core/latest/zune_core/options/struct.DecoderOptions.html#method.jpeg_set_out_colorspace

crates/resvg/src/image.rs
LaurenzV: We do, but we already print a warning here in case JPEG decoding fails (i.e. the function returns `None`): https://github.com/linebender/resvg/blob/a53573b7f49a48ea73e61d92c40f77df96267a4a/crates/resvg/src/image.rs#L65

So I think this should be enough. :)

crates/resvg/tests/tests/structure/image/embedded-jpeg-luma.svg
DJMcNab, line 1: What's the provenance of the embedded PNG data?

crates/resvg/src/image.rs
```diff
@@ -87,7 +87,25 @@ mod raster_images {
...
-        let img_data = decoder.decode().ok()?;
+        decoder.decode_headers().ok()?;
+        let output_cs = decoder.get_output_colorspace()?;
+
+        let img_data = {
+            let data = decoder.decode().ok()?;
+            match output_cs {
+                ColorSpace::RGBA => data,
+                // `set_output_color_space` is not guaranteed to actually always set the output space
+                // to RGBA (see https://docs.rs/zune-core/latest/zune_core/options/struct.DecoderOptions.html#method.jpeg_set_out_colorspace).
```
DJMcNab, line 98:
```suggestion
                // to RGBA (its docs say "we do not guarantee the decoder can convert to all colorspaces").
```
I don't really like to link directly to docs.rs in Rust code, because the same information is already available to you locally as you're reading this code (as `jpeg_set_out_colorspace` is used 10 lines above)

DJMcNab (approved):

crates/resvg/tests/tests/structure/image/embedded-jpeg-luma.svg
LaurenzV: I've taken it from the test suite of my own crate (https://github.com/LaurenzV/krilla/tree/main/assets/images), which in turn is just a spin-off from similar tests in resvg (https://github.com/linebender/resvg/blob/main/crates/resvg/tests/tests/painting/display/none-on-linearGradient.png). I guess at some point I changed the color from yellow to white, don't remember anymore. And then I've basically just re-encoded the resulting image in different file formats using the `image` crate

DJMcNab: Thanks for confirming!

PR #875, opened 2024-12-27
Support floating point viewbox in resvg binary

### LINKED ISSUE #810: viewbox values get rounded to nearest int
Not much to explain here, the issue is quite simple. I hope the fix will be simple too 🤞🏻 I've attached some renders of a test image.

```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1.5 1.5">
	<rect fill="green" width="1.7" height="1.7"/>
	<rect fill="blue" width="1.5" height="1.5"/>
	<rect fill="red" width="1.2" height="1.2"/>
</svg>
```

| Impl | Image | viewbox size | command |
|--------|--------|--------|--------|
| Your Browser | test | - | - |
| rsvg | test_rsvg | 1.5x1.5 | `rsvg-convert test.svg -w 100 -o test_rsvg.png` |
| inkscape | test_inkscape | 1.5x1.5 | `inkscape "/home/riedler/Downloads/test.svg" --export-filename="test_inkscape.png" -w 100` |
| resvg | test_resvg | 2x2 | `resvg test.svg -w 100 test_resvg.png ` |

note that this is less of an issue when rounded up, but quite a big issue when rounded down, since some SVGs are perfectly cut to their contents. I feel that a suitable hotfix would be always rounding up, and cutting the resulting bitmap image to the proper size afterwards.

I didn't test whether aspect ratio is preserved e.g. when `viewBox="0 0 1.5 2"`, but I assume that's also an issue.

jermy:
Avoid rounding sizes for images until creating the pixmap, and always use the ceil for that to avoid truncating images. This fixes #810.

This requires a corresponding change to tiny-skia to add scale_by/scale_to_width/scale_to_height functions to tiny_skia_path::Size to match the implementations in IntSize.

This won't compile as-is, since it depends on linebender/tiny-skia#146 or a similar change.


crates/resvg/src/main.rs
```diff
@@ -526,13 +525,13 @@ fn parse_args() -> Result<Args, String> {
     let mut default_size = usvg::Size::from_wh(100.0, 100.0).unwrap();
     if let (Some(w), Some(h)) = (args.width, args.height) {
         default_size = usvg::Size::from_wh(w as f32, h as f32).unwrap();
-        fit_to = FitTo::Size(w, h);
+        fit_to = FitTo::Size(w as f32, h as f32);
```
DJMcNab, line 528: Would it make sense for these arguments to themselves be floats? I think *probably* not, because they need to be positive and finite. But creating a massive image can still cause issues.

RiedleroD:
don't want to be annoying, but - ping! I'm still waiting on this so I can stop having inkscape as a dependency in my software =w=

LaurenzV:
What’s missing for this? Just a bump in tiny skia?

jermy:
That's my understanding, yes. Once that's available then we can bump dependencies here (and consider a resvg release?).

The only question is if it's useful to have width and height directly parsed from the command line as f32 as @DJMcNab has pondered. If you used `--width 33.5 --height 27.6` you'd still end up with a 34x28 PNG, but maybe with content outside of the bounding box rather than truncating it?

LaurenzV:
Had to update two test cases, but only with very minimal pixel differences. I added a git dependency now because I think it'll still be a while until the next release.

I think we can merge this as is, @DJMcNab wdyt?

RiedleroD:
hey, hate to bump, but what is the course of action from here? is there any way I can help get this through faster? I would really like to have this so I can replace inkscape with resvg as my primary SVG rasterizer…

LaurenzV:
I'm wondering though, is that really such a critical thing to have? To me it seems like a pretty small edge case, no?

RiedleroD:
in my use-case, a lot of the images I use a rasterizer on are small enough (in SVG units) to notice discrepancies at larger scales (in pixels)

the only reason I even noticed is because it really fucked with a bunch of my images. for example:

| resvg | inkscape |
|--------|--------|
| blobcat_thinkage_sceptic | blobcat_thinkage_sceptic_is |

↑ one of the less bad examples. I don't have time to look for a worse one rn, but notice how the bottom and right sides are cut off a bit in the resvg render

luisbg:
This is unblocked now. The reason it stalled (needing `Size::scale_by/scale_to_width/scale_to_height` from tiny-skia) no longer applies: those all shipped in the released tiny-skia-path 0.12.0 that resvg already depends on.

I confirmed the bug is still live on main. Red extends to pixel 60, i.e. 1.2 ÷ 2.0, so the 1.5 viewBox is being rounded up.

I rebased locally to check it works: dropped the git pins, resolved the conflicts, builds clean and all 1,793 tests pass. Afterwards red extends to pixel 80 (1.2 ÷ 1.5), the corner is blue, and there's no green anywhere.

Two things worth noting from the rebase:

1. The two golden PNG updates are no longer needed. main moved on and the current references now match what this branch renders. I kept main's versions and the suite is green.
2. The main.rs conflict isn't mechanical. main gained graceful handling for the `i32::MAX/4` pixmap limit while this PR uses `.unwrap()` with "Unwrap is safe, because size is already valid." That reasoning no longer holds once `ceil()` can push a size over the limit. I resolved it by keeping the float sizing and main's error handling.


@jermy  are you still up for finishing this? Happy to push the rebase if it's easier. @RiedleroD  , sorry it took so long.

RiedleroD:
all that matters is that it's getting done eventually ^^ I'm just happy it's not being dropped

jermy:
> @jermy are you still up for finishing this? Happy to push the rebase if it's easier. @RiedleroD , sorry it took so long.

I'm fairly busy at the moment - I don't have any issue with you taking this and doing that rebase and rework as you see fit. Thank you.

quantum-incongruity:
> This is unblocked now. The reason it stalled (needing `Size::scale_by/scale_to_width/scale_to_height` from tiny-skia) no longer applies: those all shipped in the released tiny-skia-path 0.12.0 that resvg already depends on.

Hey @luisbg are you a new maintainer to resvg?
Do you know when a new version will be released?

luisbg:
I am just helping, not a maintainer. Let me ask when a new version will be released.

luisbg:
No release scheduled, unless there is something important.

PR #876, opened 2024-12-28
usage of `cargo clippy --fix --lib -p usvg` with rust 1.83.0

mothsART:
Pull requests that include:

- clippy fixes

### DIFF
--- a/crates/usvg/src/parser/clippath.rs
+++ b/crates/usvg/src/parser/clippath.rs
@@ -57,9 +57,7 @@ pub(crate) fn convert(
         clip_path = convert(link, state, object_bbox, cache);
 
         // Linked `clipPath` must be valid.
-        if clip_path.is_none() {
-            return None;
-        }
+        clip_path.as_ref()?;
     }
 
     let mut id = NonEmptyString::new(node.element_id().to_string())?;
--- a/crates/usvg/src/parser/paint_server.rs
+++ b/crates/usvg/src/parser/paint_server.rs
@@ -331,7 +331,7 @@ fn convert_stops(grad: SvgNode) -> Vec<Stop> {
     if stops.len() >= 3 {
         let mut i = 0;
         while i < stops.len() - 2 {
-            let offset1 = stops[i + 0].offset.get();
+            let offset1 = stops[i].offset.get();
             let offset2 = stops[i + 1].offset.get();
             let offset3 = stops[i + 2].offset.get();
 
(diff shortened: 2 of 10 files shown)


crates/usvg/src/parser/clippath.rs
```diff
@@ -57,9 +57,7 @@ pub(crate) fn convert(
         clip_path = convert(link, state, object_bbox, cache);
 
         // Linked `clipPath` must be valid.
-        if clip_path.is_none() {
-            return None;
-        }
+        clip_path.as_ref()?;
```
DJMcNab, line 60: I don't think this is better. If you do want to change this, I'd say:
```rust
// Bail out if the linked clip path is invalid
let clip_path_attr = convert(...)?;
clip_path = Some(clip_path_attr);
```
But I think the original code is *fine*...

crates/usvg/src/parser/paint_server.rs
```diff
@@ -331,7 +331,7 @@ fn convert_stops(grad: SvgNode) -> Vec<Stop> {
     if stops.len() >= 3 {
         let mut i = 0;
         while i < stops.len() - 2 {
-            let offset1 = stops[i + 0].offset.get();
+            let offset1 = stops[i].offset.get();
```
DJMcNab, line 334: Again... I can see the rationale, but I think the previous version was probably better.

crates/usvg/src/text/layout.rs
```diff
@@ -1418,7 +1418,7 @@ fn shape_text_with_font(
 
                 glyphs.push(Glyph {
                     byte_idx: ByteIndex::new(idx),
-                    cluster_len: end.checked_sub(start).unwrap_or(0), // TODO: can fail?
+                    cluster_len: end.saturating_sub(start), // TODO: can fail?
```
DJMcNab, line 1421: The semantics are the same, but the meaning this suggests to the reader is different

DJMcNab (commented):
A lot of these changes are at best marginal

LaurenzV:
> A lot of these changes are at best marginal

My understanding was that all of these are changes automatically applied by clippy, no? 😄 In which case we probably would have to adapt which lints we want to have.

crates/usvg/src/text/layout.rs
tomcur: I don't know the history of this line, but seems like a runtime panic was too risky. Maybe a `debug_assert` above the push asserting `start <= end` would be nice here?

tomcur: Exactly, that's my reading as well.

DJMcNab:
I've just opened my first PR to resvg, and seen that "pull requests that include: clippy fixes" is part of the section that ends with "will not be accepted.".

The author has also not been responsive, and so nothing's likely to change here.

PR #878, opened 2025-01-20
Update to svgtypes 0.15.3

tomcur:
This update of svgtypes included some changes in rounding behavior, so some tests are updated.

My spot checks only found pixels that were off by one rounding unit.


waywardmonkeys:
Thanks, that's what I expected. The new rounding match `color`, but also match the various web platform tests that I've checked out.

waywardmonkeys (approved):

tomcur:
> but also match the various web platform tests that I've checked out.

That's good!

PR #880, opened 2025-02-04
Update changelog for 0.45

laurmaedje:
I went through all changes since 0.44.0 and added the missing ones to the changelog. I tried to keep the changelog style as close to the typical one.


CHANGELOG.md
```diff
@@ -7,16 +7,37 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 This changelog also contains important changes in dependencies.
 
 ## [Unreleased]
+This is the first release under the stewardship of [Linebender][], who is now responsible for maintenance of this crate.
```
DJMcNab, line 10: Should this link have a corresponding definition?

Same for the other links in this change

laurmaedje: Nevermind apparently they are all at the bottom. Then I don't get the second pair of brackets. It seems unnecessary.

LaurenzV:
I mean, seems to look fine to me now, so I'll merge, thanks!

PR #892, opened 2025-03-02
Root without xmlns

### LINKED ISSUE #847: `usvg` fails to parse SVGs if `xmlns` attribute is missing
`usvg` will currently fail to parse otherwise valid SVGs is the `xmlns` attribute is missing. This is [the correct behaviour for standalone SVG files, but not correct for "inline SVGs" embedded within HTML documents](https://stackoverflow.com/a/18468348/140293).

As such, I would like to propose that an option is added to [usvg::Options](https://docs.rs/usvg/latest/usvg/struct.Options.html) that determines whether the namespace should check is enabled.

I would be able to create a PR for this.

JosefKuchar:
Closes #847 

This PR implements support for SVG files without xmlns tag.
Existing tests pass.

### DIFF
--- a/crates/usvg/src/parser/svgtree/parse.rs
+++ b/crates/usvg/src/parser/svgtree/parse.rs
@@ -136,9 +136,8 @@ pub(crate) fn parse_tag_name(node: roxmltree::Node) -> Option<EId> {
         return None;
     }
 
-    match node.tag_name().namespace() {
-        None | Some(SVG_NS) => {}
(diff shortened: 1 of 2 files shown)


LaurenzV (approved):
Seems fine to me in principal to not be too strict here, but would like to see approval from one other person before merging.

RazrFalcon:
Should be fine I guess. Hard to tell.

Issue #902, opened 2025-04-14
Mac builds failure investigation


thekingofcity:
After some investigation into the Mac build/release failure, here are the findings and possible solutions:

#### 🔍 Findings

1. **`Install QT: ERROR: Failed building wheel for pyzstd` in `v0.45.0` official pipeline**  
   This error occurs during the `jurplel/install-qt-action` step.  
   ✅ **Fix:** Upgrade to `jurplel/install-qt-action@v4`.

2. Then **`Project ERROR: failed to parse default search paths from compiler output`on next step `Build viewsvg`**
   This error is caused by an incompatibility with **default Xcode 15** on [`macos-13` image](https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md#xcode).  
   - See QT bug: [QTBUG-117225](https://bugreports.qt.io/browse/QTBUG-117225)
   - This issue is only fixed in **Qt 5.15.15** or higher, which is a **paid version**.
   - `aqtinstall` does not support downloading paid versions. See: miurahr/aqtinstall#636

#### ✅ Options

1. **Ignore `viewsvg` builds on Mac**  
   - `resvg` continues to work fine.  
   - Easiest path forward.

2. **Upgrade to Qt 6**  
   - This issue is fixed in Qt 6.  
   - ⚠️ Significant migration effort required.

3. **Force Xcode 14 on `macos-13` image**  
   - Forked builds using this setup are working successfully. See [my fork builds](https://github.com/thekingofcity/resvg/actions/runs/14436506271). And [code changes](https://github.com/thekingofcity/resvg/blob/21fb57e5bd56bc040a740d80c92f07d4decbe84a/.github/workflows/main.yml#L16).
   - ✅ Viable short-term fix with minimal changes.

Let me know which path we’d like to proceed with. I'm happy to open a pull request for option 1 or 3 if that is preferred :)

DJMcNab:
Thanks for digging into this.

My impulse here is to remove the QT integration from this repository entirely (if anyone needs it, they can vendor it themselves). This would be for similar reasons as #897, that we don't have the will to maintain it, and it's causing us issues.
(To be entirely clear, I am **NOT** proposing to remove the C api, which the QT integration itself uses).

If at a later point, someone who is motivated to restore viewsvg chooses to do so, we'd happily engage with that. Ideally, we'd have a test integration of the C API in this repository as well (but removing `viewsvg` doesn't meaningfully change that state, because we don't test it in CI except on a release).

[...]

Issue #891, opened 2025-02-27
Tagged Release action needs updating


waywardmonkeys:
The tagged release action (https://github.com/linebender/resvg/blob/main/.github/workflows/tagged-release.yml) failed to run correctly when doing the 0.45 release and needs several fixes / updates.

(I'll be doing this, this is just to track it so that I don't forget.)

flying-sheep:
e.g. there’s no `v` prefix (the tag is `0.45.0` instead of `v0.45.0`)

thekingofcity:
Mac builds might be fixed in #902 :)

LaurenzV:
@RazrFalcon How do you manage that the titles of tagged releases are in the form of "v0.x.y" instead of "refs/tags/v0.x.y"? Did you always manually update them?

flying-sheep:
You should probably just replace `github.ref` with `github.ref_name` in the `name` field

https://github.com/linebender/resvg/blob/9c4431d10bbf970ede3bd404392e7944f7544192/.github/workflows/tagged-release.yml#L22-L23

→

```yaml
tag_name: ${{ github.ref }} 
name: ${{ github.ref_name }} 
```

thekingofcity:
Just want to know as we fix the pipeline, will we see the mac build real soon?

Would be great to have v0.45.1 with mac build :)

waywardmonkeys:
Sounds good. Do you want to do it or is there a part that I need to do?

LaurenzV:
I can do it on GitHub, only the crates.io release would need to be done by you, I think.

waywardmonkeys:
Sent you invites on crates.io to solve that!

PR #904, opened 2025-04-14
Fix the tagged release action

LaurenzV:
This seems to work fine on my fork now: https://github.com/LaurenzV/resvg/releases/tag/v0.52.0


.github/workflows/tagged-release.yml
```diff
@@ -20,9 +20,8 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
           tag_name: ${{ github.ref }}
-          name: ${{ github.ref }}
+          name: ${{ github.ref_name }}
```
DJMcNab, line 23: Do we know why this previously worked?

DJMcNab (approved):
Thanks!

Does this fix #902? It doesn't seem to be marked that it does.
Does it also fix https://github.com/linebender/resvg/issues/891?

PR #906, opened 2025-04-15
Only allow releasing `v*` tags

flying-sheep:
I think this was missed in #891

Please also retag https://github.com/linebender/resvg/tree/0.45.0 as `v0.45.0`!


LaurenzV:
Let me see if I can retag.

PR #908, opened 2025-04-16
feat: added support for variable weight axis

pamburus:
This pull request adds minimal support for fonts with a variable “wght” (weight) axis.

When a font file includes a weight axis, it can be registered multiple times with different face entries — each specifying its own weight.

With this change, outlining and rendering will honor the configured weight instead of ignoring it.


RazrFalcon:
I doubt this how it should be implement. `resvg` doesn't support variable fonts at all right now.

PR #909, opened 2025-04-16
Update Ubuntu version in tagged release action

LaurenzV:
I presume that after this being merged, I need to delete the currently-existing tag and retag to the newest main?


DJMcNab:
Yeah, I think that would work. Thanks

DJMcNab (approved):

PR #913, opened 2025-04-19
Fix `clippy::semicolon_if_nothing_returned` lints

waywardmonkeys:


waywardmonkeys:
I meant to set this to auto-merge, not merge it. Whoops.

DJMcNab (commented):
Yes, we don't have branch protection set up here.

We can count this as an approval; luckily this is about the only PR where merging it without review is definitely fine; it only adds semicolons and everything compiles!

PR #917, opened 2025-05-05
Decouple the build of viewsvg from that of resvg

Kidev:
### Decouples the build of `viewsvg` from that of `resvg`

- Builds of `viewsvg` are now described in their own CI file (`viewsvg-release.yml`), **are manual**, and target a specific release tag. This workflow builds `viewsvg` for Windows, Linux and macOS, and adds their binaries inside the release that was targeted. Vast majority of changes have been moved to that file. See [my post below](https://github.com/linebender/resvg/pull/917#issuecomment-2859811558) for details and more screenshots  
image

- The changes outside of that new CI file are the following (detailed reasons [here](https://github.com/linebender/resvg/pull/917#issuecomment-2854175927)):
  - Update action `actions/checkout` to `v4`
  - Update runners to `macos-15`, `windows-2022` and `ubuntu-24.04`
  - Make use of C++20 inside the file `tools/viewsvg/viewsvg.pro` file (Qt6 requires >= C++17)
  - Remove `QApplication::setAttribute(Qt::AA_EnableHighDpiScaling)` from `tools/viewsvg/main.cpp` (default behavior in Qt6)
  - Make variable capture explicit inside lambdas in `tools/viewsvg/svgview.cpp`

### DIFF
--- a/.github/workflows/tagged-release.yml
+++ b/.github/workflows/tagged-release.yml
@@ -22,6 +22,7 @@ jobs:
           tag_name: ${{ github.ref }}
           name: ${{ github.ref_name }}
           body: |
+            - `viewsvg` is a simple application that showcases *resvg* capabilities
             - `resvg-0.*.0.tar.xz` is a sources archive with vendored Rust dependencies
             - `resvg-explorer-extension.exe` is an SVG thumbnailer for Windows Explorer
 
@@ -88,7 +89,7 @@ jobs:
 
   release-windows:
     name: Release Windows
-    runs-on: windows-2019
+    runs-on: windows-2022
     needs: ["create-release"]
     steps:
       - name: Checkout
@@ -126,6 +127,11 @@ jobs:
         run: |
           "%programfiles(x86)%\Inno Setup 6\iscc.exe" "installer.iss"
 
+      - name: Install Qt
+        uses: jurplel/install-qt-action@v4.2.1
+        with:
+          version: '6.8.3'
+
       # Unlike other binaries, viewsvg isn't built with crt-static
       - name: Build C API
         working-directory: crates/c-api
@@ -134,12 +140,24 @@ jobs:
       - name: Prepare Developer Command Prompt for MSVC
         uses: ilammy/msvc-dev-cmd@v1
 
+      - name: Build viewsvg
+        working-directory: tools/viewsvg
+        run: |
+          qmake6
+          nmake
+          mkdir viewsvg-bin
+          cp release/viewsvg.exe viewsvg-bin/viewsvg.exe
+          windeployqt --no-translations viewsvg-bin/viewsvg.exe
+          cd viewsvg-bin
+          7z a -tzip -mx9 viewsvg-win64.zip *
+
       - name: Collect
         run: |
           mkdir bin
           cp target/release/resvg-win64.zip bin/
           cp target/release/usvg-win64.zip bin/
           cp tools/explorer-thumbnailer/install/resvg-explorer-extension.exe bin/
+          cp tools/viewsvg/viewsvg-bin/viewsvg-win64.zip bin/
 
       - name: Upload binaries
         uses: alexellis/upload-assets@0.2.2
--- a/.github/workflows/tagged-release.yml
+++ b/.github/workflows/tagged-release.yml
@@ -53,7 +53,7 @@ jobs:
           strip -s usvg
           tar czf resvg-linux-x86_64.tar.gz resvg
           tar czf usvg-linux-x86_64.tar.gz usvg
-          mkdir -p ../../bin
+          mkdir ../../bin
           cp resvg-linux-x86_64.tar.gz ../../bin/
           cp usvg-linux-x86_64.tar.gz ../../bin/
 
--- a/.github/workflows/tagged-release.yml
+++ b/.github/workflows/tagged-release.yml
@@ -37,7 +37,7 @@ jobs:
     needs: ["create-release"]
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
 
       - name: Build resvg
         run: cargo build --release
@@ -92,7 +92,7 @@ jobs:
     needs: ["create-release"]
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
 
       # Toolchain is stable-x86_64-pc-windows-msvc by default. No need to change it.
 
@@ -150,11 +150,11 @@ jobs:
 
   release-macos:
     name: Release macOS
-    runs-on: macos-13
+    runs-on: macos-15
     needs: ["create-release"]
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v4
 
       # Some weird CI glitch. Make sure we have the latest Rust.
       - name: Install latest stable toolchain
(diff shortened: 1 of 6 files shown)


DJMcNab:
Agreed. I'd strongly push back against having this as part of the release process. QT has shown itself to be fragile to build conditions, and I don't want to go through this same process again in a year.

I can see arguments for setting up an "advisory" job, which checks that this still builds. But we still don't really have the bandwidth for that.

But I agree - @RazrFalcon can make a decision here.

RazrFalcon:
I appreciate the patch, but you have done too much. We do not need AppImage. We do not need C++20. We do not need an app icon. Etc.
We just want a quick fix for a CI. And a fix that is easy to maintain.

Since I'm no longer maintaining this project, it's for others to decide. If others find it complicated, then we're closing it.

The whole point behind `viewsvg` is to test Qt C API wrapper and have a GUI demo for resvg rendering. `viewsvg` basically an ad.
But since `resvg` is already popular enough, we can probably deprecate `viewsvg` completely.

Kidev:
That’s exactly why I prefer to comment or offer guidance on CI in repos I’m less familiar with, modifying the CI code often feels like getting involved in a family matter :grin: 

However there's really not that much I added, here's my rationale


--- 
> Make use of C++20 and `Qt6` to build the binary `viewsvg`

➣ For Qt6, this is the recommended version. The minimum required is C++17  

> Add back the generation of the `viewsvg` binary for MacOS and Windows tagged releases

➣ I used the history of the file `tagged-release.yml` to add back only what you removed because it did not work with Qt5

> Add the generation of the `viewsvg` binary (AppImage) for Linux tagged releases

➣ Maybe the most controversial I guess, it looks a bit verbose but the process is very similar to MacOS or Windows, except Qt does not provide a linuxdeploy binary, so I used [this popular, nice and maintained tool](https://github.com/linuxdeploy/linuxdeploy). My OCD wanted to fix that 2/3

> Remove the now useless and deprecated `QApplication::setAttribute(Qt::AA_EnableHighDpiScaling)`

➣ This simply removes a function that has no effect in Qt6 as it is now the default behavior

> Make variable capture explicit inside lambdas (implicit capture is deprecated since C++20)

➣ It always was bad practice, but it is as of C++20 deprecated to implicitly capture `this`, as it is a very common source of annoying lifetime issues. Here in reality, the issue is that this is not at all the right way to do this using Qt. The correct way would be to leverage Qt signals, so that the lifetime of `this` nor a thread error cannot be an issue ever, for free
```diff
-    // Run method in the m_worker thread scope.
-    QTimer::singleShot(1, m_worker, [this, s](){
-        m_worker->render(s);
-    });
+    // Emit signal to request rendering in worker thread
+    emit renderRequested(s);
```

> Update actions:

➣ The old versions you were using are unsupported, and for example, `install-qt-action` prior to `v4.2.1` is unlikely to work anymore because Github entirely removed `action/cache@v3` and below

> Update runners:

➣ Used sensible versions that are the long-term good idea: 
- MacOS for Xcode15
- Windows because MSVC2019 is [a month away from deprecation](https://github.com/actions/runner-images/issues/12045)       
- Linux because using `latest` is the best way to create trouble for your future selves when Github changes its meaning overnight and breaks your CI

What are your concerns exactly? I can change anything if you want I don't mind, but I really think that except for the AppImage I did the minimum here

DJMcNab:
> Add back the generation of the viewsvg binary for MacOS and Windows tagged releases
↳ I used the history of the file tagged-release.yml to add back only what you removed because it did not work with Qt5. I used explicitly qmake6 instead of just qmake which would have worked as well, to make it very clear that this is now dealing with Qt6

As we have stated both in this thread and elsewhere, we didn't remove this because it didn't work with qt5, we removed it because we don't have the expertise to meaningfully maintain C++ code or a build which has proven itself to be fragile. And so having it in the repository is fine, but blocking releases working properly on it not fine.

My suggested way forward would be to revert the changes in ci.yml. That is, only have the changes inside the `tools/viewsvg` folder. Bumping the macOS, linux and windows runners (plus actions/checkout) is worthwhile, although should probably be a different PR. I don't expect you to make that PR.

Kidev:
> As we have stated both in this thread and elsewhere, we didn't remove this because it didn't work with qt5, we removed it because we don't have the expertise to meaningfully maintain C++ code or a build which has proven itself to be fragile. And so having it in the repository is fine, but blocking releases working properly on it not fine.
Oh..! well I missed that. Honestly I discovered this repo because it referenced a **pinned** closed issue on `aqtinstall`, that was misleading people because pinned issues always appear on top even when closed (I wrote the whole official qt installer feature for `aqtinstall` and `install-qt-action` so I was sad a Github feature made people miss it). When I saw the issue post, I mainly saw this:

> Upgrade to Qt 6
> - This issue is fixed in Qt 6.
> - ⚠️ Significant migration effort required.

I use Qt daily and the repo looked very simple and was using features untouched by the migration, so I figure I could help since for me I knew it would be easy (having experience with doing so on large 3D projects). Did not read all that much more, so my knowledge of your project is more down-top than the opposite

 
> My suggested way forward would be to revert the changes in ci.yml. That is, only have the changes inside the `tools/viewsvg` folder. Bumping the macOS, linux and windows runners (plus actions/checkout) is worthwhile, although should probably be a different PR. I don't expect you to make that PR.
I can move all that is `viewsvg` related in its own CI file, that is triggered only manually (`workflow_dispatch`). This action could get the latest release of `resvg`, build `viewsvg` and append the binaries to the existing release. I could do that fairly easily

> I want to be clear about what's going wrong here - I really do appreciate that you're willing to update this. But I'm not going to ask you to keep this maintained, and therefore we need to minimise how much maintenance is needed.
Don't worry I 100% get it, and this is the right decision for a project. I am not aware of the issues you had in the past, and I don't want to cause problems, in fact if I contribute to projects it's for the exact opposite reason :smile:  

I would just add that `viewsvg` is not, with my experience, a project that would take very long to maintain at all, and I'm not against maintaining a tool like that on the side really, beleive it or not I do that for fun :shipit: I could even write a quick fork that could be integrated with git-submodules to decouple it even more from the Rust part of the source, which is where all the added value of your project is

thekingofcity:
Hi, I would really appreciate your work as a resvg mac user. Sorry if the following misleads you. I just did some investigation on the build failure on mac and hope to help fix it. I'm not familiar with Qt either, nor am I a maintainer in the repo, so the following may not reflect the view from the collaborators.

> 2. **Upgrade to Qt 6**  
>    - This issue is fixed in Qt 6.  
>    - ⚠️ Significant migration effort required.

Kidev:
No worries! I literally crashed into a closed issue and threw a patch, I can't then just stand there and complain that I misread a thread that I barely read :laughing: 

Now that I think I understand your point of view more, I think that there is an elegant solution to your 'issue': what about generating binaries for each OS with dynamic linking to `resvg`? That way, you never have to recompile the code that does not matter to you, but you can keep using it as advertisement. For this the action file that is only manually started and that creates the binaries is a good way of doing this I think

Kidev:
So, here's how it works: your tagged releases now no longer build `viewsvg`, ever.  
To build it, a maintainer must go into `Actions` > `ViewSVG Release` > `Run workflow`. Here you can write a tag against which you'd like to build `viewsvg` with and for. You can leave it empty and it'll take the last release. It'll then build `viewsvg` for each OS, add the binaries to the existing targeted release, and if at least one OS built it well, it'll add this line to the top of the release (because easier to do). Here's screenshots:

**To run it using and on the latest release**


**To run it on `v0.46.0`**




**The result! Pay attention to the updated time**

LaurenzV:
Sorry again that you put this much work into this without it being merged, but I still think we are unlikely to need this, it adds a lot of maintenance burden for little benefit. Thanks for your contribution regardless!

PR #918, opened 2025-05-06
skip text layout in usvg with preserve-text arg

shaneeb:
Presently, when `--preserve-text` argument is passed to usvg cli, even though `<text>` nodes are passed as is in the output, usvg still does text layout / rendering as part of parsing. If all the fonts used in text elements are present, this works as expected. However, if any font is missing the associated text node is removed from the output.

We have a use case where we want to run usvg on a service that optimizes SVG files before being rendered with resvg in an environment where the fonts are present. As part of this optimization, we want the text nodes to be preserved. The current parsing logic in usvg prevents this.

This PR adds a fix that skips the text layout logic on text nodes if the `preserve-text` argument is passed to usvg cli. I can't think of any downsides to this approach but would love feedback on whether there is a better to way to accomplish this.

Thanks!

### DIFF
--- a/crates/resvg/src/main.rs
+++ b/crates/resvg/src/main.rs
@@ -578,6 +578,7 @@ fn parse_args() -> Result<Args, String> {
         font_resolver: usvg::FontResolver::default(),
         fontdb: Arc::new(fontdb::Database::new()),
         style_sheet,
+        preserve_text: false,
     };
 
     Ok(Args {
--- a/crates/usvg/src/main.rs
+++ b/crates/usvg/src/main.rs
@@ -432,6 +432,7 @@ fn process(args: Args) -> Result<(), String> {
         font_resolver: usvg::FontResolver::default(),
         fontdb: Arc::new(fontdb),
         style_sheet,
+        preserve_text: args.preserve_text,
     };
 
     let input_svg = match in_svg {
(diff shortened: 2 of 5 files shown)


LaurenzV:
Unfortunately, there is a reason why we _have_ to do text layouting for text nodes: There are certain attributes of an SVG that depend on computing the bounding box of a text node (any paint that that uses `objectBoundingBox` units), and if we don't convert the text, those units would be wrong. 

So while I definitely understand the motivation for that flag, I'm not entirely sure whether we can add this to the main repo, given that the main goal of the crate is still correct rendering of SVG files. I _think_ if you really need this for your use case, it might be the best to maintain a fork with this small patch applied. But maybe @RazrFalcon  has other thoughts.

shaneeb:
Thanks for the feedback! Super helpful!

For our use case, we will probably just maintain a fork with this patch. The SVGs we process are typically generated from popular vector software (think Adobe, Figma, etc) and tend to use "simple" units in SVGs (px, pts, etc) so we think we should be ok?

Apart from objectBoundingBox units what else can potentially break rendering, top of your head? Just trying to figure out a possible direction for further exploration...

LaurenzV:
I believe the object bounding box thing is the main issue, though I'm not 100% certain.

PR #922, opened 2025-06-11
Port resvg to `harfrust` and `fontations`

nicoburns:
## Changes made

- Replaces `rustybuzz` with `harfrust`
- Replaces `ttf_parser` with `skrifa`

## Notes

- This is failing 23 tests. These probably need to be manually checked. I'm waiting on https://github.com/linebender/resvg/pull/954 to investigate these.
- This requires an MSRV bump from `1.65` to `1.85`
- This will represent a binary size bump for users of `resvg`.

## Production profile

The analysis below uses the following "production" profile:

```toml
[profile.production]
inherits = "release"
opt-level = 3
debug = false
lto = true
codegen-units = 1
strip = true
incremental = false
```

## Binary size analysis:

According to `cargo-bloat` (not the most accurate):

  - `harfrust` and `rustybuzz` are both ~250kb
  - `ttf_parser` and `read-fonts` are both ~100kb
  - However, `ttf_parser` includes the metrics outlining functionality of `skrifa`. And skrifa is another ~250kb.

Compiling the CLI (which is a much more useful test) with the production profile above (`cargo build --profile production --bin resvg`) I get:

- `2.4mb` for `main`
- `3.0mb` for this PR

**So this PR causes a binary size increase of around `600kb`**

## Compile time analysis

Compile times on an M1 Pro (8+2 cores) are shown below (`cargo build`):

| Profile | `main` | This PR |
| --- | ---| --- |
| `debug` | 4.68s | 9.09s |
| `release` | 7.64s | 14.84s |
| `production` | 24.59s | 34.00s |

**So this PR roughly doubles compile times at regular opt settings**
(with the effect being less pronounced with LTO and other production optimisation settings enabled)


LaurenzV:
Yeah I think we first have to figure out how to deal with fontdb before doing anything with resvg. I don’t think RazrFalcon would want to switch it, and I’m not sure that fontique supports everything we need?

RazrFalcon:
It's an interesting experiment, but I don't see a point in it. Is there a reason in switching to `harfrust`? 
Last time I have checked it was still a beta, while `rustybuzz` (with all its faults) is a pretty robust solution.

PS: even if `harfrust` is faster, it doesn't matter either, because shaping isn't a bottleneck, lack of glyphs caching is.

RazrFalcon:
Oh wait, it depends on proc-macros/syn?! Burn it with fire! No proc-macros in `resvg`.

.github/workflows/main.yml
```diff
@@ -65,7 +65,7 @@ jobs:
       - name: Install toolchain
         uses: dtolnay/rust-toolchain@stable
         with:
-          toolchain: 1.67.1
+          toolchain: 1.79.0
```
xStrom, line 68: Shouldn't this (and the other `rust-version` modifications) be 1.75.0 not 79?

xStrom:
> Last time I have checked it was still a beta

Not sure about 'beta' type labels, but the first release ([0.1.0](https://github.com/harfbuzz/harfrust/releases/tag/0.1.0)) of HarfRust happened on June 10, marking a significant milestone.

> It's an interesting experiment, but I don't see a point in it. Is there a reason in switching to `harfrust`?

For an immediate effect, it seems to be far more compliant when tested against the HarfBuzz test suite.

To quote Chad:
> We’re still mulling over some public API changes and need to do some serious performance work but this is fully backed by fontations, up to date with HarfBuzz 11.2.1, and passes a significantly larger portion of the HB test suite than RustyBuzz: latest CI shows 5911 tests passing for HarfRust vs 2267 for RustyBuzz.

And the [HarfRust introduction post](https://docs.google.com/document/d/1aH_waagdEM5UhslQxCeFEb82ECBhPlZjy5_MwLNLBYo/preview) by Behdad contains the following claim:

> At the time of this writing, HarfRust passes most of the HarfBuzz shaping test suites, failing just 23 tests out of more than 6,000.

---

Moving forward, HarfRust will receive active development and continuous sync with HarfBuzz, while RustyBuzz is likely to only see critical fixes.

RazrFalcon:
>latest CI shows 5911 tests passing for HarfRust vs 2267 for RustyBuzz.

Apples and oranges, but sure.

Either way, proc-macros is the current major roadblock.

mattfbacon:
No it wouldn't be at all, since the bytemuck derive macros generate unsafe code so they would need to allow unsafe code back into the codebase.

But what is your plan here? Earlier you said wait, but for what?

behdad:
> The geometric mean of the last result is ~1.16, meaning Harfrust is now only 16% slower than Harfbuzz.

It's even better than that. See the comment:
https://github.com/harfbuzz/harfrust/discussions/257#discussioncomment-14363176

In short, there's some hb-harfrust overhead included in the spreadsheet number.

nicoburns:
Diff for the colv1 test https://www.diffchecker.com/V8j95jlx with some manual fixups (removing groups - although it could also be those causing the differences?). I'm seeing some negations (which possibly cancel each other out?), and also that skrifa is applying some opacities to groups whereas ttf-parser applies them directly to the paths.

LaurenzV:
It's possible the difference stems from the fact that the opacities are applied differently. When applying directly to a path, it will simply draw a path with that, while when applying to a group resvg will do layer isolation and apply the opacity that way instead, so you could try if this can be fixed by applying it directly.

nicoburns:
Ok, I've managed to fix the CBDT offset issue (I admit more by trial-and-error than by principle), and I've also managed to fix to fix the COLRv0 missing glyph parts (this was a silly bug where string was being thrown away rather than persisted).

That leaves:

### The COLRv1 and Zalgo tests. Diff output from the test runner for these tests is as below:

tests_text_color-font_colrv1

tests_text_text_zalgo

For the COLRv1 tests, see also the diff in the generated "lowered" SVG output (https://www.diffchecker.com/V8j95jlx)

@dfrg Do you have an opinion on whether these represent a genuine bug, a fixing of a bug, or a valid difference in interpretation?

### The negations

This seems to be negated output for each y-coord in path, which is then cancelled out by a negated y coordinate in a transform matrix. This is probably fine?

### Missing path segments

Some of the text-related `usvg` "resave" tests (parse SVG then rereserialize back to SVG) are failing. Some sample diffs:

- text_simple https://www.diffchecker.com/e2SQWqnm/
- complex_text https://www.diffchecker.com/tQk27XmG/
- preserve_text_simple_case https://www.diffchecker.com/Ip5F1whl/
- text_with_generated_gradients https://www.diffchecker.com/gYS1sYk2
- clip_path_with_text https://www.diffchecker.com/tDIazl6O/

The issue seems to be that the last segment of some (but not all) paths is being dropped in the Skrifa implementation. Although this seems to make no visual difference to the rendered SVG (see github diff), so perhaps that's also fine?

---

RazrFalcon:
>As some months have passed another benefit has appeared - performance. 

Good. It would not affect `resvg` in any way, because shaping was never a bottleneck, but still a progress.
And `rustybuzz` is dead anyways.

Enivex:
I don't know what the status of this is, but harfrust 0.6.0 was released today

https://github.com/harfbuzz/harfrust/releases/tag/0.6.0

nicoburns:
>  harfrust 0.6.0 was released today

Indeed, although it's just a dep bump release.

> I don't know what the status of this is

The remaining blocker is switching resvg to use `fontique` instead of `fontdb` (see: https://xi.zulipchat.com/#narrow/channel/465085-resvg/topic/Moving.20resvg.20to.20fontations.2Fharfrust/with/572344735). I have some half-baked (not even compiling yet) changes for this locally, but it's unfortunately not top of my priority queue right now.

Enivex:
> Indeed, although it's just a dep bump release.

The numbering is a bit strange. There were more changes in 0.5.1 and 0.5.2.

Edit: I just realized that you were the one who bumped it 😂

laurmaedje:
> The numbering is a bit strange. There were more changes in 0.5.1 and 0.5.2.

Just normal SemVer I would assume. It looks like `read_fonts` is in the public API of harfrust, so bumping its minor version is a breaking change for harfrust users.

behdad:
Can we make a move here?

nicoburns:
@behdad The latest discussion on this is https://xi.zulipchat.com/#narrow/channel/465085-resvg/topic/Moving.20resvg.20to.20fontations.2Fharfrust/with/572344735.

TLDR is that we need to find a solution for the "font database". Either switching to Fontique (preferred but more work), or porting `fontdb` to fontations and publishing a fork as we don't control that project (less work, but not ideal to have two "font database" crates on the go).

I have a branch on top of this one (https://github.com/nicoburns/resvg/commits/fontique) that partially moves resvg to fontique. However, that still needs a bunch of design work. Porting `fontdb` to fontations and publishing it as a fork would definitely be easier/quicker if there is time urgency (I also have a branch for that which only needs minor tweaks - https://github.com/nicoburns/fontdb/tree/fontations)

LaurenzV:
@RazrFalcon are you open to moving fontdb towards read-fonts?

nicoburns:
Just looked at this again, and the "blockers" (things that need to be worked out) for porting to Fontique are:

- `fontdb` has unique IDs (64bit `Copy` type) for every font in the database, and resvg/usvg currently assume that the font database works like this and that one can store an ID (returned from the lookup/query/fallback phase) and then cheaply lookup the actual raw font data using that ID. Fontique doesn't not currently assign fonts a unique ID, and the `Collection` does not even always have the actual font data (that can sometimes live in the `SourceCache`).

- The custom "font resolver" functions that allow users of resvg to customise font querying/fallback will need Fontique versions.

yisibl:
Perhaps we could start by migrating to `harfrust`, and handle the rest separately.

@behdad How does `harfrust` compare to `rustybuzz` in terms of performance currently?

nicoburns:
> Perhaps we could start by migrating to `harfrust`, and handle the rest separately.

This is also an option. Just this PR as-is would eliminate `rustybuzz` from the dependency tree. We'd still have `ttf-parser`, but that's relatively small.

behdad:
> @behdad How does `harfrust` compare to `rustybuzz` in terms of performance currently?

In our measurement, it's multiple times faster on a variety of benchmarks. Benchmarks from last year:

https://docs.google.com/spreadsheets/d/1lyPPZHXIF8gE0Tpx7_IscwhwaZa4KOpdt7vnV0jQT9o/preview

We have improved it more since.

nicoburns:
I have:
- Updated this to Skrifa 0.42 and HarfRust 0.8
- Rebased on latest main
- Finished off the [fontdb fontations port](https://github.com/RazrFalcon/fontdb/compare/master...nicoburns:fontdb:fontations?expand=1), and linked this PR to it using a git dep.

RazrFalcon:
I think `fontdb` should simply use minimal `ttf-parser` fork. We need to parse just a couple of fields. `ttf-parser` become an overkill a long time ago. And `read-fonts` looks like a way too heavy dependency for that too.

As for repo, I can transfer it to someone. No need for a fork.

Otherwise I'm fine with switching to `harfrust`. Not like there is a choice.

nicoburns:
Ok, we may need to move on this more quickly now that `rustybuzz` and `ttf-parser` have been marked as unmaintained. People are getting RUSTSEC advisories, and are starting to raise issues against dependent crates. I think we should try to land this without the Fontique migration (which will require a bit of thought around architecture).

@RazrFalcon Would you be willing to transfer me the the `fontdb` repository, and add me as an owner on crates.io. My intention is to port it to `read-fonts`.

@dfrg Do you think we could get the `skrifa::string` module moved to `read-fonts`? That would allow `fontdb` to depend only on `read-fonts` without pulling in `skrifa`.

Shnatsel:
Are there no people with publish access to those crates who could push security updates if the need arises? That's all the "unmaintained" status really was for, back when I was a RustSec maintainer. If there are still people who can publish updates, the unmaintained advisories should be withdrawn.

behdad:
> Are there no people with publish access to those crates who could push security updates if the need arises? That's all the "unmaintained" status really was for, back when I was a RustSec maintainer. If there are still people who can publish updates, the unmaintained advisories should be withdrawn.

But it *is* unmaintained. The HarfBuzz developers don't have the capacity to update it, and the original author doesn't either.

dfrg:
> @dfrg Do you think we could get the `skrifa::string` module moved to `read-fonts`? That would allow `fontdb` to depend only on `read-fonts` without pulling in `skrifa`.

Do you need something specific from skrifa? I was considering moving the BCP-47 language map and the english_or_first helper to read-fonts. Would that be sufficient?

LaurenzV:
Seems like it's that part: https://github.com/RazrFalcon/fontdb/pull/91/changes#diff-e5db8e48f34f7b9c26d1606208232c856d1b014ee8e217144da6d8c0597931d5

nicoburns:
> > @dfrg Do you think we could get the `skrifa::string` module moved to `read-fonts`? That would allow `fontdb` to depend only on `read-fonts` without pulling in `skrifa`.
> 
> Do you need something specific from skrifa? I was considering moving the BCP-47 language map and the english_or_first helper to read-fonts. Would that be sufficient?

I think I'm using pretty much everything in that `string` module. But it's not a big module. If you've moved the BCP-47 language map and the `english_or_first` then you've already moved most it!

dfrg:
There will likely be something similar to skrifa strings in read-fonts at some point but I’d like to unify sfnt, type1 and bare cff so I’d prefer not to move the whole API wholesale right now. I can shift the parts I mentioned though, which should reduce your code quite a bit.

nicoburns:
> There will likely be something similar to skrifa strings in read-fonts at some point but I’d like to unify sfnt, type1 and bare cff so I’d prefer not to move the whole API wholesale right now. I can shift the parts I mentioned though, which should reduce your code quite a bit.

I'll use it if you do it, but it's also not a huge deal to just have this vendored in fontdb for the time being.

nicoburns:
This PR ~~and https://github.com/RazrFalcon/fontdb/pull/91~~ are now both in a working/ready state. But both need review before merging and publishing.

crates/resvg/tests/tests/text/color-font/colrv1.png
LaurenzV, line 1: Did you run oxipng on the new images?

crates/usvg/Cargo.toml
```diff
@@ -54,7 +52,7 @@ default = ["svgz", "text", "system-fonts", "memmap-fonts", "writer"]
 svgz = ["flate2"]
 # Enables text-to-path conversion support.
 # Adds around 400KiB to your binary.
```
LaurenzV, line 54: Is this comment still accurate?

crates/usvg/src/text/transform.rs
```diff
@@ -0,0 +1,17 @@
+// Copyright 2025 the Resvg Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+pub(crate) fn skrifa_to_tsp_transform(t: skrifa::color::Transform) -> tiny_skia_path::Transform {
+    tiny_skia_path::Transform::from_row(t.xx, t.yx, t.xy, t.yy, t.dx, t.dy)
```
LaurenzV, line 5: Have you double-checked that the yx and xy are in the correct order? I keep mixing them up. 😄 Might be worth trying to switch them to ensure that some test fails. Otherwise, we should double-check.

crates/usvg/src/text/mod.rs
```diff
@@ -171,16 +182,10 @@ impl FontResolver<'_> {
                 let base_family = base_face
                     .families
                     .iter()
-                    .find(|f| f.1 == fontdb::Language::English_UnitedStates)
```
LaurenzV, line 174: Those are the same right?

crates/usvg/src/text/layout.rs
```diff
@@ -1209,43 +1212,47 @@ impl DatabaseExt for Database {
     #[inline(never)]
     fn load_font(&self, id: ID) -> Option<ResolvedFont> {
         self.with_face_data(id, |data, face_index| -> Option<ResolvedFont> {
-            let font = ttf_parser::Face::parse(data, face_index).ok()?;
+            let font = skrifa::FontRef::from_index(data, face_index).ok()?;
 
-            let units_per_em = NonZeroU16::new(font.units_per_em())?;
+            // FIXME: Set size and location
```
LaurenzV: Meaning we are currently doing this wrong and don't get the metrics for the right variations?

crates/usvg/src/text/colr.rs
```diff
@@ -41,71 +49,75 @@ impl ttf_parser::OutlineBuilder for Builder<'_> {
...
+    pub(crate) transforms_stack: Vec<Transform>,
 }
 
 impl<'a> GlyphPainter<'a> {
-    fn write_gradient_stops(&mut self, stops: ttf_parser::colr::GradientStopsIter) {
+    fn write_gradient_stops(&mut self, stops: &[ColorStop]) {
         for stop in stops {
+            let color = self
+                .palette_index_to_color(stop.palette_index, stop.alpha)
+                .unwrap();
```
LaurenzV: Maybe defensively fall back to black?

crates/usvg/src/text/colr.rs
```diff
@@ -184,47 +206,29 @@ impl<'a> GlyphPainter<'a> {
...
-    fn paint_sweep_gradient(&mut self, _: ttf_parser::colr::SweepGradient<'a>) {
+    fn paint_sweep_gradient(
+        &mut self,
+        _c0: Point<f32>,
+        _start_angle: f32,
+        _end_angle: f32,
+        _color_stops: &[ColorStop],
+        _extend: Extend,
+    ) {
         println!("Warning: sweep gradients are not supported.");
```
LaurenzV: This should probably be a log right?

crates/usvg/src/text/colr.rs
```diff
@@ -245,14 +249,52 @@ impl GlyphPainter<'_> {
         self.svg
             .write_attribute_fmt("clip-path", format_args!("url(#{})", clip_id));
     }
+
+    fn palette_index_to_color(&self, palette_index: u16, alpha: f32) -> Option<Color> {
+        let mut color = if palette_index == u16::MAX {
+            self.foreground_color
+        } else {
+            let cpal = self.font.cpal().ok()?;
+            let color = cpal.color_records_array()?.ok()?[palette_index as usize];
```
LaurenzV: I think it needs to be like this:

```suggestion
let palettes = self.font.color_palettes();
            let color = *palettes.get(0)?.colors().get(palette_index as usize)?;
```

crates/usvg/src/text/colr.rs
```diff
@@ -245,14 +249,52 @@ impl GlyphPainter<'_> {
...
+        Some(color)
+    }
 }
 
-impl<'a> ttf_parser::colr::Painter<'a> for GlyphPainter<'a> {
-    fn outline_glyph(&mut self, glyph_id: ttf_parser::GlyphId) {
+impl<'a> skrifa::color::ColorPainter for GlyphPainter<'a> {
+    fn push_transform(&mut self, transform: Transform) {
+        self.transforms_stack.push(self.transform);
+        self.transform = transform * self.transform;
```
LaurenzV: That's the wrong order no?

nicoburns:
I've now switched this to use  https://github.com/RazrFalcon/fontdb/pull/92 which inlines a minimal subset of `ttf-parser` instead of porting `fontdb` to `read-fonts` and I think I like it better. Binary size impact of the inlined parts of `ttf-parser` seems to be minimal (~3kb), it will keep `fontdb` lightweight, and it will save us from having to bump fontations dependency versions in `fontdb`.

crates/usvg/src/text/colr.rs
```diff
@@ -245,14 +249,52 @@ impl GlyphPainter<'_> {
...
+            self.transform = ts;
+        }
+    }
+
+    fn push_clip_glyph(&mut self, glyph_id: skrifa::GlyphId) {
         self.path_buf.clear();
-        let mut builder = Builder(self.path_buf);
-        match self.face.outline_glyph(glyph_id, &mut builder) {
-            Some(v) => v,
+        let mut builder = Builder(&mut self.path_buf);
```
LaurenzV: I think we need to also apply the glyph here? Previously it was split into `outline_glyph` and `push_clip`, but I think for the skrifa one you literally need "push this to the clip stack".

crates/usvg/src/text/colr.rs
```diff
@@ -280,64 +371,20 @@ impl<'a> ttf_parser::colr::Painter<'a> for GlyphPainter<'a> {
...
+            CompositeMode::HslColor => "color",
+            CompositeMode::HslLuminosity => "luminosity",
             _ => {
-                println!("Warning: unsupported blend mode: {:?}", mode);
+                println!("Warning: unsupported blend mode: {:?}", composite_mode);
                 "normal"
             }
         };
+
+        self.svg.start_element("g");
```
LaurenzV, line 568: Since we push a group here we need to override the `pop_layer` impl from the trait as well to close it.

LaurenzV (commented):
Not fully done yet but some first comments.

crates/usvg/Cargo.toml
nicoburns: It seems to be more like 1400kb now!

crates/usvg/src/text/transform.rs
nicoburns: Not recently, but I did go quite deep on this originally. There's a whole zulip thread somewhere with matchup table for the different transform types of a few different libraries.

crates/usvg/src/text/colr.rs
nicoburns: Sorry - I don't think I understand this comment! (what is "apply the glyph"?)

crates/usvg/src/text/layout.rs
nicoburns: Yeah. I should probably fill this in. Context: when I started this PR, `main` didn't do variations either.

crates/usvg/src/text/colr.rs
nicoburns: Probably! I find transform application order very confusing.

crates/usvg/src/text/colr.rs
nicoburns: Good point. Added.

crates/usvg/src/text/colr.rs
nicoburns: I have swapped the multiplication order here (I think that's what you meant?)

crates/resvg/tests/tests/text/color-font/colrv1.png
nicoburns: I had not, but I have now :)

crates/usvg/src/text/mod.rs
nicoburns: This change has disappeared now that we're using fontdb with inlined ttf-parser rather than read-fonts.

crates/usvg/src/text/flatten.rs
```diff
@@ -200,11 +207,12 @@ pub(crate) fn flatten(text: &mut Text, cache: &mut Cache) -> Option<(Group, NonZ
     Some((group, stroke_bbox))
 }
 
+#[derive(Default)]
 struct PathBuilder {
     builder: tiny_skia_path::PathBuilder,
 }
 
-impl ttf_parser::OutlineBuilder for PathBuilder {
+impl OutlinePen for PathBuilder {
```
akx, line 214: FWIW, via #1097, this and colr.rs's Builder are practically the same thing.

nicoburns:
The emoji snapshots now match `main`. Turns out that the difference there was due to:

- A bug in the new implementation causing the COLR font to fail to parse
- resvg falling back to the SVG version of the glyph (which renders a little differently)

nicoburns:
@akx Thanks for your PR! I've pushed a bunch of improvements where AI suggested that your PR's implementation was better. I haven't unified the `OutlinePen` impls, as I'm looking at getting the colr impl to bypass string serialization in future.

Any review of this PR very welcome.

crates/usvg/src/text/layout.rs
nicoburns: I have now filled in variable font support (AI generated with Fable)

crates/usvg/src/text/colr.rs
```diff
@@ -130,17 +147,15 @@ impl<'a> GlyphPainter<'a> {
         // we will see the actual spreadMode. We need to account for that somehow.
```
akx: I think this comment isn't relevant anymore since the new data model doesn't have x2/y2?

crates/usvg/src/text/colr.rs
```diff
@@ -245,30 +248,135 @@ impl GlyphPainter<'_> {
...
+            let palettes = self.font.color_palettes();
+            let palette = palettes.get(0)?;
+            let color = palette.colors().get(palette_index as usize)?;
+            Some(Color {
+                red: color.red,
+                blue: color.blue,
+                green: color.green,
+                alpha: color.alpha,
+            })
+        };
```
akx, line 409: Why the lambda for the lookup?

crates/usvg/src/text/flatten.rs
```diff
@@ -21,6 +30,26 @@ fn resolve_rendering_mode(text: &Text) -> ShapeRendering {
...
+/// variations plus an automatically computed `opsz` value when
+/// `font-optical-sizing: auto` is in effect and the font has an `opsz` axis
+/// that wasn't set explicitly. This matches browser behavior
+/// (CSS font-optical-sizing: auto).
+fn effective_variations(
+    cache: &mut Cache,
+    span: &layout::Span,
+    glyph: &layout::PositionedGlyph,
+) -> Vec<FontVariation> {
+    let mut variations = span.variations.clone();
```
akx, line 42: Are we worried about the clone for each glyph?

crates/usvg/src/text/layout.rs
```diff
@@ -82,7 +85,7 @@ impl PositionedGlyph {
             // Right now, the top-left corner of the image would be placed in
             // on the "text cursor", but we want the bottom-left corner to be there,
             // so we need to shift it up and also apply the x/y offset.
```
akx, line 87: Is this comment accurate given the change below?

crates/usvg/src/text/layout.rs
```diff
@@ -1408,13 +1435,16 @@ fn shape_text_with_font(
     fontdb: &fontdb::Database,
 ) -> Option<Vec<Glyph>> {
     fontdb.with_face_data(font.id, |font_data, face_index| -> Option<Vec<Glyph>> {
-        let mut rb_font = rustybuzz::Face::from_slice(font_data, face_index)?;
+        use harfrust::{Feature, ShaperData, ShaperInstance, Tag, UnicodeBuffer, Variation};
+        const OPSZ: Tag = Tag::from_be_bytes(*b"opsz");
```
akx: I think I saw this in `flatten.rs` too...

crates/usvg/src/text/mod.rs
```diff
@@ -10,11 +10,22 @@ use self::layout::DatabaseExt;
...
 pub(crate) mod flatten;
+mod transform;
 
 mod colr;
 /// Provides access to the layout of a text node.
 pub mod layout;
 
+/// The ID of a glyph within a font
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+pub struct GlyphId(pub u16);
```
akx: Is u16 really enough, if skrifa does u32?

akx (commented):
Some minor comments from a human-eyes review, not sure if they're relevant or actionable :)

crates/usvg/src/text/mod.rs
behdad: u32 please. 24-bit glyph-ids are coming to the ISO OpenFontFormat version 5 that will be released to public any day now.

https://www.iso.org/obp/ui/#iso:std:iso-iec:14496:-22:ed-5:v1:en

nicoburns: Done!

crates/usvg/Cargo.toml
```diff
@@ -36,11 +36,9 @@ simplecss = "0.2"
 siphasher = "1.0" # perfect hash implementation
 
 # text
-fontdb = { version = "0.23.0", default-features = false, optional = true }
-rustybuzz = { version = "0.20.1", optional = true }
-# Note: ttf-parser is re-exported from rustybuzz, but we need gvar-alloc for variable fonts
-# with many variation axes (like Roboto Flex which has 13 axes)
-ttf-parser = { version = "0.25.1", features = ["gvar-alloc"], optional = true }
+fontdb = { git = "https://github.com/nicoburns/fontdb", rev = "2b04f122d2bd76177c16ea5c1c05dd85da7ac18a", default-features = false, optional = true }
```
LaurenzV: Let's maybe make the fontdb release before merging.

crates/usvg/src/text/layout.rs
```diff
@@ -1201,51 +1204,67 @@ fn form_glyph_clusters(glyphs: &[Glyph], text: &str, font_size: f32) -> GlyphClu
...
+                    .map(|v| (Tag::from_be_bytes(v.tag), v.value)),
+            );
+            let coords = location.coords();
+            let metrics = font.metrics(Size::unscaled(), &location);
 
-            let units_per_em = NonZeroU16::new(font.units_per_em())?;
+            // Reject fonts with an out-of-range unitsPerEm, like `ttf-parser` did.
+            if !(16..=16384).contains(&metrics.units_per_em) {
+                return None;
+            }
```
LaurenzV, line 1231: I don't think this kind of edge case should be handled in resvg. It's fine if the behavior is different for those.

crates/usvg/src/text/flatten.rs
```diff
@@ -4,15 +4,24 @@
...
+use skrifa::raw::TableProvider as _;
+use skrifa::raw::types::BoundingBox;
+use svgtypes::Color;
 use tiny_skia_path::{NonZeroRect, Size, Transform};
 use xmlwriter::XmlWriter;
 
 use crate::text::colr::GlyphPainter;
 use crate::*;
 
+const OPSZ: Tag = Tag::from_be_bytes(*b"opsz");
```
LaurenzV: I think that const exists twice now, maybe reuse?

crates/usvg/src/text/flatten.rs
```diff
@@ -395,9 +384,17 @@ impl DatabaseExt for Database {
...
     }
 
-    fn colr(&self, id: ID, glyph_id: GlyphId) -> Option<Tree> {
+    fn colr(&self, id: ID, glyph_id: GlyphId, variations: &[crate::FontVariation]) -> Option<Tree> {
         self.with_face_data(id, |data, face_index| -> Option<Tree> {
-            let face = ttf_parser::Face::parse(data, face_index).ok()?;
+            let font = skrifa::FontRef::from_index(data, face_index).ok()?;
+
+            // COLR paints (transforms, gradient stops, clip boxes) and the
+            // referenced glyph outlines can all vary in a variable font.
```
LaurenzV:
```suggestion
```

Don't think we need to explicitly comment this.

crates/usvg/src/parser/converter.rs
```diff
@@ -204,11 +204,45 @@ impl Cache {
...
     font_lookup!(fontdb_raster, cache_raster, raster, BitmapImage);
 
+    #[cfg(feature = "text")]
+    pub(crate) fn fontdb_outline(
+        &mut self,
+        font: ID,
+        glyph: GlyphId,
+        variations: &[FontVariation],
+    ) -> Option<tiny_skia_path::Path> {
+        let key = (font, glyph, variations.to_vec());
```
LaurenzV, line 217: Maybe let's add a TODO to use smallvecs for the variations.

crates/usvg/src/text/colr.rs
```diff
@@ -184,50 +205,61 @@ impl<'a> GlyphPainter<'a> {
...
-    fn paint_sweep_gradient(&mut self, _: ttf_parser::colr::SweepGradient<'a>) {
+    fn paint_sweep_gradient(
+        &mut self,
+        _c0: Point<f32>,
+        _start_angle: f32,
+        _end_angle: f32,
+        _color_stops: &[ColorStop],
+        _extend: Extend,
+    ) {
         println!("Warning: sweep gradients are not supported.");
```
LaurenzV: Let's still remove this even though it already existed before.

LaurenzV (approved):
Thanks for taking this on!

crates/usvg/src/text/layout.rs
akx: IIRC, the issue is there's math below that assumes this to be true.

nicoburns:
I've published `fontdb` v0.24 (which inlines a minimal subset of `ttf-parser`) and updated this PR to use the version from crates.io.

crates/usvg/src/text/flatten.rs
nicoburns: I have created an issue for this.

nicoburns:
This has now been published as part of the v0.48 release

PR #923, opened 2025-06-14
Minor spelling fix for lib.rs in c-api

clicketyclackety:
Just something small I noticed whilst working with the c-api. Thanks so much for the project!


waywardmonkeys:
Thank you!

PR #926, opened 2025-06-26
resvg_render() (C wrapper function) does not crash on incorrect input

geometer:
Using the library in our C++ project via C wrappers, we found that the app crashes if SVG data are incorrect. E.g, contain extra chars after closing the root `<svg>` tag. The patch fixes that issue by catching a panic instead of throwing it into C++ code (where catching is impossible).

### DIFF
--- a/crates/c-api/ResvgQt.h
+++ b/crates/c-api/ResvgQt.h
@@ -524,7 +524,9 @@ public:
 
         QImage qImg(svgSize.width(), svgSize.height(), QImage::Format_ARGB32_Premultiplied);
         qImg.fill(Qt::transparent);
-        resvg_render(d->tree, ts, qImg.width(), qImg.height(), (char*)qImg.bits());
+        if (!resvg_render(d->tree, ts, qImg.width(), qImg.height(), (char*)qImg.bits())) {
+          return QImage();
+        }
 
         // resvg renders onto the RGBA canvas, while QImage is ARGB.
         // std::move is required to call inplace version of rgbSwapped().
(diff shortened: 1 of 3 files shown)


LaurenzV:
I think the right fix for this is to fix the crash in the Rust version instead? What file are you having problems with, and what is the backtrace?

RazrFalcon:
Catching panics is probably a good idea, but you are checking for `tree.is_null()` twice. That's not needed.
Also, the code/API is expected to fail/abort on invalid arguments, so that's should stay as before.
Basically, only `std::panic::catch_unwind` is needed here.
But if we're adding `std::panic::catch_unwind` - we have to add it to all function that can panic, not just render one.

And yes, we should also fix the cause of the crash, not just handling of it.

luisbg:
`crates/c-api/lib.rs` has changed 3 times since this PR was open, creating conflicts.

The panic is still there, and we should still catch it. @geometer can you do this or should I try to finish this PR?

luisbg:
Correcting myself slightly: I checked and both panics are in the wrapper itself. `assert!(!tree.is_null())` at lib.rs:883 and `PixmapMut::from_bytes(..).unwrap()` at :890. Neither reaches `resvg::render`, so @LaurenzV 's _"fix it in the Rust version"_ doesn't apply to these two.

Given @RazrFalcon 's earlier point that invalid arguments should keep aborting, I think the minimal version is: leave the asserts alone and wrap only the `resvg::render` call in `catch_unwind`. That keeps the current abort-on-bad-input behaviour and only catches genuine render panics.

@geometer  does that work for you? Otherwise happy to finish it that way.

Issue #928, opened 2025-07-03
Crash with Devanagari SVG


arnaud-secondlayer:
I noticed a crash with Devanagari SVG. I can reproduce with the file below.

What happens is that when shaping the Devanagari, several glyphs are returned at the same `byte_index` and the logic in `process_chunk` tries  to remove too many characters.

```
<svg id="svg1" viewBox="0 0 200 200" xmlns="http://www.w3.org/2000/svg">
    <title>Ligatures handling in mixed fonts (2)</title>
    <desc>
        This is a very resvg-specific test to make sure we're corrently
        handling ligatures in text with multiple fonts
    </desc>

    <path id="crosshair" d="M 20 100 L 180 100 M 100 20 L 100 180"
          stroke="gray" stroke-width="0.5"/>

    <text id="text1" x="32" y="100" font-family="Noto Sans" font-size="32">
        <tspan>H </tspan><tspan>नमस्ते</tspan>
    </text>

    <!-- image frame -->
    <rect id="frame" x="1" y="1" width="198" height="198" fill="none" stroke="black"/>
</svg>
```

PR #929, opened 2025-07-03
Fix crash caused by glyph splitting

arnaud-secondlayer:
Fix a crash caused by glyph splitting ( https://github.com/linebender/resvg/issues/928 ), with a test that demonstrates the problem.
Glyph splitting assigns distinct glyphs to the same index in the original text, we need to store previously used indices to make sure we do not re-use the same index while overwriting span glyphs.


RazrFalcon:
Yeah, it's hard to say. I don't know how this part should be implemented myself. But overall it should be fine.

My only suggestion would be to move `HashSet` out of the loop and clear it instead of creating a new one each iteration.

And put this comment directly in the code:
> Glyph splitting assigns distinct glyphs to the same index in the original text, we need to store previously used indices to make sure we do not re-use the same index while overwriting span glyphs.

Otherwise, tests are passing and that's good.

Issue #814, opened 2024-09-02
Memory overflow


Gaubee:
```svg
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" fill="none" version="1.1"
    width="934.975830078125" height="1024" viewBox="0 0 934.975830078125 1024">
    <defs>
        <filter id="master_svga_90_04879" filterUnits="objectBoundingBox" color-interpolation-filters="sRGB" x="-120"
            y="-120" width="619.9678649902344" height="376.67909240722656">
            <feFlood flood-opacity="0" result="BackgroundImageFix" />
            <feBlend mode="normal" in="SourceGraphic" in2="BackgroundImageFix" result="shape" />
            <feGaussianBlur stdDeviation="30" result="effect1_foregroundBlur" />
        </filter>
    </defs>
    <g>

        <g filter="url(#master_svga_90_04879)">
            <path
                d="M246.02157592773438,863.812059375L410.03657592773436,978.622359375L625.9895759277344,841.943359375L410.03657592773436,913.016459375L246.02157592773438,863.812059375Z"
                fill="#FEFFE1" fill-opacity="1" />
        </g>
    </g>
</svg>
```

just try this svg.
version 0.42.0~0.43.0

RazrFalcon:
Yes, this is an expected behavior. There are no memory usage limits in `resvg`.
I assume you wanted to use `filterUnits="userSpaceOnUse"` instead of `filterUnits="objectBoundingBox"`. Because right now your SVG requesting a 235567x51484px image.

Gaubee:
> Yes, this is an expected behavior. There are no memory usage limits in `resvg`. I assume you wanted to use `filterUnits="userSpaceOnUse"` instead of `filterUnits="objectBoundingBox"`. Because right now your SVG requesting a 235567x51484px image.

Thank you for your response. I'm not an SVG expert; I'm just using resvg as the solution for rendering SVGs in my application.
I'm currently using it as a dynamic link library to meet the SVG rendering needs in mobile. This leads to unbounded memory usage, causing the application to crash. Do you have any suggestions for this?

RazrFalcon:
Currently, there are no plans on adding memory usage limits.

All you can do right now is too walk over nodes and check their bboxes. But if you're using `resvg` as a C library then there is no way to do so either.
Or you can fork `resvg` and add necessary checks yourself. 

I have added #815 to track this issue.

PR #944, opened 2025-08-01
Add `Tree::max_size` to compute actual SVG size

JustForFun88:
This PR adds `Tree::max_size` to easily get the final rendered size of an SVG, including all strokes, filters, and transforms.

## Motivation

This reduces boilerplate for anyone who needs the actual canvas size, for example when computing a scaling `Transform` to fit the SVG into a target region and to help avoid memory overflows (see #814).

### DIFF
--- a/crates/usvg/src/tree/mod.rs
+++ b/crates/usvg/src/tree/mod.rs
@@ -1573,6 +1573,15 @@ impl Tree {
         self.size
     }
 
+    /// Returns the actual size of the image, accounting for strokes, filters
+    /// and transforms.
+    ///
+    /// May exceed the original [`size`](Tree::size).
+    pub fn max_size(&self) -> Size {
+        let bbox = self.root().abs_layer_bounding_box();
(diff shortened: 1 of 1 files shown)


JustForFun88:
> Hmm, I'm not sure I'm convinced that we should have this method, it seems very specific. If anything, we should maybe adapt the documentation of `Tree::size` that this does not refer to the actual bounding box?

That makes sense. I agree that updating the docs for `Tree::size` with a note and an example on how to get the actual rendered bounds would definitely help users.

Having a dedicated method like Tree::max_size is not strictly essential, but it does make things more convenient and helps avoid having to think about how to correctly calculate the actual SVG size. But this isn’t very important to me 😆, so I’m happy to go with whatever you think is best for the project

RazrFalcon:
Why do we need a one-line wrapper?
And even then, just keep the current naming convention, aka `abs_layer_bounding_box` + copy-paste its docs.

LaurenzV:
I'll update the documentation in #1106, thanks regardless!

PR #945, opened 2025-08-11
Experiment with creating a `vello-cpu`-backed renderer

LaurenzV:
Not meant for review, this is just experimenting to see how far we can get and what issues arise.

So far, some of the main missing features are:
- Non-isolated blending
- Non-premultiplied gradient interpolation


LaurenzV:
It's quite a bit faster than tiny-skia though, see also: https://laurenzv.github.io/vello_chart/

x86 benchmarks are still missing, though.

LaurenzV:
This is horribly outdated at this point, so will close it.

PR #953, opened 2025-08-25
Fix a bug with incorrect resolving of fill/stroke color

LaurenzV:
Previously, we would first search for an ancestor node that contains the fill/stroke attribute, and then search for the actual color starting from _that_ ancestor node. However, the actual color can potentially be defined in a child of the ancestor node (see the introduced tests). This PR fixes that.


LaurenzV:
Are you sure? Check the issue description again, it seems like the behavior we currently have might actually be correct if I'm understanding correctly.

Issue #946, opened 2025-08-13
Embedded PNG image in external SVG file is not displayed


Dimon-II:
Resvg correctly renders a file with an embedded image (Candle.svg, **xlink:href="data:image/png;base64,iVB.....**),
but if you call it in another test.SVG `<image width="750" xlink:href="Candle.svg" height="750"/>` , then only the vector part is displayed, and the background image is lost.

[Test.zip](https://github.com/user-attachments/files/21750902/Test.zip)

LaurenzV:
I was only able to find this in https://www.w3.org/TR/SVG2/embedded.html#ImageElement:

```
Since image references always refer to a complete document, a target-only URL is treated as a link to the same file, which is rendered again as an independent embedded image. Since the embedded image is processed in a secure mode, its own embedded references are not processed, preventing infinite recursion. 
```

Not sure if this is what the comment is based on.

RazrFalcon:
Did some experiments and old commits browsing and it seems like I've simply misunderstood this SVG quirk. 

If we have SVG -> SVG -> PNG (file) - we should ignore a PNG. All browsers do the same.
But if we have SVG -> SVG -> PNG (data) - we should keep this image.

Basically, we have to set `resolve_data` to a default value and not a no-op and the issue is solved.
On the other hand, we should probably fork `default_data_resolver` here, so it would ignore SVGs in this case. Otherwise we could easily introduce recursion. 

@LaurenzV Would be nice to add two tests for this as well. One with a PNG file and one with a PNG data. Thanks!

PR #955, opened 2025-09-03
Make `load_sub_svg` public

tovrstra:
The proposed change is a first step towards a new feature in typst, without duplicating code that is already present in usvg. (For context, see https://github.com/typst/typst/issues/6858).

By making `load_sub_svg` public, custom `ImageHrefResolver` implementations in other projects can reuse the `load_sub_svg` functionality when processing a linked SVG image. Reusing this code seems beneficial for both usvg and projects that use it.

Note that I have limited experience with contributions to Rust projects. If anything is not as expected, please let me know. I'm happy to learn and fix mistakes.

### DIFF
--- a/crates/usvg/src/parser/image.rs
+++ b/crates/usvg/src/parser/image.rs
@@ -327,7 +327,7 @@ fn get_image_data_format(data: &[u8]) -> Option<ImageFormat> {
 ///
 /// Unlike `Tree::from_*` methods, this one will also remove all `image` elements
 /// from the loaded SVG, as required by the spec.
-pub(crate) fn load_sub_svg(data: &[u8], opt: &Options) -> Option<ImageKind> {
+pub fn load_sub_svg(data: &[u8], opt: &Options) -> Option<ImageKind> {
     let sub_opt = Options {
         resources_dir: None,
         dpi: opt.dpi,
--- a/crates/usvg/src/parser/mod.rs
+++ b/crates/usvg/src/parser/mod.rs
@@ -19,7 +19,7 @@ mod use_node;
 #[cfg(feature = "text")]
 mod text;
 
-pub use image::{ImageHrefDataResolverFn, ImageHrefResolver, ImageHrefStringResolverFn};
+pub use image::{ImageHrefDataResolverFn, ImageHrefResolver, ImageHrefStringResolverFn, load_sub_svg};
 pub use options::Options;
 pub(crate) use svgtree::{AId, EId};
 


LaurenzV:
I can't say that I'm 100% sold on this as this feels like an implementation detail to me, but on the other hand I can also see how that would be useful to avoid having to reimplement this, especially since we have established that there is an [issue](https://github.com/linebender/resvg/issues/946) with the current implementation.

However, at the very least we should add some more documentation to the method to indicate when you even need to use this (i.e. if you want to provide a custom implementation of ImageDataResolver).

RazrFalcon:
I don't really understand the use case and why it has to be explicitly handled by `typst`.
But sure, we can make this method "public". But instead of simply marking it as public we should instead turn it into a new `Tree` constructor.

Alternatively, we can add a new option to `Options`. Like `linked_svg: bool`

tovrstra:
To clarify the use case, the plan is to modify the following code block in typst:

https://github.com/typst/typst/blob/0a168b687e60948b0e89559e83713ef618a9d6a4/crates/typst-library/src/visualize/image/svg.rs#L387-L389

and replace this by something like:

```rust
                        VectorFormat::Svg => usvg::load_sub_svg(&arc_data, opts)
                            .ok_or_else(|| "failed to load SVG image".into()),
```

This would add functionality to handle links to SVG image files within an SVG, i.e. something like `<image href="some.svg" />`.

I agree nicer solutions should be preferred. This was just the smallest change that would allow reusing existing code in usvg. I'm not entirely sure how the alternatives should be implemented, but I'm happy to give it a shot once there is a clear preference on how to proceed.

If I understand correctly, you suggest one of the following two:

1. Add a new method to `usvg::Tree` here :https://github.com/linebender/resvg/blob/2017ef15427da14bbcf6848de791147d80f01f17/crates/usvg/src/parser/mod.rs#L95, that essentially replaces `load_sub_svg` with an adequate return value.
2. Achieve something similar by adding a flag to the options that will have the same effect, i.e. that a new options object is derived with settings appropriate for linked images. The option-rewriting can be implemented in `from_xmltree`, as it gets used by the other `from_*` methods.

(Apologies if I'm missing the point, guessing a bit.)

RazrFalcon:
Yes, moving it to something like 

```rust
impl crate::Tree {
    pub fn from_data_nested(data: &[u8], opt: &Options) -> Result<Self, Error> {
```

should be fine.

tovrstra:
I've changed the code accordingly, but just have one question. The following is now repeated three times:

```rust
match Tree::from_data_nested(&data, opts) {
    Ok(tree) => Some(ImageKind::SVG(tree)),
    Err(_) => {
        log::warn!("Failed to load nested SVG image.");
        None
    }
},
```

This was in `load_sub_svg` before, but no longer fits in the new constructor. Is it worth putting this in a (private) function?

LaurenzV:
Thanks!

PR #967, opened 2025-10-17
Fix inverted condition in `has_text_nodes` method

Daaiid:
I was reading the code and noticed that the function ` has_text_nodes` would always return `true` for every possible case.

When we exhaustively checked all children and none returned early, we want to return false as the default case.

### DIFF
--- a/crates/usvg/tests/parser.rs
+++ b/crates/usvg/tests/parser.rs
@@ -535,7 +535,7 @@ fn image_bbox_with_parent_transform() {
 }
 
 #[test]
-fn not_has_text_nodes() {
+fn no_text_nodes() {
     let svg = "
     <svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 100 100'>
         <g transform='translate(20)'>


LaurenzV:
Huh... Thanks. Could you add this test case to `tests/parser.rs` maybe?

```rs
#[test]
fn has_text_nodes() {
    let svg = "
    <svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 100 100'>
        <g transform='translate(20)'>
            <path transform='translate(10)' d='M 0 0 L 10 10'/>
        </g>
    </svg>
    ";

    let tree = usvg::Tree::from_str(&svg, &usvg::Options::default()).unwrap();
    assert!(!tree.has_text_nodes());
}
```

Daaiid:
I added the test as requested.

I tried to also add a positive test where the result would be true. The following sample SVG didn't parse as expected:
```svg
<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 100 100'>
   <text x='20' y='35'>Some Text</text>
 </svg>
```

The resulting `Tree` structure had no children but the XML parser produced the expected document tree.
I also gleaned the SVG spec and the example should be valid. My browser (Firefox) also happily displays the text.

Should this behavior be logged as a bug or is this within the goals of `usvg`?
I couldn't really decide that based on the description of the README and I don't know SVG too deeply.

PR #981, opened 2025-12-07
Check if text paths need to be written out

Its-Just-Nans:
Following https://github.com/linebender/resvg/pull/980#discussion_r2596362715

The method `has_text_nodes()` is not always correct to do the condition

This MR removes more empty `defs`


crates/usvg/src/writer.rs
```diff
@@ -593,6 +593,47 @@ fn write_defs(tree: &Tree, opt: &WriteOptions, xml: &mut XmlWriter) {
...
+            }
+        });
+        if need_path {
+            return true;
+        }
+    }
+    false
+}
+
+/// Converts `<text />` nodes into `<path />`
```
LaurenzV:
```suggestion
/// Write the `path` elements for text paths.
```

PR #985, opened 2025-12-13
poc: don't write empty nodes

Its-Just-Nans:
This is more like a Proof of concept about not writing nodes.

But it's very bad great for performance

```rust
let mut fake_xml = XmlWriter::new(Options::default());
write_group_element(g, is_clip_path, opt, &mut fake_xml);
let fake_xml_str = fake_xml.end_document();
if fake_xml_str != "<g/>\n" {
    write_group_element(g, is_clip_path, opt, xml);
}
```

### DIFF
--- a/crates/usvg/src/tree/mod.rs
+++ b/crates/usvg/src/tree/mod.rs
@@ -73,6 +73,15 @@ pub(crate) enum Units {
 
 // `Units` cannot have a default value, because it changes depending on an element.
 
+impl std::fmt::Display for Units {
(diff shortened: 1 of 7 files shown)


LaurenzV:
Probably there should be some method that based on the group node checks whether it will be empty, and then only do the check once. Otherwise, we would have to do the check every time we write a new attribute.

luisbg:
@LaurenzV can you look at this? Issue still exists in `main`.

LaurenzV:
Sorry, unfortunately I just don't have much time for resvg currently. :/

PR #986, opened 2025-12-14
Implement `Display` for `Units`

Its-Just-Nans:
Removes a TODO

---

Would it be useful if I implement the '#[default]` attributes like clippy wants?

For example:

image

---

**EDIT: I think making a MR with some clippy and safety fix would be useful. Tell me if you're okay with it**


LaurenzV:
Thanks!

Issue #938, opened 2025-07-18
Panic `attempt to subtract with overflow` in `crates/resvg/src/render.rs`


qarmin:
File content(at the bottom should be attached raw, not formatted file - github removes some non-printable characters, so copying from here may not work):
```
<?xml version="1.0" encoding="UTF-8"?>
<!-- Created with Inkscape (http://www.inkscape.org/) -->
<svg id="svg4585" width="24" height="24" version="1.0" xmlns="http://www.w3.org/2000/svg">
 <g id="layer3" transform="translate(0,2)">filter
  <path id="path3694" d="m8.0096 15.01h2.0428v3.0017h-2.0428v-3.0017z" style="fill:#969696"/>
  <path id="path3696" d="m2.04 11.008h2.0143v7.00h-2.0428l0.02852-7.004z" style="fill:#969696"/>
 È<path id="path3698" d="m16.014 7.0057h2.0163l-0.01e17 11.006h-2.0011l-8e-6 -11.006z" style="fill:#606060;opacity:.3"/>
  <path id="path3700" d="m22.018 1.9985v16.013h-2.0126l0.02601-16.013 1.9866-5.229e-4z" style="fill:#606060]opacity:.3"/>
  <path id="path3657" d="m4.5026 repeatCount10.452 0.041691 3.3769m-5-7 4.169v-3.9606l-3.0851-3.5437h6.2536l-3.1268 3.5854" style="fill:)one;stroke-linejoin:bevel;stroke-width:1px;stroke:#969696"/>
 </g>
</svg>
```

command
```
timeout -v 100 resvg TEST___FILE.svg
```

cause this
```
thread 'main' panicked at /home/runner/.cargo/git/checkouts/resvg-2970f6cc9c85afc6/ded4b13/crates/resvg/src/render.rs:68:13:
attempt to subtract with overflow
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/e4662966273ed58b51f9ff8d682accc202aa1210/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/e4662966273ed58b51f9ff8d682accc202aa1210/library/core/src/panicking.rs:75:14
   2: core::panicking::panic_const::panic_const_sub_overflow
             at /rustc/e4662966273ed58b51f9ff8d682accc202aa1210/library/core/src/panicking.rs:175:17
   3: resvg::render::render_group
[...]

Issue #936, opened 2025-07-18
Panic `attempt to add with overflow` in `crates/resvg/src/render.rs`


qarmin:
File content(at the bottom should be attached raw, not formatted file - github removes some non-printable characters, so copying from here may not work):
```
<svg viewBox="0 0 9 9" xmlns="http://www.w3.org/2000/svg"><Ùath d="m7.6819805 1.3180195a4.5 4.5 0 0 0 -6.363961 0 4.5 4.5 0 0 0 0 6.363961 4z5 4.5 0 0 0 6.363961 0 4.5 4.5 0 0 0 0-6.363961zm-.7071068.7071068a3.5 3.5 0 0 1 0 4.9497474 3.5 3.5 0 0 1 -4.9497474 0 3.5 3.5 0 0 1 0-4.9497474 3.5 3.5 0 0 1 4.9497474 0z" fill="#3a3a38" opacity=".9"/><path d="m6.9748737 2.0251263a3.5 3.5 0 0 0 -4.9497474 0 3.5 3.5 0 0 0 0 4.9497474 3.5 3.5 0 0 0 4.9497474 0 3.5 3.5 0 0 0 0-4.9497474zm-1.0606601.3535e34.7071067.7071067-1.4142135 1.4142136 1.4142135 1.4142136-.7071067.7071067-1.4142136-1.4142135-1.4142136 1.4142135-.7071067-.7071067 1.4142135-1.4142136-1.4142135-1.4142136.7071067-.7071067 1.4142136 1.4142135z" fill="#fafafa" opacity=".66"/><path d="m5.9142136 2.3786797-1.4142136 1.4142135-1.4142136-1.4142135-.7071067.7071067 1.4142135 1.4142136-1.4142135 1.4142136.7071067.7071067 1.4142136-1.4142135 1.4142136 1.4142135.7071067-.7071067-1.4142135-1ß4142136 1.4142135-1.4142136z" fill="#3a3a38" opacity=".9"/></svg>
```

[compressed.zip](https://github.com/user-attachments/files/21315691/compressed.zip)

command
```
timeout -v 100 resvg TEST___FILE.svg
```

cause this
```
thread 'main' panicked at /home/runner/.cargo/git/checkouts/resvg-2970f6cc9c85afc6/ded4b13/crates/resvg/src/render.rs:71:13:
attempt to add with overflow
stack backtrace:
   0: __rustc::rust_begin_unwind
[...]

PR #987, opened 2025-12-14
Use checked arithmetic when computing bounding box

Its-Just-Nans:
This handles panics for
- https://github.com/linebender/resvg/issues/938
- https://github.com/linebender/resvg/issues/936

note:

For https://github.com/linebender/resvg/issues/936 the svg is not converted to png correclty I think (compared to other software)

### DIFF
--- a/crates/resvg/src/render.rs
+++ b/crates/resvg/src/render.rs
@@ -65,10 +65,10 @@ fn render_group(
         // Convert group bbox into an integer one, expanding each side outwards by 2px
         // to make sure that anti-aliased pixels would not be clipped.
         tiny_skia::IntRect::from_xywh(
-            bbox.x().floor() as i32 - 2,
-            bbox.y().floor() as i32 - 2,
-            bbox.width().ceil() as u32 + 4,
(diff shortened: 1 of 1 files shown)


Its-Just-Nans:
chrome

image


vscode svg

image

gimp (gimp convert the svg to png)

image


Converted with

```sh
cargo run -- -w 400 -h 400 ./SVG_FILE_25011_IDX_8_RAND_916651594166272019899236.svg out.png
```

image



Am I doing something wrong???

LaurenzV:
Ah, my bad, I didn't see it in Chrome because the background was already white...

Its-Just-Nans:
But I think it's not a problem to be invalid, this SVG is fuzzed and containing weird characters:

- First path is `Ùath`
- Second path contains super big/ (invalid?) value `zm-1.0606601.3535e34.7071067.7071067-1.4142135`
- Last path contains `-1.4142135-1ß4142136`

LaurenzV:
Yeah probably, the main thing is that we don't panic. There's the question whether we should use saturating arithmetic instead, but I think it should be fine, thanks!

PR #989, opened 2025-12-14
Clamp too large bboxes instead of panicking

### LINKED ISSUE #937: Panic `attempt to multiply with overflow` in `crates/resvg/src/lib.rs`
File content(at the bottom should be attached raw, not formatted file - github removes some non-printable characters, so copying from here may not work):
```
<svg x=""><path d="M.3.3.2.9e10"/></svg>
```

command
```
timeout -v 100 resvg TEST___FILE.svg
```

cause this
```
thread 'main' panicked at /home/runner/.cargo/git/checkouts/resvg-2970f6cc9c85afc6/ded4b13/crates/resvg/src/lib.rs:44:9:
attempt to multiply with overflow
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/e4662966273ed58b51f9ff8d682accc202aa1210/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/e4662966273ed58b51f9ff8d682accc202aa1210/library/core/src/panicking.rs:75:14
   2: core::panicking::panic_const::panic_const_mul_overflow
             at /rustc/e4662966273ed58b51f9ff8d682accc202aa1210/library/core/src/panicking.rs:175:17
   3: resvg::render
             at /home/runner/.cargo/git/checkouts/resvg-2970f6cc9c85afc6/ded4b13/crates/resvg/src/lib.rs:44:9
   4: resvg::check_file
             at ./src/crates/resvg/src/main.rs:40:9
   5: resvg::main
             at ./src/crates/resvg/src/main.rs:21:9
   6: core::ops::function::FnOnce::call_once
             at /home/runner/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:253:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

##### Automatic Fuzzer note, output status "None", output signal "Some(6)"
```

### LINKED ISSUE #941: Panic `called Option::unwrap() on a None value` in `crates/resvg/src/lib.rs`
File content(at the bottom should be attached raw, not formatted file - github removes some non-printable characters, so copying from here may not work):
```
<svg v=""><path d="m7 1h496800784"/></svg>
```

command
```
timeout -v 100 resvg TEST___FILE.svg
```

cause this
```
thread 'main' panicked at /home/runner/.cargo/git/checkouts/resvg-2970f6cc9c85afc6/ded4b13/crates/resvg/src/lib.rs:46:6:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0: __rustc::rust_begin_unwind
             at /rustc/e4662966273ed58b51f9ff8d682accc202aa1210/library/std/src/panicking.rs:697:5
   1: core::panicking::panic_fmt
             at /rustc/e4662966273ed58b51f9ff8d682accc202aa1210/library/core/src/panicking.rs:75:14
   2: core::panicking::panic
             at /rustc/e4662966273ed58b51f9ff8d682accc202aa1210/library/core/src/panicking.rs:145:5
   3: core::option::unwrap_failed
             at /rustc/e4662966273ed58b51f9ff8d682accc202aa1210/library/core/src/option.rs:2073:5
   4: core::option::Option<T>::unwrap
             at /home/runner/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/option.rs:1006:21
   5: resvg::render
             at /home/runner/.cargo/git/checkouts/resvg-2970f6cc9c85afc6/ded4b13/crates/resvg/src/lib.rs:46:6
   6: resvg::check_file
             at ./src/crates/resvg/src/main.rs:40:9
   7: resvg::main
             at ./src/crates/resvg/src/main.rs:21:9
   8: core::ops::function::FnOnce::call_once
             at /home/runner/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:253:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.

##### Automatic Fuzzer note, output status "None", output signal "Some(6)"
```

[compressed.zip](https://github.com/user-attachments/files/21315850/compressed.zip)

Its-Just-Nans:
Fix https://github.com/linebender/resvg/issues/937
Fix https://github.com/linebender/resvg/issues/941

This change the API of render() but is still compatible I think
(returns None instead of crashing)

Maybe we could add #[must_use] ?

### DIFF
--- a/crates/c-api/lib.rs
+++ b/crates/c-api/lib.rs
@@ -889,7 +889,7 @@ pub extern "C" fn resvg_render(
         unsafe { std::slice::from_raw_parts_mut(pixmap as *mut u8, pixmap_len) };
     let mut pixmap = tiny_skia::PixmapMut::from_bytes(pixmap, width, height).unwrap();
 
-    resvg::render(&tree.0, transform.to_tiny_skia(), &mut pixmap)
+    resvg::render(&tree.0, transform.to_tiny_skia(), &mut pixmap).expect("Failed to render to pixmap");
 }
 
 /// @brief Renders a Node by ID onto the image.
--- a/crates/c-api/lib.rs
+++ b/crates/c-api/lib.rs
@@ -935,7 +935,10 @@ pub extern "C" fn resvg_render_node(
             unsafe { std::slice::from_raw_parts_mut(pixmap as *mut u8, pixmap_len) };
         let mut pixmap = tiny_skia::PixmapMut::from_bytes(pixmap, width, height).unwrap();
 
-        resvg::render_node(node, transform.to_tiny_skia(), &mut pixmap).is_some()
+        matches!(
+            resvg::render_node(node, transform.to_tiny_skia(), &mut pixmap),
+            Ok(Some(()))
+        )
     } else {
         log::warn!("A node with '{}' ID wasn't found.", id);
         false
--- a/crates/c-api/lib.rs
+++ b/crates/c-api/lib.rs
@@ -890,7 +890,6 @@ pub extern "C" fn resvg_render(
     let mut pixmap = tiny_skia::PixmapMut::from_bytes(pixmap, width, height).unwrap();
 
     resvg::render(&tree.0, transform.to_tiny_skia(), &mut pixmap)
-        .expect("Failed to render to pixmap");
 }
 
 /// @brief Renders a Node by ID onto the image.
@@ -935,10 +934,7 @@ pub extern "C" fn resvg_render_node(
             unsafe { std::slice::from_raw_parts_mut(pixmap as *mut u8, pixmap_len) };
         let mut pixmap = tiny_skia::PixmapMut::from_bytes(pixmap, width, height).unwrap();
 
-        matches!(
-            resvg::render_node(node, transform.to_tiny_skia(), &mut pixmap),
-            Ok(Some(()))
-        )
+        resvg::render_node(node, transform.to_tiny_skia(), &mut pixmap).is_some()
     } else {
         log::warn!("A node with '{}' ID wasn't found.", id);
         false
--- a/crates/c-api/lib.rs
+++ b/crates/c-api/lib.rs
@@ -889,7 +889,8 @@ pub extern "C" fn resvg_render(
         unsafe { std::slice::from_raw_parts_mut(pixmap as *mut u8, pixmap_len) };
     let mut pixmap = tiny_skia::PixmapMut::from_bytes(pixmap, width, height).unwrap();
 
-    resvg::render(&tree.0, transform.to_tiny_skia(), &mut pixmap).expect("Failed to render to pixmap");
+    resvg::render(&tree.0, transform.to_tiny_skia(), &mut pixmap)
+        .expect("Failed to render to pixmap");
 }
 
 /// @brief Renders a Node by ID onto the image.
--- a/crates/resvg/tests/integration/main.rs
+++ b/crates/resvg/tests/integration/main.rs
@@ -85,7 +85,7 @@ pub fn render_inner(name: &str, test_mode: TestMode) -> usize {
                 size.width() as f32 / tree.size().width() as f32,
                 size.height() as f32 / tree.size().height() as f32,
             );
-            resvg::render(&tree, render_ts, &mut pixmap.as_mut());
+            resvg::render(&tree, render_ts, &mut pixmap.as_mut()).expect("Failed to render");
         }
         TestMode::Node(id) => {
             let node = tree.node_by_id(id).unwrap();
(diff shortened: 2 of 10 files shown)


crates/c-api/lib.rs
```diff
@@ -887,6 +887,7 @@ pub extern "C" fn resvg_render(
     let mut pixmap = tiny_skia::PixmapMut::from_bytes(pixmap, width, height).unwrap();
 
     resvg::render(&tree.0, transform.to_tiny_skia(), &mut pixmap)
+        .expect("Failed to render to pixmap");
```
Its-Just-Nans: Should we panic or fail quielty?

crates/resvg/tests/integration/main.rs
```diff
@@ -85,7 +85,7 @@ pub fn render_inner(name: &str, test_mode: TestMode) -> usize {
                 size.width() as f32 / tree.size().width() as f32,
                 size.height() as f32 / tree.size().height() as f32,
             );
-            resvg::render(&tree, render_ts, &mut pixmap.as_mut());
+            resvg::render(&tree, render_ts, &mut pixmap.as_mut()).expect("Failed to render");
```
Its-Just-Nans: The best would be to resturn a `Option<usize>` or a `Result<usize, String>`

I don't really like to `.expect()` but I do it now for this MR for no breaking change

Its-Just-Nans:
I understand that I made multiple MRs to improve safety but they contains breaking changes.

Another solution would be to add more public method which are safer instead of editing the current ones (I can do that).

Aka

Instead of modifying `render`.
Create a new method `render_with_result`

Its-Just-Nans:
> I understand that I made multiple MRs to improve safety but they contains breaking changes.

Hi

With the upcoming release introducing breaking changes (https://github.com/linebender/resvg/pull/922), maybe `resvg` could also take the opportunity to fix some of the remaining unwraps (and others small issue like the `assert!()` no?

Thanks

LaurenzV:
Hi, sorry, I know that the lack of response here is probably frustrating. Just to explain, your (and other) PRs are not being ignored, the problem is simply that no one has really had the bandwidth to work on resvg (at least for me it's lack of time but also a bit lack of motivation), we've managed to pull of the harfrust migration because it's pretty important, but that's about it. This is why there are so many unmerged PRs.

But anyway, a second reason is that even though replacing panics with errors is a good idea in general and seems easy to do on the surface, it still requires a bit more consideration about whether we really want to make that change, and that requires digging into the codebase again which I've not had the time to do!. While in general removing panics is good, we need to distinguish between unwraps that can actually fail (like the filter ones) and unwraps which shouldn't be able to fail and therefore are fine keeping. For example:

`let target_size = tiny_skia::IntSize::from_wh(pixmap.width(), pixmap.height())?;`

I don't think there is any point in removing the unwrap here because the pixmap size itself is already guaranteed to be non-zero IIRC, so this should never fail. For the filter case, I'm still unsure whether it's really worth it to make rendering fallible because of this single case, because other than that I don't think there are that many fallible operations in `resvg`. Maybe it is, but maybe it would be better to just clamp instead (because honestly, no one should really be rendering to `u32::MAX`...). I hope that explains why it's not super straight-forward to merge these changes and why it's been taking so long.

I'll try to make more time for resvg in the near future, but of course I can't make promises, because it's just in my free time after all. :)

crates/resvg/tests/integration/main.rs
LaurenzV: It's fine, it's just for testing.

LaurenzV:
I've decided to just clamp now. If you are trying to rasterize SVGs with millions of pixel in a single dimension, I think you have other problems to worry about. 😄

LaurenzV:
#937 is not fixed by this, but this seems to be a bug in tiny-skia itself with very large paths.
#941 will run into all eternity now, but at least not panic.

The correct fix for both will be to introduce a fallible API and allow restricting the maximum dimensions, but that's a whole other can of worms we don't need to address in this PR now. Apart from that,  users can also detect the SVG size themselves and not even start rendering if they want to.

So I think this is good enough for now, thanks!

Its-Just-Nans:
> The correct fix for both will be to introduce a fallible API and allow restricting the maximum dimensions, but that's a whole other can of worms we don't need to address in this PR now.

On my side, I'm also interested in the `render()` function to return a `Result<()>` so that the user knows if there was a problem or not.

I think this could give the lib user some error and a way to know if the render failed ^^

Thanks

Issue #853, opened 2024-12-06
Thread 'main' panicked


yanschor:
While processing the attached SVG file the process fails with the following error message:

```bash
Warning (in usvg::parser::filter:93): Failed to parse a filter value cause expected '(' not '^' at position 2. Skipping.
thread 'main' panicked at crates/resvg/src/main.rs:735:78:
called 'Option::unwrap()' on a 'None' value
```

Our team discovered the issue during a current research project focusing on the schema-based generation of test-data for different XML specifications. 

### Steps to Reproduce

1. Execute `resvg <filepath> test.png`

### Environment:

• OS: macOS 15.1.1

• reSVG version: 0.44.0

### SVG File:
```xml
<?xml version="1.0" encoding="UTF-16"?>
<tg:svg xmlns:tg="http://www.w3.org/2000/svg" xmlns:xs="http://www.w3.org/2001/XMLSchema"
    xmlns:xml="http://www.w3.org/XML/1998/namespace" xmlns:xlink="http://www.w3.org/1999/xlink"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" id="id0" xml:base="Dw" class=" &#167;"
    color="navy" color-interpolation="auto" color-rendering="optimizeQuality" fill="	"
    stroke-dasharray="" font-family="&#xEEe2;&#xeE95;&#x18093;" font-size-adjust="none"
    font-stretch="extra-expanded" font-variant="normal" cursor="s-resize" display="compact"
    filter="B^" pointer-events="all" shape-rendering="auto" visibility="inherit" color-profile=""
    marker-mid="#" alignment-baseline="inherit" direction="inherit"
    glyph-orientation-vertical="+31deg" text-anchor="middle" text-decoration="none"
    unicode-bidi="inherit" word-spacing="exact" writing-mode="rl" overflow="hidden" onfocusin=""
    onclick="&#144;&#xFedE;&#xE66C;


" onmousedown="
" onmouseup="&#x10EF5A;" onmouseover=""
    onload="" onunload="" onabort="
" y="694e2" width="6e8" height="01.e+8"
    contentScriptType="multipartᘢ𩡩ﬤ𤤽𦏮悁" contentStyleType="image">
    <tg:a id="id1" xml:base="^d" xlink:role="&lt;&#38;" xlink:title="" xlink:show="replace"
        xlink:href="z" requiredExtensions="&#xA52Fc;
" externalResourcesRequired="false"
[...]

PR #990, opened 2025-12-14
Avoid unwraps on pixmap creation

Its-Just-Nans:
Fix https://github.com/linebender/resvg/issues/853

### DIFF
--- a/crates/resvg/src/main.rs
+++ b/crates/resvg/src/main.rs
@@ -675,16 +675,16 @@ fn render_svg(args: &Args, tree: &usvg::Tree) -> Result<tiny_skia::Pixmap, Strin
             None => return Err(format!("SVG doesn't have '{}' ID", id)),
         };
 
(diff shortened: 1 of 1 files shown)


crates/resvg/src/main.rs
```diff
@@ -703,10 +703,11 @@ fn render_svg(args: &Args, tree: &usvg::Tree) -> Result<tiny_skia::Pixmap, Strin
             let size = args
                 .fit_to
                 .fit_to_size(tree.size().to_int_size())
-                .ok_or_else(|| "target size is zero".to_string())?;
+                .ok_or("target size is zero")?;
 
             // Unwrap is safe, because `size` is already valid.
```
LaurenzV: Does that mean the comment is incorrect? If so, why? We should remove it then.

Its-Just-Nans: Pixmap's width is limited by i32::MAX/4, we handle the creation error.

LaurenzV:
Thanks!

PR #992, opened 2025-12-19
feat: Parse `light-dark()` CSS function by extracting the first argument and add tests.

### LINKED ISSUE #900: FR: Support parsing style with `light-dark()`
This PR asks for supporting [`light-dark()`](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/light-dark), a newly available feature, in style.

### Background and Motivation

`light-dark()` enables setting two colors for an element. For example, draw.io generates the following SVG segment:

```svg
<g>
  <rect
    x="120"
    y="0"
    width="120"
    height="60"
    fill="#ffffff"
    stroke="#000000"
    pointer-events="all"
    style="fill: light-dark(#ffffff, var(--ge-dark-color, #121212)); stroke: light-dark(rgb(0, 0, 0), rgb(255, 255, 255));"/>
</g>
```

When using this kind of SVG in downstream crates (e.g., typst), usvg fails to parse them and falls back to black:

```console
> resvg test.drawio.svg test.png
Warning (in usvg::parser::style:140): Failed to parse fill value: 'light-dark(#ffffff, var(--ge-dark-color, #121212)'. Fallback to black.
```

It would be nice to support parsing `light-dark()`.

### Discussion

It's weird for me to have two colors for an SVG that will be converted to PNG or PDF. Thus, my first thought would be to ignore the color of the dark theme and only use the color of the light theme. Since downstream users may use this feature to decide how to render SVG themselves, providing both color codes in `svgtypes::Paint` may be helpful.

tacshi:
Closes [#900](https://github.com/linebender/resvg/issues/900)

Added support for the CSS `light-dark()` function to handle SVGs exported from Draw.io with dark mode support. Previously, usvg failed to parse `light-dark()` and fell back to black, making text invisible on dark backgrounds.

Already tested in Zed

### DIFF
--- a/crates/resvg/src/main.rs
+++ b/crates/resvg/src/main.rs
@@ -206,6 +206,8 @@ OPTIONS:
   --export-area-drawing         Use drawing's tight bounding box instead of image size.
                                 Used during normal rendering and not during --export-id
 
+  --color-scheme SCHEME         Sets the color scheme for resolving CSS light-dark() function
+                                [default: light] [possible values: light, dark]
   --perf                        Prints performance stats
   --quiet                       Disables warnings
 
@@ -240,6 +242,7 @@ struct CliArgs {
     skip_system_fonts: bool,
     list_fonts: bool,
     style_sheet: Option<path::PathBuf>,
+    color_scheme: usvg::ColorScheme,
 
     query_all: bool,
     export_id: Option<String>,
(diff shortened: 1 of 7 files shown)


LaurenzV:
Thanks for the PR. However, while this might work for those tests, I'm not sure whether this is the right approach unfortunately. Firstly, I think it's a bit too hacky that we resolve `light-dark` by building a new string that replaces it with the evaluated value (if I'm understanding it correctly). I also don't think that we should try to parse `light-dark` recursively, as this doesn't make a whole lot of sense (but the problem here is more generally that we don't have a generalized mechanism for handling functions).

If I'm understanding correctly, we are also attempting to parse `light_dark` for every single declaration, which seems a bit wasteful to me, even if it probably won't have a noticeable performance impact.

I think the proper approach would be to support parsing light-dark inside of `svgtypes` when resolving paints, since `light-dark` essentially needs to be treated like a color. However, I need to look into it more when I find some time. Other suggestions are welcome.

Issue #993, opened 2025-12-21
[usvg] Failed to parse `marker-start` value `none`


sww1235:
When parsing a SVG file with `style="marker:none"` used, I get the following warnings emitted. According to the w3.org validator, this is valid SVG. 

Unclear if there is an issue with usvg here, or if usvg is holding SVG files to an even tighter standard than w3.org.

If I remove `style="marker:none" then the errors go away.

This is not something that is critical for my usecase as I can remove the `style` attributes but figured you would want to know this error.

Thanks for all the work on this library!

```
2025-12-21T18:40:14.438Z WARN  [usvg::parser::svgtree] Failed to parse marker-start value: 'none'.
2025-12-21T18:40:14.438Z WARN  [usvg::parser::svgtree] Failed to parse marker-mid value: 'none'.
2025-12-21T18:40:14.438Z WARN  [usvg::parser::svgtree] Failed to parse marker-end value: 'none'.
```

```svg
<svg xmlns="http://www.w3.org/2000/svg" version="1.0" viewBox="0 0 347 100">
  <g fill-rule="evenodd" stroke="#000" stroke-width="5">
    <path d="M140 66a19 19 0 1 1-38 0 19 19 0 1 1 38 0z" fill="#fff" />
    <path d="M238 221h93m-315 0h93m21-58 94 61" fill="none" stroke-linejoin="round" color="#000" overflow="visible" style="marker:none" transform="translate(-9 -155)" />
    <path d="M228 66a19 19 0 1 1-38 0 19 19 0 1 1 38 0z" fill="#fff" />
  </g>
</svg>
```

This is a file downloaded from [wikipedia](https://upload.wikimedia.org/wikipedia/commons/4/46/SPST-Switch.svg)

PR #994, opened 2025-12-21
Don't emit warning when encountering link with `none`

Its-Just-Nans:
Potential fix for https://github.com/linebender/resvg/issues/993


sww1235:
Thanks for fixing this so quickly!

PR #1000, opened 2026-01-09
feat: Add font hinting support via skrifa

oetiker:
Implement optional font hinting using Google's skrifa library for better
text rendering at small sizes. Hinting grid-fits glyph outlines to pixel
boundaries, improving legibility.

Features:
- New `hinting` cargo feature flag (requires `text` feature)
- When compiled with hinting, enabled by default (matching browser behavior)
- CSS text-rendering property controls per-element hinting:
  - optimizeLegibility/optimizeSpeed: hinting enabled
  - geometricPrecision: hinting disabled (preserve exact outlines)
- Works with variable fonts
- Hinting options propagated to nested/embedded SVGs

Implementation:
- SkrifaPen adapter converts skrifa OutlinePen to tiny_skia_path
- Hinted outlines drawn at ppem scale, then scaled back to font units
  to work with existing transform pipeline
- ppem calculation: font_size * dpi / 72.0

Tests:
- 7 new integration tests demonstrating hinting effects
- Example script for generating comparison images
- Demo SVG showing hinted vs unhinted text at various sizes


RazrFalcon:
I agree. #922 should be merged first.

PR #991, opened 2025-12-17
Change some assertions to debug assertions

Its-Just-Nans:
Fix https://github.com/linebender/resvg/issues/934

But it's looks like a "breaking" change since arithmetic() is public

Issue #815, opened 2024-09-02
Add memory limits


RazrFalcon:
Add an ability to limit memory usage somehow. Or at least limit maximum `Pixmap`s size.

LaurenzV:
I think in most cases you can just check what the size of the SVG is. If the SVG contains filters, things might be a bit trickier.

thekingofcity:
Or is it possible to render the image part by part? Maybe cut into smaller tiles?

RazrFalcon:
Not really possible with SVG.

Its-Just-Nans:
> But it would mean that rendering can "fail" now.

I think the rendering should be able to fails safely (no panic), and the result of `render()` would be
- `Result<Option<Warning>, Error>`
or
- `Result<Option<Vec<Warning>>, Error>`

```rust
// pseudo code
match render(...) {
  Ok(None) => //perfect
  Ok(Some(..)) => // best effort
  Err(..) => // real error
}
```

Its-Just-Nans:
> Returning warnings is basically impossible and would require a significant rewrite.

Okay yes. and technically we can catch logs with a custom logger
 
> Returning an error does not make sense, because the only error is allocation.

yes error in resvg is allocation related but just the fact that "there was an error during render" is useful to know for library users.
For example it allows the library user to add a message like "error during render, the render may be incorrect" or something like that

PR #1008, opened 2026-01-13
fix: change `assert!()` to `Result` with `Error`

### LINKED ISSUE #1007: Specific SVG input file causing panic
Unexpected SVG input can cause the program to panic due to this assert: https://github.com/linebender/resvg/blob/b8e58f5ae69d5ee5033e31549e5dd8310b836a56/crates/resvg/src/filter/lighting.rs#L140

I believe this is a bug as a library should not panic depending on user-dependent input data.

Steps to reproduce:
- Install https://flathub.org/en/apps/io.github.davidoc26.wallpaper_selector
- Feed its icon into resvg

Code to reproduce panic (just `cargo add resvg` + `cargo run`)

```rust
fn main() {
    let width = 96;
    let height = 96;
    let file_path = "/var/lib/flatpak/exports/share/icons/hicolor/scalable/apps/io.github.davidoc26.wallpaper_selector.svg";

    let options = resvg::usvg::Options {
        style_sheet: Some("svg { color: white }".into()),
        ..Default::default()
    };

    let data = std::fs::read(file_path).unwrap();

    let tree = resvg::usvg::Tree::from_data(data.as_slice(), &options).unwrap();
    let svg_size = tree.size();
    let scale_x = (width as f32) / svg_size.width();
    let scale_y = (height as f32) / svg_size.height();

    let mut pixmap = resvg::tiny_skia::Pixmap::new(width as _, height as _).unwrap();
    let transform = resvg::usvg::Transform::from_scale(scale_x, scale_y);
    resvg::render(&tree, transform, &mut pixmap.as_mut());
}
```

Its-Just-Nans:
fix #1007

The CLI will produce 
```txt
Warning (in resvg::filter:351): Filter has an invalid region.
```

### DIFF
--- a/crates/resvg/src/filter/composite.rs
+++ b/crates/resvg/src/filter/composite.rs
@@ -9,10 +9,6 @@ use usvg::ApproxZeroUlps;
 ///
 /// - `src1` and `src2` image pixels should have a **premultiplied alpha**.
 /// - `dest` image pixels will have a **premultiplied alpha**.
-///
-/// # Panics
-///
-/// When `src1`, `src2` and `dest` have different sizes.
 pub fn arithmetic(
     k1: f32,
     k2: f32,
--- a/crates/resvg/src/filter/composite.rs
+++ b/crates/resvg/src/filter/composite.rs
@@ -21,9 +21,13 @@ pub fn arithmetic(
     src1: ImageRef,
     src2: ImageRef,
     dest: ImageRefMut,
-) {
-    assert!(src1.width == src2.width && src1.width == dest.width);
-    assert!(src1.height == src2.height && src1.height == dest.height);
+) -> Result<(), crate::filter::Error> {
+    if src1.width != src2.width || src1.width != dest.width {
+        return Err(crate::filter::Error::InvalidRegion);
+    }
+    if src1.height != src2.height || src1.height != dest.height {
+        return Err(crate::filter::Error::InvalidRegion);
+    }
 
     let calc = |i1, i2, max| {
         let i1 = i1 as f32 / 255.0;
@@ -49,4 +53,5 @@ pub fn arithmetic(
 
         i += 1;
     }
+    Ok(())
 }
--- a/crates/resvg/src/filter/composite.rs
+++ b/crates/resvg/src/filter/composite.rs
@@ -1,7 +1,7 @@
 // Copyright 2020 the Resvg Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-use super::{ImageRef, ImageRefMut, f32_bound};
+use super::{Error, ImageRef, ImageRefMut, f32_bound};
 use rgb::RGBA8;
 use usvg::ApproxZeroUlps;
 
@@ -17,12 +17,12 @@ pub fn arithmetic(
     src1: ImageRef,
     src2: ImageRef,
     dest: ImageRefMut,
-) -> Result<(), crate::filter::Error> {
+) -> Result<(), Error> {
     if src1.width != src2.width || src1.width != dest.width {
-        return Err(crate::filter::Error::InvalidRegion);
+        return Err(Error::InvalidRegion);
     }
     if src1.height != src2.height || src1.height != dest.height {
-        return Err(crate::filter::Error::InvalidRegion);
+        return Err(Error::InvalidRegion);
     }
 
     let calc = |i1, i2, max| {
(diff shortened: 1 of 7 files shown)


galister:
Thanks for jumping on this so quickly!

I noticed that the the comments also have the following, which can be removed:

```rust
/// # Panics
///
/// When `src1`, `src2` and `dest` have different sizes.
```

galister:
In case maintaining API compatibility is important, it would be viable to have 2 variants of the fn:

`arithmetic_checked` return Result, while the compat fn `arithmetic` can just call `arithmetic_checked` and unwrap the result.

That being said, I can't imagine a use-case where resvg is used as a library, and the developer would want it to panic depending on a faulty input file.

galister:
i have submitted an SVG that triggers this in https://github.com/linebender/resvg/issues/1007

Its-Just-Nans:
> If there is a panic in filters code it means we have parsed the SVG wrong.

I agree that fixing bugs to have a correct parser and rendered is the good thing to do.

But penalizing crates which uses resvg by allowing it to panic is not great for a library
And **I would really be happy** to remove all unwrap and panics (some of my MRs are already going this way #989 #991 and this one).

---

I think the problem #1007 is not from the parsing but from this part

```rust
// crates/resvg/src/lib.rs
let target_size = tiny_skia::IntSize::from_wh(pixmap.width(), pixmap.height()).unwrap();
let max_bbox = tiny_skia::IntRect::from_xywh(
    -(target_size.width() as i32) * 2,
    -(target_size.height() as i32) * 2,
    target_size.width() * 5,
    target_size.height() * 5,
)
.unwrap();
```

If I increase the `max_bbox`, the render works.
Any reason for the chosen value `*2` and `*5` ?

Its-Just-Nans:
> As for 2 and 5 - this is to limit the filter region to 5x the size. The limit is arbitrary.


Shouldn't this limit be set by the user using options args/params ? (kinda linked to https://github.com/linebender/resvg/issues/815)

For example, if I change the size like in my latest commits, the render is working


<details>

<summary>
`main()` code

</summary>

```rust
fn main() {
        if let Ok(()) = log::set_logger(&LOGGER) {
        log::set_max_level(log::LevelFilter::Warn);
    }
    let width = 96;
    let height = 96;
    let file_path = "534586290-631255f5-efc4-40bf-966e-69d08a6d3d48.svg"; // downloaded from the issue

    let options = resvg::usvg::Options {
        style_sheet: Some("svg { color: white }".into()),
        ..Default::default()
    };

    let data = std::fs::read(file_path).unwrap();

    let tree = resvg::usvg::Tree::from_data(data.as_slice(), &options).unwrap();
    let svg_size = tree.size();
    let scale_x = (width as f32) / svg_size.width();
    let scale_y = (height as f32) / svg_size.height();

    println!("{width} {height}");
    let mut pixmap = resvg::tiny_skia::Pixmap::new(width as _, height as _).unwrap();
    let transform = resvg::usvg::Transform::from_scale(scale_x, scale_y);
    resvg::render(&tree, transform, &mut pixmap.as_mut());
    let buf = pixmap.encode_png().map_err(|e| e.to_string()).unwrap();
    std::fs::write("a.png", &buf).unwrap();
}
```

</details>

RazrFalcon:
It's such a rare edge case that it's not important.

yisibl:
A panic error was also reported here: https://github.com/servo/servo/issues/42258

galister:
Hi, I would like to revive the conversation about this since there are multiple downstream projects that need this to not panic.

As far as I understand:
- there is an arbitrary limit for the filter region size that is defined in code
- no matter how large the limit, a potentially unsound SVG could still exceed that and cause a panic
- we can't really remove the limit or make it depend on the SVG contents; an unsound SVG could end up consuming a disproportionate amount of resources

If this is the case, I believe the only way forward would be to change the signature to return results instead.

LaurenzV:
#991 and #989 have been merged with some tweaks now, see the comments there.

As far as I can tell, there still are a few ways to trigger panics with filters, but the correct fix here is to fix the underlying bug instead of removing the assertions (now debug assertions) that guard the invariants.

> If this is the case, I believe the only way forward would be to change the signature to return results instead.

We can fix the panics, but I don't think you will ever get to a point where you can render with resvg without potentially running into a resource exhaustion problem. There's just too many ways in which you could craft an SVG to make usvg/resvg run endelessly, even with exhaustive checks.

Given the above, I think this PR is stale now. Thanks again!

PR #1011, opened 2026-01-28
feat(usvg): add warning when parsing malformed paths

Its-Just-Nans:
Linked to https://github.com/linebender/resvg/issues/938

```sh
$ cargo run --package usvg a.svg b.svg

Warning (in usvg::parser::shapes:33): Error during path parsing: invalid number at position 9
```


LaurenzV:
I've removed the warning from transform parsing, because we already print warnings for nodes attributes that fail to parse, so it's being printed twice:

https://github.com/linebender/resvg/blob/67c2c65a590d62a29facd63d08e305a8db48b907/crates/usvg/src/parser/svgtree/mod.rs#L302-L308

But the one for paths looks fine to me (those are just parsed as raw strings so they aren't detected earlier.

LaurenzV:
Thanks!

Issue #890, opened 2025-02-27
Font rendering with vertical text


sugi2000:
It is great to accept vertical writing-mode.
I tried Japanse vertical text.
However, some characters are rotated wrongly.

left: SVG on browser
right: svg2png-wasm using resvg
![Image](https://github.com/user-attachments/assets/ba77a2b1-d809-4fd4-be34-7c7e8e844b9a)

sugi2000:
Here is the text element.

```svg
<text style="writing-mode: vertical-rl;">（すーっ）</text>
```

RazrFalcon:
This character is marked as _Transformed typographically, with fallback to Upright_ in [unicode-vo](https://github.com/RazrFalcon/unicode-vo/tree/master). But we handle only _Upright_ in `usvg`.
Therefore I assume [this code](https://github.com/linebender/resvg/blob/7b5464acdb42fc37d1c681800ab52b7ee90d5195/crates/usvg/src/text/layout.rs#L1016) should be adjusted.

sugi2000:
Thank you for pointing this out. I am wondering some characters like the following are rotated correctly:

- U+FF08 FULLWIDTH LEFT PARENTHESIS
- U+30FC KATAKANA-HIRAGANA PROLONGED SOUND MARK

https://www.unicode.org/reports/tr50/tr50-19.html

sugi2000:
I saw your unicode-vo [code](https://github.com/RazrFalcon/unicode-vo/blob/master/src/lib.rs). What is the difference between `Orientation::TransformedOrUpright` and `Orientation::Upright`?

RazrFalcon:
>What is the difference between Orientation::TransformedOrUpright and Orientation::Upright?

I have absolutely no idea...

sugi2000:
I understood the way the current code does it. It just rotates upright characters 90 degrees. But we should change the glyph for vertical. The font has the information, but I don't know how to extract it.

RazrFalcon:
>The Japanese fonts have the vertical glyphs, but I don't know how to extract them.

Are you sure? Fonts can have vertical layout specific metrics, but not separate glyphs.
So the current code is correct, we just have to figure out how to improve it.

sugi2000:
Yes, I'm sure.

[...]

PR #1013, opened 2026-02-01
feat: add support for vertical writing mode

KewenCode:
1. **Add support for alternative glyphs in vertical text mode when the font supports the 'vert' attribute.**
    - Base on document [UAX#50](https://unicode.org/reports/tr50).
    - Add two test for vertical text.
      |Test|Chrome-Origin|Chome-Current|
      |:--|:-:|:-:|
      |symbol-text|symbol-text-chrome|symbol-text-resvg|
      |text-cjk|text-cjk-chaome|text-cjk-resvg|
      -  In 'text-cjk', after applying feature `vhal` to the character `Ấ`, the `harfbuzz` rendering differs from that of Chrome, resulting in reduced spacing.
      ---
    - **Finish #890**
      - The direction of the character 'っ' (TR) was incorrect, and it has now been corrected to meet the relevant requirements of UAX#50.
      ---
    - Two tests underwent changes due to improved functionality
      |text\textPath\complex|text\writing-mode\tb-and-punctuation|
      |---|---|
      |tests_text_textPath_complex|tests_text_writing-mode_tb-and-punctuation|
    ---
2. **Add font to support all CJK text.**
    - Font `SourceHanSansHWSC-Regular.subset.ttf` to support tests.
    - Due to the presence of CJK text in the four tests, the fallback font has been changed to new font.
      |Location|Diff|Location|Diff|
      |:--|:-:|:--|:-:|
      |text\text\xml-lang=ja|tests_text_text_xml-lang=ja|text\writing-mode\tb-with-rotate|tests_text_writing-mode_tb-with-rotate|
      |text\writing-mode\mixed-languages-with-tb-and-underline|tests_text_writing-mode_mixed-languages-with-tb-and-underline|text\writing-mode\tb-with-rotate-and-underline|tests_text_writing-mode_tb-with-rotate-and-underline|
    ---
4. **Add two OpenType features (`vhal` `halt`) to accommodate CJK text.**


KewenCode:
> Wow! That's a serious change. I'm no Unicode/Text Layout expert, so I would assume it's better. As long as it does not break existing test - we're fine.
> 
> Does it actually fixes #890 ?
> 
> Also, 16MB for SourceHanSansHWSC-Bold.otf‎ is too much. Getting it below 1MB would be nice. Can you try subsetting it as mentioned [here](https://github.com/linebender/resvg/blob/main/crates/resvg/tests/fonts/README.md)?
> 
> Also, can you post new test file results before and the patch? Just for the reference. I want to see how exactly thing have changed. Thanks!

Sorry to keep you waiting. The setup of fonttools has puzzled me for quite a while。
And thank you for your suggestion. I have optimized the font file and added a comparison image.

crates/usvg/src/text/layout.rs
```diff
@@ -1477,14 +1599,51 @@ fn shape_text_with_font(
...
+                    )
+                };
 
-            let positions = output.glyph_positions();
-            let infos = output.glyph_infos();
+            let mut pre_pos = rustybuzz::GlyphPosition::default();
+            let mut vertical_sub = 0;
 
-            for i in 0..output.len() {
+            for i in 0..output_horizontal.len() {
```
StefanoD, line 1642: Some concerns by Claude: Ignore, if this is noise:

**Cross-indexing the two shaping passes can panic / mismatch glyphs.**
The loop is bounded by `output_horizontal.len()`, but in vertical mode `positions`/`infos` come from `output_vertical` (and `sub_positions`/`sub_infos` are the other pass). This assumes both passes return the same glyph count *and* a 1:1 positional correspondence. rustybuzz/HarfBuzz guarantees neither across different directions and feature sets (`vert`/`vrtr`/`vhal` substitutions, ligatures, cluster regrouping). If the other pass yields fewer glyphs, `positions[i]` / `sub_positions[i]` is an out-of-bounds panic — and because `sub_*` is indexed unconditionally, this can fire on **horizontal** text too, not just vertical.

Suggest bounding by `min(len)` (or `.get(i)`) and, better, looking up the backup glyph by cluster index rather than positional `i`. A property test over mixed Latin+CJK+ligature runs would cover this.

Related: `sub_text[start..end]` below slices using `start = info.cluster` from one pass and `end` from another — if cluster order isn't monotonic for the chosen pass, `start > end` panics. Worth a guard.

crates/usvg/src/text/layout.rs
```diff
@@ -1477,14 +1599,51 @@ fn shape_text_with_font(
...
+            }
+
+            // Note: rustybuzz automatically applies 'vert' and 'vkrn' features for vertical text
+            // 'vert' - glyphs to be presented upright
+            // 'vrtr' - glyphs to be presented sideways (rotate 90°)
+            // Use 'vrtr' when text-orientation: 'text-orientation' is required
+
+            // Perform text shaping for both orientations
+            let output_horizontal = rustybuzz::shape(&rb_font, &features, buffer_horizontal);
+            let output_vertical = rustybuzz::shape(&rb_font, &features, buffer_vertical);
```
StefanoD, line 1619: Some concerns by Claude: Ignore, if this is noise:

**Both shaping passes run unconditionally — perf regression on the common path.**
`output_vertical` is always computed, but in horizontal mode it's never read (the `if writing_mode == TopToBottom` branches that consume `sub_*` are skipped). This roughly doubles shaping cost for all existing horizontal text, and this function is on the hot path for every `<text>` in every SVG. Please gate the second pass (and its buffer) behind `writing_mode == WritingMode::TopToBottom`.

crates/usvg/src/text/layout.rs
```diff
@@ -1039,31 +1043,53 @@ fn apply_length_adjust(chunk: &TextChunk, clusters: &mut [GlyphCluster]) {
...
-            cluster.descent = -cluster.width / 2.0;
-        } else {
+            let half_width = cluster.width / 2.0;
+            cluster.ascent = half_width;
+            cluster.descent = -half_width;
+        } else if matches!(
+            orientation,
+            // Characters which are displayed sideways, rotated 90 degrees clockwise compared to the code charts.
+            unicode_vo::Orientation::TransformedOrRotated | _
+        ) {
```
StefanoD, line 1092: Some concerns by Claude: Ignore, if this is noise:

**This `matches!` is always true — the `else if` is just an `else`.**
The `| _` wildcard matches every variant, so naming `TransformedOrRotated` is misleading and it trips Clippy's `wildcard_in_or_patterns`. Make it a plain `else`, or restrict to the orientation(s) you actually mean.

```suggestion
        } else {
```

crates/usvg/src/text/layout.rs
```diff
@@ -1175,7 +1230,21 @@ fn form_glyph_clusters(glyphs: &[Glyph], text: &str, font_size: f32) -> GlyphClu
...
+        // GlyphCluster.width is visual width in logical coordinates
+        let glyph_width = if writing_mode == WritingMode::TopToBottom {
+            // In vertical writing mode:
+            // - Rotated glyphs (Latin, Cyrillic, etc.) use their original width (set in fn `shape_text_with_font`).
+            //   (will be rotated 90 degrees)
+            // - Upright glyphs (CJK, etc.) use vertical advance (dy * 2)
+            if glyph.text_orientation == unicode_vo::Orientation::Rotated {
+                glyph.width as f32 * sx // Use original width for rotated characters
+            } else {
+                glyph.dx.abs() as f32 * 2.0 * sx // Vertical advance for upright characters
```
StefanoD, line 1242: Some concerns by Claude: Ignore, if this is noise:

**Comment/code mismatch + undocumented assumption.**
The comment says "vertical advance (dy * 2)" but the code uses `dx`. This looks like it's reconstructing the advance from HarfBuzz's vertical-centering x-offset (≈ −width/2) — clever but brittle and undocumented. Please fix the comment to match, and explain why `dx*2` recovers the advance (or read the actual vertical advance directly).

crates/usvg/src/text/layout.rs
```diff
@@ -1141,24 +1167,53 @@ fn apply_word_spacing(chunk: &TextChunk, clusters: &mut [GlyphCluster]) {
...
+    // Whole glyphs should in the same orientation.
+    let mut glyphs_orientation = unicode_vo::Orientation::Rotated;
+    let mut glyphs_replaced = false;
+
     let mut positioned_glyphs = vec![];
 
     for glyph in glyphs {
         let sx = glyph.font.scale(font_size);
 
+        glyphs_orientation = glyph.text_orientation;
```
StefanoD, line 1190: Some concerns by Claude: Ignore, if this is noise:

**Cluster orientation reflects only the last glyph.**
`glyphs_orientation`/`glyphs_replaced` are overwritten each iteration, so a multi-glyph cluster ends up with whatever the final glyph had. The comment states the "same orientation" assumption but nothing enforces it — a mixed cluster would silently mis-rotate. Consider asserting it (debug) or deriving from the first glyph deliberately.

crates/resvg/tests/tests/text/text/xml-lang=ja.svg
StefanoD, line 1: Some concerns by Claude: Ignore, if this is noise:

**This swap may defeat the test's purpose.**
This test (per the linked "your code displays Japanese wrong") exists to show that `xml:lang=ja` vs `zh-HANT` select *different* glyphs via language-aware fallback. Pinning a single subset font risks the three lines rendering identically, removing the behavior under test. Same concern applies to the `Amiri`→`Source Han Sans HW SC` swaps in `mixed-languages-with-tb-and-underline.svg` and `tb-with-rotate*.svg`, where the text still contains Arabic. If the goal is just deterministic CI rendering, prefer keeping the original fonts and adding *new* vertical-mode tests — or explain in the PR why each swap preserves the original intent.

Also: these files dropped their trailing newline and were fully re-indented (4→6 spaces, attributes rejoined), which buries the one real change. Please revert the reformatting and change only `font-family`.

PR #1014, opened 2026-02-03
Add support for 2 point conical radial gradient

wmedrano:
- Use `fr` from SVG radial gradient
  - Requires upgrading to tiny-skia 0.12 which just added support for focal radius.
- Diffs
  - Gradient changes are expected
  - The rest are tiny diffs due to subtle changes within tiny skia

### DIFF
--- a/crates/usvg/src/tree/mod.rs
+++ b/crates/usvg/src/tree/mod.rs
@@ -391,7 +391,7 @@ impl RadialGradient {
         self.fy
     }
 
-    /// Focal ragius.
+    /// Focal radius.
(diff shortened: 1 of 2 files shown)


crates/usvg/src/tree/mod.rs
```diff
@@ -389,6 +390,11 @@ impl RadialGradient {
     pub fn fy(&self) -> f32 {
         self.fy
     }
+
+    /// Focal ragius.
```
LaurenzV:
```suggestion
    /// Focal radius.
```

Issue #1006, opened 2026-01-12
Vendored source code of version 0.46.0 fails to build


igormironchik:
```bash
igor@localhost:~/Work/Tmp/resvg-0.46.0/crates/c-api> cargo build --release
warning: `/home/igor/Work/Tmp/resvg-0.46.0/.cargo/config` is deprecated in favor of `config.toml`
  |
  = help: if you need to support cargo 1.38 or earlier, you can symlink `config` to `config.toml`
error: failed to calculate checksum of: /home/igor/Work/Tmp/resvg-0.46.0/vendor/bitflags-1.3.2/tests/compile-fail/.gitignore

Caused by:
  failed to open file `/home/igor/Work/Tmp/resvg-0.46.0/vendor/bitflags-1.3.2/tests/compile-fail/.gitignore`

Caused by:
  No such file or directory (os error 2)
```

This is from https://github.com/linebender/resvg/releases/download/v0.46.0/resvg-0.46.0.tar.xz

Its-Just-Nans:
Warning is caused by

https://github.com/linebender/resvg/blob/b8e58f5ae69d5ee5033e31549e5dd8310b836a56/.github/workflows/tagged-release.yml#L71

Error is the generated vendor/bitflags-1.3.2/.cargo-checksum.json contains link to non existing file

Its-Just-Nans:
> [@Its-Just-Nans](https://github.com/Its-Just-Nans) I'm worried about error, will it be fixed in the archive on release page?

Hi, since I'm just a contributor I can't give an answer, but it would be nice if fixed inside the `Releases` (I think we can edit assets of a Github release)

nicoburns:
Hopefully this is fixed in 0.47.0

igormironchik:
Hi. Want to say that this was fixed in 0.47.0. Thanks.

PR #1017, opened 2026-02-05
Don't exclude gitignore file from tarballs

nicoburns:
I think this ought to fix https://github.com/linebender/resvg/issues/1006 but untested


LaurenzV (approved):
Let's hope for the best!

PR #1020, opened 2026-02-13
Support `path` attribute on `<textPath>` element

### LINKED ISSUE #417: Support the `path` property of `textPath`
https://www.w3.org/TR/SVG2/text.html#TextPathElementPathAttribute

yisibl:
Currently, there is a conflict within the SVG2 specification regarding:

> If the ‘[path](https://www.w3.org/TR/SVG2/text.html#TextPathElementPathAttribute)’ attribute contains an error, the ‘[href](https://www.w3.org/TR/SVG2/text.html#TextPathElementHrefAttribute)’ attribute must be used.
[w3.org/TR/SVG2/text.html#TextPathElement:~:text=If the ‘path’ attribute contains an error](https://www.w3.org/TR/SVG2/text.html#TextPathElement:~:text=If%20the%20%E2%80%98path%E2%80%99%20attribute%20contains%20an%20error)

Both @AmeliaBR and @karlcow believe this should be removed from the specification https://github.com/w3c/svgwg/issues/393#issuecomment-374456360, which is what the current PR does. This also maintains interoperability with browser implementations. 

I added a test case for this purpose: [textPath/with-error-path.svg](https://github.com/linebender/resvg/pull/1020/changes#diff-ab0a8dbf3d8f090d13bfbf5f7ec7e79469bc2096e0cebfefe6a671e93f3481dc)

See also:
- https://chromium-review.googlesource.com/c/chromium/src/+/7367700
- https://bugzilla.mozilla.org/show_bug.cgi?id=1446617

Fixes: https://github.com/linebender/resvg/issues/417


LaurenzV:
In Chrome I'm not seeing any text on most of those, is that intended? 
image

crates/resvg/tests/tests/text/textPath/with-error-path.svg
```diff
@@ -0,0 +1,19 @@
+<svg id="svg1" viewBox="0 0 200 200" xmlns="http://www.w3.org/2000/svg"
+     xmlns:xlink="http://www.w3.org/1999/xlink">
+    <title>With partially invalid `path` and `xlink:href` fallback (SVG 2)</title>
+    <desc>
+        If the `path` attribute contains any parse error, the `xlink:href` path must be used.
```
LaurenzV, line 5: But right now, the text is not using the `xlink:href` one, right?

LaurenzV: (Seems to be what Firefox and CHrome do, but then the comment itself confuses me.

crates/resvg/tests/tests/text/textPath/with-href-and-xlink-href.png
LaurenzV, line 1: This oen looks wrong?

LaurenzV (commented):
LGTM, just two comments.

yisibl:
Thank you for your review. I'm awaiting updates from SVGWG regarding their findings and tests.
- https://github.com/w3c/svgwg/issues/393#issuecomment-4085384503

I'll be back later.

yisibl:
Thank you for your review. I'm awaiting updates from SVGWG regarding their findings and tests.
- https://github.com/w3c/svgwg/issues/393#issuecomment-4085384503
- https://chromium-review.googlesource.com/c/chromium/src/+/7557652

I'll be back later.

yisibl:
Chrome currently does not support using [Basic shapes](https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorials/SVG_from_scratch/Basic_shapes) within `<textPath>`. 
You can test it using Firefox.
> In Chrome I'm not seeing any text on most of those, is that intended? image

luisbg:
The code looks right to me but the checked-in reference PNG is stale.

> ---- render::text_textPath_with_href_and_xlink_href stdout ----
> 
> thread 'render::text_textPath_with_href_and_xlink_href' (163722795) panicked at crates/resvg/tests/integration/render.rs:1651:55:
> assertion `left == right` failed
>  left: 5262
>  right: 0

The actual output is what the test SVG asks for, the text follows #path-expected, so href correctly wins over xlink:href. The expected PNG has it on #path-fallback, the rotated path, the behaviour this PR fixes.

> Regenerating gives a green suite (1741 passed, 0 failed):

Could you update the PR with that one regenerated PNG? Happy to be told I've got this backwards.

Issue #916, opened 2025-05-03
[Question] Resvg only use the first available font for Rendering


Kynson:
Hi, I am trying to render an svg using `text` containing both emoji and normal text. I loaded both `Fira Code` and `Noto Color Emoji` to the fontdb. The problem is if I only specify `font-family: "Fira Code, Noto Color Emoji"` then the emojis are not rendering. If I swap the order of the two then the font for normal text is not quite right (at least the ligatures are gone). 

May I know if this is the intended behaviour? Sorry if this is documented somewhere. Thanks

RazrFalcon:
This is an intended behavior. 

You have to understand that Noto Color Emoji probably has regular glyphs as well, not only emojis.
And the way `resvg` works is that it tries to "render" glyphs using "the first" font and fallbacks to other fonts it it wasn't possible.

The whole idea behind `font-family` fonts list is to provide fallback fonts in case one of the fonts is missing required glyphs. 

I would suggest not to mix fonts. Use `tspan` for each font-text pair. Like so:
```xml
<text>
<tspan font-family="Fira Code">regular text</tspan>
<tspan font-family="Noto Color Emoji">🙂</tspan>
</text>
```

Kynson:
Thanks for the reply. I guess I will have to detect whether the character is an emoji and switch font accordingly in this case. Just another question out of curiosity (I am not really familiar with font render, sorry if this is a dumb one):
If I do `font-family: "Fira Code", "Noto Color Emoji"` then I think Fira Code should not contains the glyphs for the emoji, but when I try to render the following svg, it is not falling back to Noto Color Emoji, but the replacement character is rendered. It renders as usual if I use `font-family:  "Noto Color Emoji", "Fira Code"`
```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 19 37" style="font-size: 16px; fill: #fff" font-family="Fira Code, Noto Color Emoji">
<rect width="100%" height="100%" style="fill: #fff" />
<text x="0%" y="33.33%" style="fill: #000">
😀
</text>
</svg>
```

RazrFalcon:
>but the replacement character is rendered

Which one? An empty rectangle or �?

But yes, in theory it should work just fine. Probably a bug.

Kynson:
"�", I believe:

![Image](https://github.com/user-attachments/assets/3bdc357e-210d-41c0-b62e-afadba40e64b)

Above generated using:
```rust
let mut db = Database::new();
[...]

Issue #583, opened 2023-02-20
hebrew fonts look weird, the font loads correctly but the positioning is off


isaacfink:
basically the title, the font is rendering ok but the positioning is completely off, here is how it looks when I open the svg in chrome
![image](https://user-images.githubusercontent.com/46644190/220127456-dc238303-16ac-4c27-8966-2019e7e823af.png)

and this is how it renders with resvg
![image](https://user-images.githubusercontent.com/46644190/220127510-c0b07084-a654-42a5-be84-95e37eeb0c42.png)
the relevant svg code is this
```
<text
    filter="url(#textShadow-uy58f0o68)"
      x="541"
      y="1012"
      font-family="BA Ziporen Plus Bold"
      font-size="38"
      fill="#b09554"
      alignment-baseline="hanging"
      text-anchor="middle"
      transform="rotate(0 541 1024.8)"
    >
      
          <tspan
            dy="0"
            x="541"
            text-anchor="middle"
            letter-spacing="0.2em"
            alignment-baseline="hanging"
          >
            מושי &amp;  חני
          </tspan>
          ,
          <tspan
            dy="0.9em"
            x="541"
            text-anchor="middle"
            letter-spacing="0.2em"
            alignment-baseline="hanging"
          >
            גרינבוים
          </tspan>
          
    </text>
```

isaacfink:
as i am looking at it now maybe the issue is the multiline, because it doesn't happen to other elements only to the multiline ones, so maybe the em syntax is not supported?

RazrFalcon:
Will take a look. You have a very complicated text element. Hard to say.
But we do support all those properties. Including `em` units.
It feels like `dy` is handled incorrectly here.

RazrFalcon:
Here is how your file is rendered in using other libraries.

![text-dy](https://user-images.githubusercontent.com/725494/220152319-966086d7-8ebe-4903-8d8b-b515b27b2908.png)

As you can see - it's a mess.

The first rule of SVG - never use text.

RazrFalcon:
Looks like I have to invert characters positions on RTL layout as well. No idea how to do this. RTL is hard.

For now, you can simply use `y` instead of `dy`.

RazrFalcon:
I would also suggest using separate `text` for each line. `tspan` support is extremely tricky and you will never get expected/reproducible results.

isaacfink:
[...]

PR #1036, opened 2026-03-08
Always fallback fonts using the font-family priority list

### LINKED ISSUE #301: (usvg) Prefer font-family during font fallback

bczhc:
The old font resolving approach is to scan the whole tspan text and try to find a font to match all the text. According to CSS `font-family` font resolving schema, it should resolve and fallback using the family name one by one, making font uses more ergonomic.

> Component values are a comma-separated list indicating alternatives. A user agent iterates through the list of family names until it matches an available font that contains a glyph for the character to be rendered.

This implements a simple resolving algorithm doing what it states above.

Fix #301, #916.

```svg
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<svg
        width="100.147198mm"
        height="32.596107mm"
        viewBox="0 0 57.147198 30.596108"
        version="1.1"
        id="svg1"
        xmlns="http://www.w3.org/2000/svg"
>
    <text
            xml:space="preserve"
            x="14.375127"
            id="text1"><tspan
       id="tspan1"
       style="font-family: 'Noto Sans', 'Noto Sans CJK SC', 'Noto Color Emoji';"
       x="14.375127"
       y="20.459293">Text文字😀</tspan></text>
</svg>
```

**Tests**

Before:

After:
new


bczhc:
The CI outputs:

```
---- render::text_font_family_fallback_3 stdout ----
Warning (in usvg::text:183): Fallback from Noto Sans to Noto Color Emoji.
Warning (in usvg::text:183): Fallback from Noto Serif to Noto Color Emoji.
```

On my machine it passes. Falling back to Twitter Color Emoji:

```
---- render::text_font_family_fallback_3 stdout ----
Warning (in usvg::text:183): Fallback from Noto Sans to Twitter Color Emoji.
Warning (in usvg::text:183): Fallback from Noto Serif to Twitter Color Emoji.

successes:
    render::text_font_family_fallback_3
```

Indeterministic font fallback causes the test failure. Honestly I have no idea why is that. Solution thoughts?

luisbg:
Why does this PR change `rtl.png`? It isn't clear to me.

Thank you.

bczhc:
Hmm because it fails. -- Ok though I've lost some context for this pr now but I may be sure it's because this PR that changes font resolving logic has effects on `rtl.svg` (this svg uses a comma-separated font list; before the font resolving logic change, resvg will only use the first font to render it, which is not correct).

Also btw, rtl.png rendered by resvg is not correct itself now. The expected result is:

image

because currently (at least at the time I made the pr) resvg does not have a _correct_ rtl handing.

luisbg:
True, #583 does track that RTL is mishandled.

The golde test image changing makes sense to me. The PR changes which font supplies which run, so `rtl.svg` does render differently. My only ask is that this gets written down in the PR description to make it clear.

luisbg:
@nicoburns and @DJMcNab ,  I think this PR might be hitting a CI/harness issue. It sometimes passes and sometimes not because of the iteration order of `std::fs::read_dir`.

I am going to open a PR to replace `load_fonts_dir` in the test harness with an explicit sorted load.

nicoburns:
@luisbg please can you stop tagging me in these. Tagging me does not give me any more time to actually review these PRs.

Actual review comments that explain what review steps you have performed (reviewing the implementation, comparing it with various specs, testing the code against SVGs, etc), and/or reasons for *why* you think a given PR is good/bad are helpful.

luisbg:
Harness fix at #1136 

What I checked to review this PR:
- Read the issue and the code fix to familiarize myself with things.
- The bug is real and still present. I applied the PR's test files alone to unmodified main and two tests fail.
- Applying the code changes fixes those two tests. Since there is a conflict I had to do this by hand.
- On `rtl.png`, I also confirmed renders are visibly wrong, most of the Arabic is missing either way, which is #583.
- I saw CI was not passing and I investigated why. Since it's outside the scope of this PR I fixed it in a new one. Note: a new golden image will need to be generated for this PR to pass the deterministic order.
- I had an LLM assist me on running tests and figuring out the CI issue.
- I did not read specs.

Overall I think this is worth landing: it fixes #301  (open since 2020) and #916 , from what I've read the approach matches how CSS specifies font-family resolution, and the conflict is confined to `text/layout.rs`

Issue #722, opened 2024-02-29
`abs_transform` of `use` with position regression


Enter-tainer:
The [doc](https://docs.rs/usvg/latest/usvg/struct.Path.html#method.abs_transform) of `usvg::Path::abs_transform` :
> Element’s absolute transform.
> 
> Contains all ancestors transforms excluding element’s transform.
>
> Note that this is not the relative transform present in SVG. The SVG one would be set only on groups.

It says that element's transform is excluded. How to get it, so we can get the full absolute of an element?

To make it clear, I make a small function to print all `abs_transform` of all nodes in tree:

```rust
fn print_abs_ts(node: &usvg::Node) {
    if let usvg::Node::Group(g) = node {
        println!("group transform: {:?}", g.transform());
        for child in g.children() {
            print_abs_ts(child);
        }
    }
    println!("abs transform: {:?}", node.abs_transform());
}
fn main() {
    let svg = std::fs::read_to_string("test2.svg").unwrap();
    let tree = usvg::Tree::from_str(&svg, &usvg::Options::default()).unwrap();
    for node in tree.root().children() {
        print_abs_ts(node);
    }
}

```

For this svg(note the `x=30 y=20` part):
```
<svg class="typst-doc" viewBox="0 0 100 100" width="100" height="100" xmlns="http://www.w3.org/2000/svg"
  xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:h5="http://www.w3.org/1999/xhtml">
  <g transform="translate(33, 44)">
    <rect width="100" height="100" fill="white" />
    <rect width="10" height="10" x="30" y="20" fill="black"/>
  </g>
</svg>

```

The output is 
```
❯ cargo run
    Finished dev [unoptimized + debuginfo] target(s) in 0.02s
     Running `target\debug\usvg-bug.exe`
group transform: Transform { sx: 1.0, kx: 0.0, ky: 0.0, sy: 1.0, tx: 33.0, ty: 44.0 }
abs transform: Transform { sx: 1.0, kx: 0.0, ky: 0.0, sy: 1.0, tx: 33.0, ty: 44.0 }
abs transform: Transform { sx: 1.0, kx: 0.0, ky: 0.0, sy: 1.0, tx: 33.0, ty: 44.0 }
abs transform: Transform { sx: 1.0, kx: 0.0, ky: 0.0, sy: 1.0, tx: 33.0, ty: 44.0 }
```

[...]

PR #1037, opened 2026-03-19
fix: group abs_transform should only need to apply transform once

fundon:
If there is `transform` on the `use` element, the `abs_transform` on the group will be applied transform twice.

This will not cause problems in `resvg` rendering, but if it is parsed through `usvg`, the `abs_transform` obtained is incorrect, which will cause the custom rendering to be incorrect.

The correct `scale` should be **0.5**, not **0.25**.

https://github.com/linebender/resvg/blob/3a0fdba53ccf2d346b54cc53ba7adf0ee60d0707/crates/usvg/src/parser/use_node.rs#L273-L274

https://github.com/linebender/resvg/blob/3a0fdba53ccf2d346b54cc53ba7adf0ee60d0707/crates/usvg/src/parser/converter.rs#L768

* Original SVG

navigation-icon_full

* Before

Screenshot 2026-03-19 at 5 30 43 PM

* After

Screenshot 2026-03-19 at 5 31 53 PM


crates/resvg/tests/tests/structure/use/with-transform-in-use.svg
LaurenzV, line 1: You already added a unit test that checks this behavior, right? If so, let's remove this test, since we want to try to keep the test suite lean.

LaurenzV:
This change was explicitly introduced in https://github.com/linebender/resvg/commit/09338a13629db6a31a99d39367b3c156feb299fa to solve #722. So I'm not entirely sure whether this is the right fix. It does seem to me like the absolute transform of group2 should be translate(20, 20), since the transform is on the node itself? Or am I missing something?

LaurenzV:
I agree that a scale of 0.25 does seem wrong here, but I'm not sure if the fix you applied is the right one.

fundon:
Let me check the context. Thank you.

RazrFalcon:
If it does not affect `resvg` then why some tests have changed?

fundon:
> If it does not affect `resvg` then why some tests have changed?

I have no clue. Is it an error caused by the latest `oxipng`?

LaurenzV:
Can you try restoring the old snapshots and see whether tests still pass?

fundon:
> Can you try restoring the old snapshots and see whether tests still pass?

Tried. These use cases all have several pixel differences.

LaurenzV:
Ah ok. Well, I haven't gotten to reviewing this yet, but if it's just a very small diff it's _probably_ fine and due to some floating-point stuff.

fundon:
Fixed in #1056

PR #1040, opened 2026-04-01
fix: text nodes should inherit absolute transform

fundon:
The `abs_transform` is missing when rendering text as a path.

### DIFF
--- a/crates/usvg/src/text/flatten.rs
+++ b/crates/usvg/src/text/flatten.rs
@@ -26,6 +26,7 @@ fn push_outline_paths(
     builder: &mut tiny_skia_path::PathBuilder,
     new_children: &mut Vec<Node>,
     rendering_mode: ShapeRendering,
+    abs_transform: Transform,
 ) {
     let builder = mem::replace(builder, tiny_skia_path::PathBuilder::new());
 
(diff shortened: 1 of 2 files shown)


LaurenzV:
Thanks, though keep in mind this still won't work correctly for color/bitmap glyphs.

nicoburns:
Hmm... this requires the following change to `anyrender_svg`:

https://github.com/DioxusLabs/anyrender/pull/79/changes

Not necessarily a problem in a semver-breaking release, but I wanted to make sure was intended. Without this change I ended up double-applying transforms.

LaurenzV:
The new test definitely looks correct to me now. I only took a brief look, but this part of your code:

```rs
let transform = transform * util::to_affine(&node.abs_transform());
```

looks a bit suspicious to me. If you are already keeping a manual chain of transforms, then why are you chaining it again with the _absolute_ transform? 🤔 But I might have overlooked something.

fundon:
The `abs_transform` in outline path is the same as `abs_transform` in text.
If `text.flattened` is not used, `text.abs_transform` should be used.

RazrFalcon:
Not gonna lie, the whole `abs_transform` implementation is a pure mess. Mainly because it's not actually used by `resvg` (afair?). Therefore barely tested.

Trying to convert SVG to `usvg::Tree` in one go was a mistake...

LaurenzV:
I mean, we could just not expose it in the first place, no?

RazrFalcon:
Good question. I'm pretty sure we had this API since the first version. And it is required to position/render individual nodes/elements.

The difference is that the old, rctree version of `usvg::Tree` was traversing the tree to calculate this transform, instead of precomputing it.

PR #1041, opened 2026-04-06
Prefer declared font-family entries during glyph fallback

thekingofcity:
## Summary

This changes glyph fallback to prefer fonts from the declared `font-family` list, while preserving the control of custom `FontResolver::select_fallback` implementations.

The main design change is to pass the current span's `Font` into fallback resolution via a new `FallbackRequest` struct. This lets the default fallback resolver consider the declared family list without hardcoding that policy inside the text layout code.

## Design

The key question here was where the "prefer declared families during fallback" policy should live.

I considered two approaches:

1. Keep the family-first lookup inside `shape_text`
2. Move that policy into the default fallback resolver and give resolvers the context they need

I chose the second approach.

### Why this is a better fit

`shape_text` is part of layout execution. It should ask for a fallback candidate, but it should not embed policy that overrides user-provided fallback selection.

The previous approach introduced a family-first loop in layout before calling `select_fallback`. That had two problems:

- It bypassed custom `FontResolver::select_fallback` callbacks whenever one of the declared families already matched
- It made the default fallback policy effectively un-overridable from the public API

By introducing `FallbackRequest`, the default resolver can now implement family-first fallback itself, and custom resolvers receive the same context and remain fully in control.

This keeps the boundary cleaner:

- layout is responsible for detecting missing glyphs and attempting fallback shaping
- the resolver is responsible for selecting fallback font candidates
- the default resolver can prefer declared families, but custom resolvers are not bypassed

### Why `FallbackRequest` instead of just adding `&Font` as another parameter

I used a request struct instead of extending the callback with another positional parameter because this API is likely to need more context over time.

A request object makes that evolution easier and keeps the callback signature readable. It also makes the intent clearer: the fallback resolver is handling a specific missing-glyph fallback request, not a general text-layout task.

### Scope and non-goals

This PR does not turn `FontResolver` into a full browser-style font cascade or run segmentation API.

The fallback resolver still only selects candidate fonts for missing glyphs. It does not control:

- text segmentation
- shaping strategy
- how entire runs are split into `(text, font)` sequences

That distinction is important and matches the intended scope of `FontResolver`.

This PR also does not attempt to solve cluster-aware fallback merging. The existing fallback merge behavior is left as-is.

## What changed

- Added `FallbackRequest` to carry fallback context:
  - missing character
  - current span `Font`
  - already used font IDs
- Changed `FallbackSelectionFn` to take `FallbackRequest`
- Moved the family-first fallback policy out of `shape_text` and into `FontResolver::default_fallback_selector`
- Updated layout code to always call `select_fallback` with the full request instead of doing its own pre-selection
- Updated nested parser forwarding code to match the new callback signature
- Clarified the API docs to state that fallback selection is still missing-glyph candidate selection, not full text-run segmentation or browser-style font cascading
- Preserved fallback-shaped child glyphs even when the child span's declared `font-family` cannot be resolved, so invalid child family lists no longer make already-resolved fallback text disappear
- Added tests for:
  - default resolver preferring declared families during fallback
  - custom fallback resolvers still being invoked before any default family-first policy is applied
  - invalid child font families keeping fallback glyphs visible

## Behavioral impact

For the default resolver:

- fallback now prefers fonts from the declared `font-family` list when possible

For custom resolvers:

- custom `select_fallback` callbacks are no longer bypassed by layout-level family-first logic
- the callback now receives the current `Font` via `FallbackRequest`, so it can make family-aware decisions itself

## Breaking change

This changes the public fallback callback signature from:

```rs
Fn(char, &[ID], &mut Arc<Database>) -> Option<ID>
```

to:

```rs
Fn(FallbackRequest<'_>, &mut Arc<Database>) -> Option<ID>
```

Custom `FontResolver::select_fallback` implementations will need to be updated accordingly.

## Testing

Verified with:

```sh
cargo test -p usvg
```

tspan-font-family-sheet-1
tspan-font-family-sheet-2

### DIFF
--- a/crates/usvg/src/parser/mod.rs
+++ b/crates/usvg/src/parser/mod.rs
@@ -132,8 +132,8 @@ impl crate::Tree {
             #[cfg(feature = "text")]
             font_resolver: crate::FontResolver {
                 select_font: Box::new(|font, db| (opt.font_resolver.select_font)(font, db)),
-                select_fallback: Box::new(|c, used_fonts, db| {
-                    (opt.font_resolver.select_fallback)(c, used_fonts, db)
+                select_fallback: Box::new(|request, db| {
+                    (opt.font_resolver.select_fallback)(request, db)
                 }),
             },
             ..Options::default()
--- a/crates/usvg/src/text/layout.rs
+++ b/crates/usvg/src/text/layout.rs
@@ -907,6 +907,7 @@ fn process_chunk(
 
         let tmp_glyphs = shape_text(
             &chunk.text,
+            &span.font,
             font,
             span.small_caps,
             span.apply_kerning,
@@ -1292,6 +1293,7 @@ impl DatabaseExt for Database {
 /// Text shaping with font fallback.
 pub(crate) fn shape_text(
     text: &str,
+    font_node: &crate::Font,
     font: Arc<ResolvedFont>,
     small_caps: bool,
     apply_kerning: bool,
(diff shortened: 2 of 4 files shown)


thekingofcity:
I'll mark this ready when existing tests are fixed :)

luisbg:
@thekingofcity, are you still working on this?

FYI, #1036 solves the same issue without breaking public API. To be honest, not sure what's best but API changing needs to be decided carefully.

PR #1043, opened 2026-04-09
Correctly calculate glyph advances

fundon:
In some specific languages,  the  second glyph in the cluster is not positioned as expected.

Rendering locally using `text.flattened()` with `Malayalam Sangam MN.ttc` and `Sinhala Sangam MN.ttc`.

Here are the renderings for Malayalam and Sinhalese.

> The example is from https://www.kermitproject.org/utf8.html#glass

* Before
Screenshot 2026-04-09 at 3 17 09 PM

* After
Screenshot 2026-04-09 at 3 21 43 PM


LaurenzV:
Thanks!

fundon:
Thank you. Added test case.

PR #1045, opened 2026-05-05
usvg/parser: fix width/height computation when only one is specified

simartin:
In case only one of the height/width attributes are set and we have a viewBox, we currently report an incorrect size: the height/width that is set and 100% of the viewbox's attribute for the other attribute.

This is incorrect, and we should instead compute the missing attribute value from the specified one and the viewBox' aspect ratio, like was done by https://github.com/linebender/resvg/pull/715.

I have validated that this fixes servo's
   https://github.com/servo/servo/issues/41273


LaurenzV:
Thanks!

PR #1046, opened 2026-05-08
Fix `stylesheet_injection` test not checking its 5th rect

simartin:
This test's intention is to check all the rect's, however it does not check rect5 due to a typo that this patch fixes.


LaurenzV:
Thanks, good catch!

PR #1047, opened 2026-05-09
Support styling via CSS

simartin:
Per https://www.w3.org/TR/SVG/styling.html#StylingUsingCSS, SVG2 should support this capability.

(I noticed this was not the case when inspecting servo's SVG compliance test results)


simartin:
Marking it as draft because I uncovered an issue with inheritability while running servo's SVG testsuite along with this patch.

simartin:
With this version we only improve servo's SVG compliance testsuite:

```
simon@framedora:~/Dev/projects/servo/servo$ ./mach -v test-wpt --dev tests/wpt/tests/svg
Running WPT tests with /home/simon/Dev/projects/servo/servo/target/debug/servoshell
Running 1117 tests in web-platform-tests

  ▶ PASS [expected FAIL] /svg/geometry/reftests/circle-001.svg

  ▶ PASS [expected FAIL] /svg/geometry/reftests/circle-002.svg

  ▶ PASS [expected FAIL] /svg/geometry/reftests/circle-003.svg

  ▶ PASS [expected FAIL] /svg/geometry/reftests/rect-003.svg

  ▶ PASS [expected FAIL] /svg/path/property/d-none.svg

  ▶ PASS [expected FAIL] /svg/geometry/reftests/rect-001.svg

  ▶ PASS [expected FAIL] /svg/geometry/reftests/rect-002.svg

Ran 1117 tests finished in 113.8 seconds.
  • 1110 ran as expected.
  • 7 tests passed unexpectedly
```

simartin:
Any thoughts on this patch? Thanks!

luisbg:
lgtm

@nicoburns or @DJMcNab can this be merged?

It might need a quick follow up since ellipses may not be covered with the tests but I haven't confirmed that and it is small, and not a regression.

luisbg:
Sorry for the earlier drive-by. Here's what I actually checked:

- Read the issue and fix to familiarize myself with things
- Merges cleanly to main
- Run tests before and after. Before "1,793 passed / 0 failed;", after "1,797 passed / 0 failed;"
- The net +4 is six tests added and two removed. The two deletions are deliberate and correct, they asserted the old behaviour, and their own <desc> said "Geometry attributes can be set via style only in SVG 2." The replacement files are the same SVG with the expectation flipped.
- The PR adds x, y, cx, cy, r, d, width, height. I tried an SVG that sets the geometry of a circle, a rect and an ellipse entirely through CSS. Circle and rect render, the ellipse doesn't appear at all. It has no radii, so it's treated as invalid and dropped, with no warning. rx and ry are missing.
- Not a regression though. On main the circle is dropped too, since CSS geometry doesn't work for any shape there. So this is a strict improvement and I don't think the gap should block it.
- It also matches @simartin's own WPT results above: circle and rect tests flip to passing, and no ellipse test does.
- I didn't read the SVG 2 spec myself, an LLM did. Section "Geometry Properties" defines cx, cy, r, rx, ry, x, y, width and height. `AId::Rx` and `AId::Ry` do exist in `names.rs`, they just don't appear in the `presentation-attribute` list with this PR applied.

Worth landing as-is, with a follow-up for `rx/ry` plus an ellipse test.

docs/svg2-changelog.md
```diff
@@ -62,6 +62,10 @@ NOTE: This list is not final. This just things I was able to find so far. Patche
 
 ## Styling
 
+### Changed
+
+- `height`, `width`, `x`, `y`, `cx`, `cy`, `r` and `d` are presentation attributes now. Which means that they can be resolved from CSS now.
```
DJMcNab, line 67: I think this should be an `- [x]`, right?

Otherwise, I think this makes sense, although I don't think both sentences need to end with `now`...

DJMcNab (approved):
Other than the changelog, this seems plausible to me.

I don't think we're trying to maintain backwards compatibility with SVG 1.1's limitations. As evidence for this, I can see that my browser shows the same SVG rendering for both the old and the new versions.

PR #1050, opened 2026-05-19
Add --query-bbox

emersion:
Queries the document bounding box. This can be useful from scripts, e.g. to crop the image without rasterizing it.


emersion:
Gentle ping, if you have time please review :)

LaurenzV:
Maybe we just change `query_all` so that all nodes are printed, even the ones without node ID? Then you could simply filter the first entry from the output, and we don't need to add a new flag for this.

emersion:
Hm, sounds like this could get quite verbose for documents with a lot of elements… Nonetheless, I've opened https://github.com/linebender/resvg/pull/1100 which does this.

LaurenzV:
Okay thanks, I'll think about this again.

PR #1056, opened 2026-06-05
Fix transforms being applied twice in `abs_transform`

fundon:
In some cases, if context element is `<use>` node, the `abs_transform` will be applied transform twice.

So transforms need to be set before converting children.


fundon:
@RazrFalcon @LaurenzV Please take a look. Thanks.

crates/usvg/tests/parser.rs
```diff
@@ -601,3 +601,39 @@ fn flattened_text_should_inherit_absolute_transform() {
...
 }
+
+#[test]
+fn use_node_abs_transform() {
+    let svg = "
+    <svg viewBox='0 0 200 200'
+         xmlns='http://www.w3.org/2000/svg'
+         xmlns:xlink='http://www.w3.org/1999/xlink'>
+        <defs>
+            <image id='image1' x='0' y='0' width='100' height='100' xlink:href='data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAGQAAABkCAYAAABw4pVUAAABb0lEQVR4Xu3VUQ0AIAzEUOZfA87wAgkq+vGmoGlz2exz73IZAyNIpsUHEaTVQ5BYD0EEqRmI8fghgsQMxHAsRJCYgRiOhQgSMxDDsRBBYgZiOBYiSMxADMdCBIkZiOFYiCAxAzEcCxEkZiCGYyGCxAzEcCxEkJiBGI6FCBIzEMOxEEFiBmI4FiJIzEAMx0IEiRmI4ViIIDEDMRwLESRmIIZjIYLEDMRwLESQmIEYjoUIEjMQw7EQQWIGYjgWIkjMQAzHQgSJGYjhWIggMQMxHAsRJGYghmMhgsQMxHAsRJCYgRiOhQgSMxDDsRBBYgZiOBYiSMxADMdCBIkZiOFYiCAxAzEcCxEkZiCGYyGCxAzEcCxEkJiBGI6FCBIzEMOxEEFiBmI4FiJIzEAMx0IEiRmI4ViIIDEDMRwLESRmIIZjIYLEDMRwLESQmIEYjoUIEjMQw7EQQWIGYjgWIkjMQAzHQgSJGYjhWIggMQMxnAdKSlrwlejIDgAAAABJRU5ErkJggg=='/>
```
LaurenzV: Instead of an image could we just use a rect or something so we don't need the whole base64 string in the image?

fundon: `ImageHrefResolver` does not support this(`#rect1`). So use a normal URL instead of data URL.

LaurenzV:
Thanks!

PR #1058, opened 2026-06-13
fix(usvg): accept leading whitespace before the XML declaration

### LINKED ISSUE #1057: usvg fails to parse valid SVG when whitespace precedes the <?xml ?> declaration
## Description

`usvg::Tree::from_str` (and therefore `from_data`) rejects an otherwise valid
SVG document whenever any whitespace — a newline, space, tab, or carriage
return — precedes the `<?xml ?>` declaration. The underlying XML parser
(roxmltree) only allows an XML declaration as the very first thing in the
document, so it reports:

> unexpected XML declaration at 2:1

This is easy to hit in practice with a multi-line string literal that starts on
its own line, or with a pretty-printed `.svg`/`.svgz` file that has a leading
blank line. The exact same SVG parses fine when:
- it is compressed onto a single line, or
- the declaration is removed, or
- it is loaded from a file that begins directly with `<?xml` (e.g. via
  `include_bytes!`).

## Steps to reproduce

```rust
let opt = usvg::Options::default();

let body = "<svg xmlns='http://www.w3.org/2000/svg' width='10' height='10'>\
    <rect width='10' height='10'/></svg>";
let decl = "<?xml version='1.0' encoding='UTF-8'?>";

// Works:
assert!(usvg::Tree::from_str(&format!("{decl}{body}"), &opt).is_ok());

// Fails with "unexpected XML declaration":
assert!(usvg::Tree::from_str(&format!("\n{decl}\n{body}"), &opt).is_ok());
```
Bug has been found here by an iced-user: https://github.com/iced-rs/iced/issues/3357

StefanoD:
`Tree::from_str` (and thus `from_data`) rejected an otherwise valid SVG whenever any whitespace preceded the `<?xml ?>` declaration — e.g. a string literal that starts on its own line, or a pretty-printed file with a leading blank line. An XML declaration is only valid at the very start of the document, so the underlying parser reported "unexpected XML declaration".

Trim leading whitespace before handing the text to `roxmltree`. The same input compressed onto a single line, or loaded via `include_bytes!` from a file beginning with `<?xml`, already worked; this makes the formatted variant behave identically. gzip-compressed input benefits too, as it is routed through `from_str` after decompression.

Adds regression tests for both the plain and gzip-compressed cases.

Closes #1057
Reported downstream as iced-rs/iced#3357.

Generated by Claude.

### DIFF
--- a/crates/usvg/Cargo.toml
+++ b/crates/usvg/Cargo.toml
@@ -47,6 +47,7 @@ unicode-vo = { version = "0.1", optional = true }
 
 [dev-dependencies]
 once_cell = "1.21"
+flate2 = { version = "1.1", default-features = false, features = ["rust_backend"] } # SVGZ encoding in tests
 
 [features]
 default = ["text", "system-fonts", "memmap-fonts"]
--- a/crates/usvg/src/parser/mod.rs
+++ b/crates/usvg/src/parser/mod.rs
@@ -149,8 +149,14 @@ impl crate::Tree {
             ..Default::default()
         };
 
-        let doc =
-            roxmltree::Document::parse_with_options(text, xml_opt).map_err(Error::ParsingFailed)?;
+        // An XML declaration (`<?xml ?>`) is only valid as the very first thing
+        // in the document, so any leading whitespace before it makes the
+        // underlying XML parser reject the input. Trim it so that pretty-printed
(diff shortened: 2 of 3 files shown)


StefanoD:
Closing, see comment: https://github.com/linebender/resvg/issues/1057#issuecomment-4698677428

PR #1059, opened 2026-06-13
fix(usvg): use the viewport as the bbox of a sliced <image>

### LINKED ISSUE #1034: Bug: clip-path rendering gives different result compared to chrome/firefox
Hello

I experienced a difference in the result image when using Chrome (or Chromium-based browsers)/Firefox compared to resvg.

This particular problem is inside the rendering of a clip-path which contains a few arcs (one arc counter-clockwise and going over to a clockwise arc which then does everything in reverse).

Here are the testfiles which should show the behavior quite well:

[testfiles.zip](https://github.com/user-attachments/files/25611278/testfiles.zip)

.recieved it the resvg version of the svg
.verified is the chrom(ium)/firefox rendered version

According to ChatGPT the clip-path itself is fine and standard conform.

StefanoD:
An `<image>` with `preserveAspectRatio="… slice"` is scaled to *cover* its viewport and the overflow is cut away by a rectangular clip. Its bounding box is therefore the image's `x/y/width/height` rect (the viewport), but usvg derived it from the children, i.e. the oversized scaled-to-cover image.

As a result, an `objectBoundingBox` clip-path (or mask/filter) applied to a sliced image was resolved against the wrong, much larger box. With a non-square viewport this scaled the clip shape anisotropically and made it collapse — e.g. a decorative arched window clipped to a tall image rendered as a flat top instead of the intended curve, differing from Chrome/Firefox.

Set the slice wrapper group's bounding box to the image rect so downstream bbox-relative units are resolved correctly.

Adds a regression test asserting the clip transform is built from the viewport rect and not the scaled image.

Closes #1034

Generated by Claude

### DIFF
--- a/crates/usvg/src/parser/image.rs
+++ b/crates/usvg/src/parser/image.rs
@@ -277,6 +277,21 @@ pub(crate) fn convert_inner(
         g2.children.push(Node::Group(Box::new(g)));
         g2.calculate_bounding_boxes();
 
+        // With `slice`, the image is scaled to *cover* the viewport and the
+        // overflow is cut away by the rectangular clip above. The bounding box
+        // of the element is therefore its `rect` (the viewport), not the
+        // larger scaled-to-cover image that `calculate_bounding_boxes` derives
+        // from the children. Without this correction an `objectBoundingBox`
+        // clip/mask/filter applied to the `<image>` would be resolved against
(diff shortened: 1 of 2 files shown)


RazrFalcon:
I would have to investigate this one first. `objectBoundingBox` is a very tricky thing to change blindly.

PR #1065, opened 2026-06-14
Discard invalid color declarations instead of dropping the property (…

### LINKED ISSUE #914: SVG with both hex colored stroke and display-p3 stroke color is rendered with no stroke
Figma outputs svg files with both a standard SVG color (`stroke="#abcdef"`) as well as a style parameter meant to support future wide gamut colors:

```html
<svg width="16" height="16" viewBox="0 0 16 16" fill="none" xmlns="http://www.w3.org/2000/svg">
<path d="M13 13L9.66667 9.66667M3 6.88889C3 7.39959 3.10059 7.90528 3.29602 8.3771C3.49146 8.84892 3.77791 9.27763 4.13903 9.63875C4.50015 9.99987 4.92885 10.2863 5.40068 10.4818C5.8725 10.6772 6.37819 10.7778 6.88889 10.7778C7.39959 10.7778 7.90528 10.6772 8.3771 10.4818C8.84892 10.2863 9.27763 9.99987 9.63875 9.63875C9.99987 9.27763 10.2863 8.84892 10.4818 8.3771C10.6772 7.90528 10.7778 7.39959 10.7778 6.88889C10.7778 6.37819 10.6772 5.8725 10.4818 5.40068C10.2863 4.92885 9.99987 4.50015 9.63875 4.13903C9.27763 3.77791 8.84892 3.49146 8.3771 3.29602C7.90528 3.10059 7.39959 3 6.88889 3C6.37819 3 5.8725 3.10059 5.40068 3.29602C4.92885 3.49146 4.50015 3.77791 4.13903 4.13903C3.77791 4.50015 3.49146 4.92885 3.29602 5.40068C3.10059 5.8725 3 6.37819 3 6.88889Z" stroke="#FAFAFA" style="stroke:#FAFAFA;stroke:color(display-p3 0.9804 0.9804 0.9804);stroke-opacity:1;" stroke-width="0.892857" stroke-linecap="round" stroke-linejoin="round"/>
</svg>
```

When parsed and rendered by resvg, I receive the error "Failed to parse stroke value: 'color(display-p3 0.9804 0.9804 0.9804)'. Fallback to no stroke." Instead, the application should simply discard that part of the style or discard the entire style parameter, instead of falling back to no stroke.

StefanoD:
Fixes #914.

Figma exports SVGs with a dual stroke definition: a valid hex fallback
followed by a wide-gamut value, e.g.

    style="stroke:#FAFAFA;stroke:color(display-p3 0.9804 0.9804 0.9804);stroke-opacity:1;"

When splitting a `style` attribute (or applying a CSS rule), the last
declaration for a property always won, regardless of validity. The
unsupported `color(display-p3 ...)` value overwrote the valid `#FAFAFA`
one, and since stroke parsing later falls back to "no stroke", the
element was rendered invisibly.

Per CSS error recovery, a declaration with an invalid value must be
discarded so a previously declared valid value (or the presentation
attribute) is kept. This PR validates the color-bearing presentation
properties (`fill`, `stroke`, `color`, `flood-color`, `lighting-color`,
`stop-color`) before inserting them and skips invalid declarations.

Behavior for a *single* invalid declaration is unchanged (stroke → none,
fill → default/inherited).

### Testing
- Added regression test `invalid_style_declaration_falls_back_to_valid_one`
  (fails before, passes after).
- Full `usvg` test suite passes; `cargo build` clean with no new clippy warnings.

Generated by Claude

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@ This changelog also contains important changes in dependencies.
 
 This release has an MSRV of 1.87.0 for `usvg` and `resvg` and the C API.
 
+### Fixed
+- Discard invalid `fill`/`stroke`/color declarations in a `style` attribute or CSS rule instead of
+  dropping the property, so a previously declared valid value (or the presentation attribute) is kept.
+  This fixes elements becoming invisible when an unsupported color syntax such as
+  `color(display-p3 ...)` is used alongside a hex fallback (e.g. Figma exports). (#914)
+
 ## [0.47.0] 2026-02-05
 
 This release has an MSRV of 1.87.0 for `usvg` and `resvg` and the C API.
--- a/crates/usvg/src/parser/svgtree/parse.rs
+++ b/crates/usvg/src/parser/svgtree/parse.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
 use std::collections::HashMap;
+use std::str::FromStr;
 
 use roxmltree::Error;
 use simplecss::Declaration;
@@ -275,6 +276,21 @@ pub(crate) fn parse_svg_element<'input>(
             .iter_mut()
             .position(|a| a.name == aid);
 
+        // Per CSS error recovery, a declaration with an invalid value must be discarded so that a
+        // previously declared valid value (or the presentation attribute) is kept. We only need to
+        // validate when this declaration would override an existing one; a lone invalid value is
+        // handled, as before, by the fallback in the lazy conversion step. Restricting the check to
+        // the override case also keeps the common path free of a redundant parse.
+        // See https://github.com/linebender/resvg/issues/914
+        if idx.is_some() && !is_valid_declaration_value(aid, value) {
+            log::warn!(
+                "Failed to parse {} value: '{}'. Ignoring declaration.",
+                aid,
+                value
+            );
+            return;
+        }
+
         // Append an attribute as usual.
         let added = append_attribute(
             parent_id,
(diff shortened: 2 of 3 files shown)


nicoburns (changes requested):
The concept of discarding invalid declarations so that earlier valid ones take effect seems correct, but this implementation looks like it ends up parsing the declaration twice. Could we make use of `Option<T>` so that it only needs to be parsed once?

StefanoD:
@nicoburns Good catch. The attribute store keeps raw strings and the converter parses them lazily, so I can't thread a typed `Option<T>` all the way through to conversion without a much bigger refactor of the attribute storage. But you're right that my first version parsed *every* styled `fill`/`stroke`/color declaration eagerly, even in the common single-declaration case where the converter parses it again anyway.

I've reworked it so the validation only runs when a declaration would actually **override** an existing attribute — that's the only case where discarding the invalid one matters. A lone declaration keeps the previous behavior (the lazy fallback at conversion handles it), so the common, conflict-free path incurs no extra parse.

In the reported scenario, only the unsupported `color(display-p3 ...)` value gets parsed-and-discarded; the valid `#FAFAFA` (inserted first, no conflict) is parsed just once, at conversion time.

As a nice side effect, this also fixes the case where the invalid value comes from a `style` attribute that overrides a valid presentation attribute (e.g. `stroke="#FAFAFA"` + `style="stroke:color(display-p3 ...)"`), not just two declarations within the same `style` string.

Generated by Claude.

StefanoD:
Sorry, all my PRs have been closed. I deleted my fork with the consequence that all my PRs have been closed. Feel free to take the change set and open yourself a PR.

PR #1068, opened 2026-06-14
Apply root svg transform after the viewBox mapping

### LINKED ISSUE #899: Root svg transform inconsistent with Chromium/Firefox/Inkscape
I have found that resvg renders this SVG inconsistently with Chromium/Firefox/Inkscape:

```svg
<svg xmlns="http://www.w3.org/2000/svg" height="1000" transform="scale(.5)" viewBox="5 5 100 100" width="1000">
  <g>
    <circle cx="10" cy="10" r="3" stroke="black" stroke-width=".1"/>
    <circle cx="20" cy="10" r="3" stroke="black" stroke-width=".1"/>
    <line stroke="black" stroke-width=".1" x1="10" x2="20" y1="10" y2="10"/>
  </g>
</svg>

```

Here's a comparison of the output of Firefox/Chromium/Inkscape (left) and resvg (right):

<p align="center">

</p>

I suspect the culprit here is probably the `transform` on the root SVG coupled with `viewBox`. I believe the support for `transform` on the `svg` element was added in SVG 2 and there have been inconsistencies among implementations in the past (see f.e. https://lists.w3.org/Archives/Public/www-svg/2015May/0024.html). 

The output of resvg seems to be equivalent to rendering the following SVG with f.e. Firefox:

```svg
<svg xmlns="http://www.w3.org/2000/svg" height="1000" viewBox="5 5 100 100" width="1000">
  <g transform="scale(.5)">
    <circle cx="10" cy="10" r="3" stroke="black" stroke-width=".1"/>
    <circle cx="20" cy="10" r="3" stroke="black" stroke-width=".1"/>
    <line stroke="black" stroke-width=".1" x1="10" x2="20" y1="10" y2="10"/>
  </g>
</svg>
```

Might be related to #874.

StefanoD:
Fixes #899.

## Problem
A `transform` on the outermost `<svg>` combined with a `viewBox` rendered
differently from Chromium/Firefox/Inkscape. resvg applied the transform *inside*
the viewBox space, clipping content into the corner.

## Fix
The `transform` on the outermost `<svg>` is a CSS transform in the SVG viewport
coordinate system, i.e. applied *after* the viewBox-to-viewport mapping. Combine
it with the viewBox transform in `convert_doc` so it wraps the viewBox mapping,
and skip it in `convert_group` for the root `<svg>` to avoid applying it twice.

## Testing
- New tests `root_svg_transform_with_viewbox` and `nested_svg_transform_still_applies`.
- All 1724 `resvg` render tests pass.

Generated by Claude

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ This changelog also contains important changes in dependencies.
 
 This release has an MSRV of 1.87.0 for `usvg` and `resvg` and the C API.
 
+### Fixed
+- A `transform` on the outermost `svg` element is now applied in the SVG viewport
+  coordinate system (i.e. after the viewBox-to-viewport mapping), matching Chromium,
+  Firefox and Inkscape. (#899)
+
 ## [0.47.0] 2026-02-05
 
 This release has an MSRV of 1.87.0 for `usvg` and `resvg` and the C API.
--- a/crates/usvg/src/parser/converter.rs
+++ b/crates/usvg/src/parser/converter.rs
@@ -415,7 +415,14 @@ pub(crate) fn convert_doc(svg_doc: &svgtree::Document, opt: &Options) -> Result<
         }
     }
 
-    let root_ts = view_box.to_transform(tree.size());
+    // The `transform` attribute is allowed on the outermost `svg` element since SVG 2.
+    // Unlike a transform on a nested element, it's applied in the SVG viewport
+    // coordinate system, i.e. _after_ the viewBox-to-viewport mapping, just like
+    // Chromium, Firefox and Inkscape do. Therefore it must wrap the viewBox transform
+    // and not the other way around. The transform itself is skipped in `convert_group`
+    // for the root `svg` to avoid applying it twice.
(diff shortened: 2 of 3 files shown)


nicoburns (approved):
The spec says:

> If both [transform](https://www.w3.org/TR/SVG2/coords.html#TransformProperty) (or ‘[patternTransform](https://www.w3.org/TR/SVG2/pservers.html#PatternElementPatternTransformAttribute)’) and ‘[viewBox](https://www.w3.org/TR/SVG2/coords.html#ViewBoxAttribute)’ are applied to an element two new coordinate systems are established. [transform](https://www.w3.org/TR/SVG2/coords.html#TransformProperty) establishes the first new coordinate system for the element. ‘[viewBox](https://www.w3.org/TR/SVG2/coords.html#ViewBoxAttribute)’ establishes a second coordinate system for all descendants of the element. The first coordinate system is post-multiplied by the second coordinate system.

https://www.w3.org/TR/SVG2/coords.html#ViewBoxAttribute

So I believe this is correct.

PR #1079, opened 2026-06-28
Fix panic caused by oversized filter regions

### LINKED ISSUE #1007: Specific SVG input file causing panic

### LINKED ISSUE #1021: Crash on specific svg file
Tested on v0.46.0 and v0.47.0 (Win64)

I'm trying to convert SVG to PNG using this command:
`resvg.exe password-health-TP9ZOlqL.svg password-health-TP9ZOlqL.png`

I get the following error:
> thread 'main' panicked at crates\resvg\src\filter\composite.rs:26:5:
> assertion failed: src1.height == src2.height && src1.height == dest.height
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

The problematic SVG file is attached.

Image

StefanoD:
The source pixmap of a filtered group is clamped to `max_bbox` in `render_group`, but the filter `region` in `apply_inner` was derived directly from the unclamped filter rect. When the filter region was larger than the clamped buffer, `feComposite` with the `arithmetic` operator panicked on a size mismatch, since it requires its inputs and destination to have identical dimensions:

    assertion failed: src1.height == src2.height && src1.height == dest.height

All intermediate filter images are expected to share the source's dimensions, so clamp the region to the source bounds. This keeps every buffer the same size and makes the affected SVG render correctly instead of crashing (or being silently cleared).

This also fixes the `huge-region` test, whose reference image previously captured the buggy behaviour where the filtered element disappeared; it now renders the blurred shape correctly.

Fixes #1021. Fixes #1007.

Generated by Claude.

---
*Note: I accidentally closed the original PR (#1061) by deleting my fork, which auto-closed it. This reopens the same change — the branch and commits are unchanged.*


LaurenzV:
I just noticed there seem to be quite a few other issues related to filters, but this definitely is an improvement to before, so LGTM.

PR #1080, opened 2026-06-28
Make `href` take precedence over `xlink:href`

### LINKED ISSUE #1015: SVG2: `href` must take precedence over `xlink:href` when both are present
According to the SVG 2 specification, the XLink-based URL reference attributes (such as `xlink:href`) are deprecated and replaced by the unprefixed `href` attribute.

The specification explicitly defines the conflict resolution behavior when both attributes are present on the same element:

> When the ‘href’ attribute is present in both the XLink namespace and without a namespace, the value of the attribute without a namespace shall be used. The attribute in the XLink namespace shall be ignored.

(Source: SVG 2 – Linking, “Deprecated XLink URL reference attributes”)
https://www.w3.org/TR/SVG2/linking.html

which is not compliant with the SVG 2 specification.

According to the spec, the unnamespaced `href` attribute must always take precedence when both are present, and `xlink:href` must be ignored.

This affects elements such as `<textPath>`, `<use>`, `<image>`, `<a>`, and other elements that accept URL reference attributes.

## Demo
```html
<svg width="200" height="200" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
  <defs>
    <rect id="red" x="0" y="0" width="100" height="100" fill="red"/>
    <rect id="green" x="0" y="0" width="100" height="100" fill="green"/>
  </defs>

  <!-- According to SVG 2, href MUST win, so this should be GREEN -->
  <use
    href="#green" xlink:href="#red"
    x="20" y="20" />
</svg>
```

## resvg 0.46.0
Image

## Chrome 143/Firefox 149
Image

## WPT
https://github.com/w3c/svgwg/issues/1043

StefanoD:
SVG 2 requires that when both the unprefixed `href` and the deprecated `xlink:href` are present on an element, the value of `href` is used and the XLink-namespaced attribute is ignored, regardless of source order.

Two code paths were non-compliant:

- `resolve_href` (used by `<use>`) queried `xlink:href` first and only fell back to `href`, so `xlink:href` always won. Note that `roxmltree::Node::attribute("href")` matches by local name only and ignores the namespace, so the attributes are now filtered explicitly.
- The generic attribute-copy loop stored both `href` and `xlink:href` under `AId::Href`, and lookups returned whichever came first in document order. The unprefixed `href` now always wins.

Fixes #1015

Generated by Claude

---
*Note: I accidentally closed the original PR (#1062) by deleting my fork, which auto-closed it. This reopens the same change — the branch and commits are unchanged.*


LaurenzV:
Thanks, also checked in Chrome and they seem to do the same.

PR #1084, opened 2026-06-28
Fix properties of nested SVGs being applied twice

### LINKED ISSUE #1066: Nested `<svg>` `transform` is applied twice (inconsistent with `<use>` and browsers)
## Description

A `transform` attribute on a **nested** `<svg>` element is applied **twice** by usvg/resvg. The nested `<svg>` gets wrapped in an extra group, so its `transform` (and other group properties such as `opacity`) end up doubled.

This is inconsistent with how `<use>` is handled (which is correct) and with browsers (Chromium/Firefox).

I found this incidentally while investigating #899 (a `transform` on the **root** `<svg>`). It is a separate bug with a different cause, so I'm filing it on its own.

## Steps to reproduce

```xml
<svg xmlns="http://www.w3.org/2000/svg" width="200" height="200">
  <svg transform="translate(50,50)" width="40" height="40" overflow="visible">
    <rect width="20" height="20" fill="red"/>
  </svg>
</svg>
```

Render this with resvg.

## Expected

The red 20×20 rect should be drawn with its top-left at **(50,50)** (i.e. bbox (50,50)–(70,70)), matching Chromium/Firefox.

## Actual

resvg draws it with its top-left at **(100,100)** (bbox (100,100)–(120,120)) — the `translate(50,50)` is applied twice.

## Root cause

Dumping the usvg tree shows two nested groups, each carrying the transform:

```
<g transform="matrix(1 0 0 1 50 50)">
    <g transform="matrix(1 0 0 1 50 50)">
        <path .../>
```

whereas the equivalent `<use transform="translate(50,50)">` correctly produces a single group.

In `convert_element` (`crates/usvg/src/parser/converter.rs`) a nested `<svg>` is first wrapped by `convert_group` (which reads the element's `transform`), and then `convert_svg` creates its own group as well — so the transform is applied in both. `<use>` avoids this because it is routed to its handler and returns *before* the `convert_group` wrapper. Routing nested `<svg>` the same way fixes it.

**Version:** usvg/resvg 0.47.0 (current `main`).

Generated by Claude

StefanoD:
Fixes #1066.

## Problem
A `transform` on a nested `<svg>` element was applied twice. The same happens to
other group properties (e.g. `opacity`). For example, a rect under
`<svg transform="translate(50,50)">` was rendered at (100,100) instead of (50,50).

## Cause
In `convert_element`, a nested `<svg>` was wrapped in a `convert_group` (which
reads the element's `transform`), and then `convert_svg` created its own group as
well — so the transform was applied in both. `<use>` does not have this problem
because it is routed to its handler and returns *before* the `convert_group` wrapper.

## Fix
Route a nested `<svg>` directly to `convert_svg` and return early, exactly like
`<use>`. The outermost `<svg>` (no parent element) is unaffected.

## Testing
- New regression test `nested_svg_transform_applied_once` (fails on `main` with
  `translate(20,40)`, passes with the fix as `translate(10,20)`).
- The full `usvg` test suite and all 1724 `resvg` render tests pass.

Generated by Claude

---
*Note: I accidentally closed the original PR (#1067) by deleting my fork, which auto-closed it. This reopens the same change — the branch and commits are unchanged.*


LaurenzV:
Thanks!

PR #1086, opened 2026-06-28
Inherit `dominant-baseline` into nested `<tspan>` elements

### LINKED ISSUE #864: Nested `tspan` positioning is weird
SVG:

```svg
<svg xmlns="http://www.w3.org/2000/svg" width="1500" height="600" viewBox="0 0 1500 600">
    <text x="50" y="550" text-anchor="start" dominant-baseline="text-after-edge"
          font-family="Nunito" fill="white"
          style="filter:drop-shadow(0 3px 7px black)">
        <tspan x="50" dy="-190" font-size="120px" font-weight="800">Every Time ✅</tspan>
        <tspan x="50" dy="113" font-size="66px" font-weight="800">
            <tspan>hello</tspan>
            world
        </tspan>
    </text>
</svg>
```

Render (notice how "hello" got shifted):

Screenshot 2024-12-17 at 21 28 08

My understanding is that it should be still in line.

StefanoD:
`dominant-baseline` was listed in `is_non_inheritable`, so usvg only let a `<tspan>` pick up the value from its *direct* parent element. For a nested `<tspan>` whose immediate parent does not carry the property, the value of an ancestor `<text>` was therefore lost and the span fell back to `auto` (alphabetic baseline). The nested run then sat on a different baseline than its siblings — the "weird" positioning reported in #864.

`dominant-baseline` is actually an *inherited* property, unlike its siblings `alignment-baseline` and `baseline-shift`:

- SVG 2 / CSS Inline Layout Module Level 3 define it with "Inherited: yes". https://www.w3.org/TR/SVG2/text.html#BaselineAlignmentProperties
- Chromium marks it `inherited: true` (field group `svginherited`), while `alignment-baseline` and `baseline-shift` are not inherited: https://github.com/chromium/chromium/blob/main/third_party/blink/renderer/core/css/css_properties.json5 (`name: "dominant-baseline" ... inherited: true`)
- Inkscape declares it as an inheriting `SPIEnum` (default `inherits = true`).

Removing it from `is_non_inheritable` makes `find_attribute`/`resolve_inherit` walk the full ancestor chain, matching Chrome's rendering (verified by comparison) and the spec.

Fixes #864.

Generated with Claude

---
*Note: I accidentally closed the original PR (#1070) by deleting my fork, which auto-closed it. This reopens the same change — the branch and commits are unchanged.*


luisbg:
lgtm

Approach is correct but I'm not sure the comment block in `crates/usvg/src/parser/svgtree/mod.rs` is needed.

StefanoD:
I won't have the time to continue with the PR. You can take it over.

PR #1088, opened 2026-07-01
Reduce required dependencies with new feature gates

reed-smout:
Kia ora!

This is a small modification to `usvg`'s features to reduce the size of the dependency tree for the common (imo) use case of a plain SVG parsing/simplification library.

Hopefully I'm not jumping the gun by going straight to a PR, but I thought it would be more helpful than filing an issue and making you guys do it :)

This does three things (broken into three commits):
 - Move the dependencies needed for writing SVG strings behind a new default `writer` feature. As some text functions seem to require SVG writing, the `text` feature also requires `writer`.
 - Move `pico-args` behind a new `cli` feature. This seems to only be needed by the binary, but in the interest of caution and not making any breaking changes for default users, I enabled `cli` by default too. Let me know if you'd rather leave this out of the default features.
 - Move `flate2` behind a new `svgz` feature. I've made attempting to parse SVGZ data with the feature disabled an `Error` variant, but it could just as easily be an assert or panic!/unimplemented! if preferred. This also needs a corresponding `resvg` feature following the same pattern and handling in the C api crate. Let me know if this should be handled differently by the C api.

All in all there should be no changes to any consumers of the crate(s) apart for a significant reduction in dependency count and compile time for some!

Cheers and thanks for all the great work!

### DIFF
--- a/crates/c-api/lib.rs
+++ b/crates/c-api/lib.rs
@@ -22,6 +22,8 @@ pub enum resvg_error {
     OK = 0,
     /// Only UTF-8 content are supported.
     NOT_AN_UTF8_STR,
+    /// `resvg` must be compiled with SVGZ decoding support.
+    SVGZ_UNSUPPORTED,
     /// Failed to open the provided file.
     FILE_OPEN_FAILED,
     /// Compressed SVG must use the GZip algorithm.
@@ -853,6 +855,7 @@ fn cstr_to_str(text: *const c_char) -> Option<&'static str> {
 fn convert_error(e: usvg::Error) -> resvg_error {
     match e {
         usvg::Error::NotAnUtf8Str => resvg_error::NOT_AN_UTF8_STR,
+        usvg::Error::SvgzFeatureNotEnabled => resvg_error::SVGZ_UNSUPPORTED,
         usvg::Error::MalformedGZip => resvg_error::MALFORMED_GZIP,
         usvg::Error::ElementsLimitReached => resvg_error::ELEMENTS_LIMIT_REACHED,
         usvg::Error::InvalidSize => resvg_error::INVALID_SIZE,
--- a/crates/usvg/Cargo.toml
+++ b/crates/usvg/Cargo.toml
@@ -15,7 +15,7 @@ workspace = "../.."
 
 [[bin]]
 name = "usvg"
-required-features = ["text", "system-fonts", "memmap-fonts", "writer"]
+required-features = ["svgz", "text", "system-fonts", "memmap-fonts", "writer"]
 
 [dependencies]
 base64 = { version = "0.22", optional = true } # for embedded images
@@ -28,7 +28,7 @@ xmlwriter = { version = "0.1", optional = true }
 
 # parser
 data-url = "0.3" # for href parsing
-flate2 = { version = "1.1", default-features = false, features = ["rust_backend"] } # SVGZ decoding
+flate2 = { version = "1.1", default-features = false, features = ["rust_backend"], optional = true } # SVGZ decoding
 imagesize = "0.14.0" # raster images size detection
 kurbo = "0.13.0" # Bezier curves utils
 roxmltree = "0.21.1"
@@ -49,7 +49,9 @@ unicode-vo = { version = "0.1", optional = true }
 once_cell = "1.21"
 
 [features]
-default = ["text", "system-fonts", "memmap-fonts", "writer"]
+default = ["svgz", "text", "system-fonts", "memmap-fonts", "writer"]
+# Enables SVGZ decoding
+svgz = ["flate2"]
 # Enables text-to-path conversion support.
 # Adds around 400KiB to your binary.
 text = ["fontdb", "rustybuzz", "ttf-parser", "unicode-bidi", "unicode-script", "unicode-vo", "writer"]
--- a/crates/usvg/Cargo.toml
+++ b/crates/usvg/Cargo.toml
@@ -15,16 +15,16 @@ workspace = "../.."
 
 [[bin]]
 name = "usvg"
-required-features = ["text", "system-fonts", "memmap-fonts"]
+required-features = ["text", "system-fonts", "memmap-fonts", "writer"]
 
 [dependencies]
-base64 = "0.22" # for embedded images
+base64 = { version = "0.22", optional = true } # for embedded images
 log = "0.4"
 pico-args = { version = "0.5", features = ["eq-separator"] }
 strict-num = "0.1.1"
 svgtypes = "0.16.1"
 tiny-skia-path = "0.12.0"
-xmlwriter = "0.1"
+xmlwriter = { version = "0.1", optional = true }
 
 # parser
 data-url = "0.3" # for href parsing
@@ -49,11 +49,13 @@ unicode-vo = { version = "0.1", optional = true }
 once_cell = "1.21"
 
 [features]
-default = ["text", "system-fonts", "memmap-fonts"]
+default = ["text", "system-fonts", "memmap-fonts", "writer"]
 # Enables text-to-path conversion support.
 # Adds around 400KiB to your binary.
-text = ["fontdb", "rustybuzz", "ttf-parser", "unicode-bidi", "unicode-script", "unicode-vo"]
+text = ["fontdb", "rustybuzz", "ttf-parser", "unicode-bidi", "unicode-script", "unicode-vo", "writer"]
 # Enables system fonts loading.
 system-fonts = ["fontdb/fs", "fontdb/fontconfig"]
 # Enables font files memmaping for faster loading.
 memmap-fonts = ["fontdb/memmap"]
+# Enables converting trees back to SVG.
+writer = ["xmlwriter", "base64"]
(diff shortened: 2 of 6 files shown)


RazrFalcon:
```
> cargo bloat --release --crates -n 0
 File  .text     Size Crate
16.6%  29.7% 418.7KiB usvg
13.0%  23.2% 326.7KiB std
10.6%  18.9% 265.7KiB rustybuzz
 4.8%   8.6% 121.2KiB ttf_parser
 1.9%   3.4%  47.8KiB roxmltree
 1.9%   3.3%  47.1KiB svgtypes
 1.4%   2.5%  34.7KiB tiny_skia_path
 1.2%   2.2%  31.2KiB fontdb
 1.2%   2.1%  29.4KiB simplecss
 0.8%   1.4%  19.8KiB imagesize
 0.6%   1.0%  14.4KiB miniz_oxide
 0.4%   0.7%  10.5KiB data_url
 0.3%   0.6%   8.1KiB unicode_bidi
 0.3%   0.6%   8.1KiB pico_args
 0.3%   0.5%   6.4KiB xmlwriter
 0.2%   0.3%   4.6KiB unicode_ccc
 0.2%   0.3%   4.1KiB kurbo
 0.1%   0.2%   2.9KiB [Unknown]
 0.1%   0.2%   2.3KiB flate2
 0.1%   0.2%   2.2KiB unicode_vo
 0.1%   0.1%   1.5KiB crc32fast
 0.0%   0.1%   1.1KiB base64
 0.0%   0.1%     996B smallvec
 0.0%   0.0%     476B simd_adler32
 0.0%   0.0%     472B memmap2
 0.0%   0.0%     460B unicode_bidi_mirroring
 0.0%   0.0%     380B unicode_script
 0.0%   0.0%     376B unicode_properties
 0.0%   0.0%     184B log
 0.0%   0.0%      20B __rustc
55.9% 100.0%   1.4MiB .text section size, the file size is 2.5MiB
```

As you can see, your change wouldn't do much. Text is the biggest dependency. And `usvg` itself is quite heavy.

reed-smout:
I'm not overly concerned with final binary size, like you say it wouldn't actually change much (I get a ~30KiB reduction mainly due to removing `miniz_oxide`).

However (for `--no-default-features`, my primary concern), these changes do reduce the number of compilation units by ~40% (28 to 17), representing a significant compile time reduction of ~25% (4.3s to 3.2s) on lower-end CPUs (`--jobs=4`) and ~10% (2.8s to 2.5s) on CPUs with more cores (`--jobs=8`). This also shaves ~1s (20.8s to 19.7s) off the normal compile time for my project (Ryzen 5600XT CPU).

Fewer dependencies pulled in means less compilation units, less parallelism needed during compilation and a lighter dependency tree to manage. It's also just good library design imo to gate optional functionality that requires extra dependencies behind cargo features.

If you wanted, I could restrict the new cargo features to just `usvg` and leave `resvg` and it's C bindings untouched? People using `resvg` probably want all the bells and whistles anyway. I'm also open any other suggestions you might have!

All that said, if you really don't think this is workable, you're welcome to close this PR.

RazrFalcon:
More build features - harder maintenance burden. 
The `cli` one would not work. People have tried removing it multiple times already. Search issues.
Also, `base64` is used by the parser as well.

`svgz` and `writer` are probably fine.

reed-smout:
I've removed the `cli` feature commit. I couldn't find where in the parsing `base64` is used? I found a few statements saying base64 data is decoded by `usvg` but couldn't see the actual implementation that did that... maybe I'm not looking in the right place?

RazrFalcon:
I guess `data-url` already does it for us. My bad. We had way too many rewrites.

crates/usvg/src/parser/mod.rs
```diff
@@ -30,6 +31,10 @@ pub enum Error {
     /// Only UTF-8 content are supported.
     NotAnUtf8Str,
 
+    /// `svgz` feature is required to parse SVGZ data.
+    #[cfg(not(feature = "svgz"))]
+    SvgzFeatureNotEnabled,
```
LaurenzV: I don't think it's a good idea to feature-gate a public error variant, let's just always include it.

crates/usvg/src/parser/mod.rs
```diff
@@ -22,6 +22,7 @@ mod text;
 pub(crate) use converter::Cache;
 pub use image::{ImageHrefDataResolverFn, ImageHrefResolver, ImageHrefStringResolverFn};
 pub use options::Options;
+#[allow(unused_imports)]
```
LaurenzV: Can we instead feature-gate it properly?

crates/usvg/src/parser/mod.rs
```diff
@@ -94,12 +103,19 @@ impl<T> OptionLog for Option<T> {
 impl crate::Tree {
     /// Parses `Tree` from an SVG data.
     ///
-    /// Can contain an SVG string or a gzip compressed data.
+    /// Can contain an SVG string or a gzip compressed data (processing gzip compressed SVG
+    /// requires the `svgz` cargo feature).
```
LaurenzV: Don't think we need to explicitly mention this in the comment here.

crates/usvg/Cargo.toml
```diff
@@ -15,26 +15,26 @@ workspace = "../.."
...
 # parser
 data-url = "0.3" # for href parsing
-flate2 = { version = "1.1", default-features = false, features = ["rust_backend"] } # SVGZ decoding
 imagesize = "0.14.0" # raster images size detection
 kurbo = "0.13.0" # Bezier curves utils
 roxmltree = "0.21.1"
 simplecss = "0.2"
 siphasher = "1.0" # perfect hash implementation
 
+# SVGZ decoding
```
LaurenzV: Let's keep the deps in their original location, just make them optional.

crates/usvg/Cargo.toml
```diff
@@ -45,15 +45,23 @@ unicode-bidi = { version = "0.3", optional = true }
...
 
 [features]
-default = ["text", "system-fonts", "memmap-fonts"]
+default = ["svgz", "text", "system-fonts", "memmap-fonts", "writer"]
+# Enables SVGZ decoding
+svgz = ["flate2"]
 # Enables text-to-path conversion support.
 # Adds around 400KiB to your binary.
-text = ["fontdb", "rustybuzz", "ttf-parser", "unicode-bidi", "unicode-script", "unicode-vo"]
+text = ["fontdb", "rustybuzz", "ttf-parser", "unicode-bidi", "unicode-script", "unicode-vo", "writer"]
```
LaurenzV, line 57: Why is the `writer` feature needed for this one?

reed-smout: `colr` in `text` uses `xmlwriter` to construct some things then parses it to a `Tree`. I figured changing that to not rely on `xmlwriter` was way out of scope (and my knowledge of the codebase), so I made `text` depend on `writer`. Let me know if you have a better idea.

LaurenzV: I see, thanks for letting me know!

LaurenzV (approved):
Thanks!

crates/c-api/lib.rs
```diff
@@ -22,6 +22,8 @@ pub enum resvg_error {
     OK = 0,
     /// Only UTF-8 content are supported.
     NOT_AN_UTF8_STR,
+    /// `resvg` must be compiled with SVGZ decoding support.
+    SVGZ_UNSUPPORTED,
```
kleisauke, line 26: This also needs to be added after this line:
https://github.com/linebender/resvg/blob/68b14c4c3bccdb60344c777406486b54c36ec1a4/crates/c-api/resvg.h#L29-L32

Otherwise, in resvg 0.48.1 its C API, the following SVG:
```svg
<svg xmlns="http://www.w3.org/2000/svg" width="0" height="0"></svg>
````

returns `RESVG_ERROR_PARSING_FAILED` instead of `RESVG_ERROR_INVALID_SIZE`, because all subsequent enum values are shifted by one.

FWIW, this was noticed in wasm-vips here:
https://github.com/kleisauke/wasm-vips/blob/v0.0.18/test/unit/test_foreign.js#L1109-L1112

while trying to update to resvg 0.48.1.

LaurenzV: Oh no, sorry about that, that wasn't intentional!

Just for my understanding though, in theory this is "just" a breaking change, which would be fine since we bumped from 0.47 to 0.48, right? Or is there something else I'm missing? Unfortunately I'm not super familiar with C-APIs in Rust.

Obviously it should be documented and not happen in the first place since it's unnecessary, but still.

kleisauke: AFAIK, inserting a new enum value in the middle can break ABI compatibility (at least in C). I think that's acceptable for resvg, though, since SemVer allows breaking changes during the 0.y.z development phase, see:
https://semver.org/#spec-item-4

LaurenzV: Okay, thanks. 👍 I'll open an issue for it, we should probably still fix this and document it in the changelog.

PR #1089, opened 2026-07-03
Fix panic on filter regions with coordinates outside the i32 range

jschwe:
NonZeroRect::to_int_rect panics when the rect has coordinates outside the i32 range.
Instead, add a clamped_int_rect helper, which performs the clamping without panicking.
Adds a test, which panics before this fix.

Background: We started fuzzing servo more intensively recently, and `resvg` is a source of panics. This PR fixes one such panic. There are a couple more crashes pending.


luisbg:
The panic this fixes no longer exists in `main`. Regression test passes.

#989  got there first, by replacing the panicking `bbox.to_int_rect()` with `IntRect::from_xywh(...)?`.

But we shouldn't close this PR because the two fixes are not equivalent:
-  main (#989 ) returns None on an out-of-range rect, so the whole group is skipped and the content silently disappears.
-  #1089  does the arithmetic in f64 and clamps to the canvas, so the visible part still renders.

Want to rewrite the PR?

Its-Just-Nans:
https://github.com/linebender/tiny-skia/pull/182 is another fix (which fix the underlying crate)

luisbg:
The issue this PR tries to fix was resolved by #989 .

I checked by applying this PR's regression test alone to current main, with none of its source changes and it passes. The `<desc>` in the test SVG says _"the filtered element is invisible, but the app should not panic"_, and that's exactly what main does now.

We should close this PR.

PR #1097, opened 2026-07-21
Switch rustybuzz + ttf-parser to harfrust + skrifa

akx:
Per

* [RUSTSEC-2026-0206](https://rustsec.org/advisories/RUSTSEC-2026-0206), `rustybuzz` is unmaintained, and `harfrust` is an alternative.
* [RUSTSEC-2026-0192](https://rustsec.org/advisories/RUSTSEC-2026-0192), `ttf-parser` is unmaintained, and `skrifa` is an alternative.

This PR switches use of those crates in `usvg` to those alternatives, with the requisite API changes.

There are minor (to the tune of e.g. 128 -> 160 or vice versa; verified by eye and Beyond Compare) alpha coverage changes in the render tests, hence regenerated fixtures.
Tests were otherwise not touched by this change, which seems like a Good Thing.

> [!NOTE]
> The initial pass of this transformation was implemented with the assistance of Claude Opus 4.8 (as noted in the Assisted-By trailer).
>
> I've read every line here, though, and rewrote some things the model deemed would have been fine to duplicate or reimplement across the project.
>  
> (The model had also decided on an implementation that would've been about 40 times slower than the original; this was only noticed by me running the tests and wondering why one took a surprisingly long time.)
>
> I also took manual care to try and keep the diff minimal where possible.


Its-Just-Nans:
Related to https://github.com/linebender/resvg/pull/922

akx:
@Its-Just-Nans Ah! I didn't think to look at the second page of PRs 😄 

For context, I started working on this because cargo-deny downstream on [egui](https://github.com/emilk/egui)'s CI became red due to RUSTSEC-2026-0206.

Its-Just-Nans:
> For context, I started working on this because cargo-deny downstream on [egui](https://github.com/emilk/egui)'s CI became red due to RUSTSEC-2026-0206.

Yes no problem, that makes sense! there are multiple links to #922

LaurenzV:
Thank you for the contribution, but as mentioned, #922 already exists and we intend on landing this in the next few days.

PR #1100, opened 2026-07-28
Print nodes without ID in --query-all

emersion:
This can be useful to figure out the bounding box of an SVG document.

Alternative to https://github.com/linebender/resvg/pull/1050

### DIFF
--- a/crates/resvg/src/main.rs
+++ b/crates/resvg/src/main.rs
@@ -199,7 +199,7 @@ OPTIONS:
                                 Useful for debugging
 
 
-  --query-all                   Queries all valid SVG ids with bounding boxes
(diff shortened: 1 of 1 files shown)


luisbg:
Both of these are by @emersion and #1100  is explicitly an alternative to #1050, so I think we should pick one. Both merge clean and pass the suite.

#1050  prints the document bbox directly. #1100  prints every node's bbox and the caller unions them, it never prints the root, so it doesn't directly answer "what's the document bbox". It also changes `--query-all` output contract, though ID'd lines keep their content and order.

I personally prefer #1050  as additive with no compat risk, but changing the `--query-all` output is a policy call.

@LaurenzV  @nicoburns , preference?

PR #1103, opened 2026-07-31
Upgrade `strict-num` and `base64` dependencies

nicoburns:
Version bumps:

- strict-num 0.1 -> 0.2
- base64 0.22 -> 0.23


LaurenzV:
I think tiny skia uses strict-num 0.1 still, so maybe this should be reverted so we don’t have it twice?

PR #1107, opened 2026-08-01
Update README + changelog

LaurenzV:
I think now should be a good enough cut-off point. I removed two entries that are not really user-facing (for example the production profile).


RazrFalcon:
Congrats 🎉 

Worth removing this section completely: https://github.com/linebender/resvg#resvg-project

nicoburns (approved):

PR #1114, opened 2026-08-03
Add ARM emulation to win explorer installer

one-good-chance:
Allow ARM emulation in the explorer installer


luisbg:
Correct value, `x64compatible` is also the non-deprecated spelling since Inno 6.3, and the runner has 6.7.1, so the release build is fine.

Two things worth noting rather than blocking:
- Have you been able to confirm thumbnails actually render on ARM64? The installer still ships the x64 `server.dll` and registers it via `InprocServer32`, so this depends on the shell hosting it in an emulated surrogate. CI can't tell us.
- #1132  just landed a `windows-11-arm` job, so we could now build a native ARM64 `server.dll` and install the matching one (probably the real fix). Happy to open an issue for that.

PR #1115, opened 2026-08-03
Render monochrome, grayscale and BGRA bitmap glyphs

oetiker:
Implements the `// TODO: implement other bitmap formats` in `flatten.rs`. Until now only
`BitmapData::Png` was handled and `BitmapData::Mask` / `BitmapData::Bgra` returned `None`,
so glyphs from monochrome bitmap fonts silently fell back to the outline, or rendered as
nothing at all in bitmap-only fonts.

### Masks

A mask stores coverage rather than color, so it is painted with the fill of the text span,
just like an outline glyph would be. All four bit depths the spec allows (1, 2, 4 and 8 bpp)
are supported, in both the row-packed and the byte-aligned layout.

Because the result now depends on the fill, the raster cache is keyed by the color and
opacity as well. Non-solid paints (gradients, patterns) cannot be expressed by an image and
fall back to black.

### BGRA

BGRA data is premultiplied while PNG stores straight alpha, so the color channels are
unpremultiplied during the conversion.

### Encoding

`ImageKind` can only carry encoded image data, so the decoded bitmaps have to be re-encoded
as PNG. This adds `png` to `usvg` behind the `text` feature. It is already in the dependency
tree via `tiny-skia`, so it costs no additional transitive crates.

### Tests

`mask_coverage` is covered by unit tests across all bit depths and both layouts, since a
single font can only exercise one combination.

For the render tests, `BitmapMono.subset.ttf` is added: a 7.6 KiB subset of Terminus (TTF)
carrying a single 16ppem 1bpp `EBDT` strike. `pyftsubset` drops `EBDT`/`EBLC`, so the strike
is pruned and re-attached by `make-bitmap-mono.py`, which is included next to the font and
documented in the fonts README. Terminus is OFL with a Reserved Font Name, so the script
also renames the font, as required for a Modified Version.

Verified that these tests actually cover the new path: on `main` the test SVG flattens to
zero `<image>` nodes (outline fallback), with this change it flattens to one per glyph.

Full suite passes (1734 render tests, no reference image changes to existing tests).


oetiker:
Pushed two follow-ups after looking more closely at the rendered output.

Mask glyphs are now drawn without interpolation. A mask comes from a pixel font, drawn for one specific size, so smoothing blurs the very grid it was drawn on. It also bleeds into the transparent border of the glyph box wherever a stem touches the edge, which was visible on Terminus' `m`, whose left stem sits on column 0 of its 8px cell.

The render tests now cover the strike at its own 16px size as well as scaled to 32px, where the larger line is a pixel for pixel doubling of the smaller one. The fill test also gained a gradient fill, covering the fallback to black for paints an image cannot express.

oetiker:
One more, prompted by a good point about how pixel fonts are meant to work: a font carries a strike per size it was drawn for, and ships outlines for everything in between. Asking for an unscaled size, as the code did, always picks the largest strike and scales it, which throws that intent away and produces a worse glyph than the outline would.

Monochrome strikes are now only used when one matches the font size, unless the glyph has no outline to fall back to. Color bitmaps still take the largest strike, since an emoji font tends to carry a single large one for every size and no outline behind it, so nothing changes for those.

The test font grew a second strike so the sample shows both cases. At 16px and 24px the glyphs come straight from a strike, with two alpha levels and no antialiasing at all; at 20px and 28px they are antialiased outlines with 20+ levels.

RazrFalcon:
I'm not sure about this one. Bitmap fonts are extremely rare and this only adds complexity without solving much.

oetiker:
@RazrFalcon the reason I wrote these patches is that I use resvg to render eink panel content ... there bitmap fonts are very helpful. https://github.com/oetiker/byonk

RazrFalcon:
Ok, this makes a bit more sense.

I also do not like `cache_raster: HashMap<(ID, GlyphId, u32, [u8; 4])`. We should turn key into a struct.
Do we even need those two new fields? I saw the comment, but I would prefer to share the implementation with current bitmap fonts.

Also worth extracting some code from `text/flatten.rs` to a dedicated bitmap fonts handling file.

oetiker:
Both of the structural points are done, and I have an answer on the third.

The bitmap handling now lives in `text/bitmap.rs` together with its tests, leaving `text/flatten.rs` to walking the positioned glyphs of a span. And the cache key is a struct with a documented field per thing a bitmap glyph depends on, which doubles as the argument of the lookup itself, so the key and the parameters cannot drift apart.

On whether the two extra fields are needed: they are, and both fall out of features rather than being incidental. The size picks the strike, so the same glyph genuinely differs between sizes. The fill is there because a mask carries coverage only, with no color of its own, so it has to be painted in the fill of the span the way an outline glyph is; without it, colored text in a bitmap font comes out black. `ImageKind` can only carry encoded image data, so the color has to be baked in at that point rather than applied later.

On sharing the implementation with the existing bitmap path, I tried it. A single rule of "prefer the strike that matches the size, else the outline, else the best strike" leaves `text_color_font_cbdt` passing but breaks `text_color_font_sbix`, because an sbix font ships outlines behind its bitmaps as a fallback for renderers without sbix support, so preferring the outline inverts what the font intends.

That is what the split is really about, and I have reworded the comment to say so: a color bitmap is the intended appearance of the glyph, while a mask is one size specific rendering of the same design the outline already describes. That also makes the conservative behaviour the right one for a color font that does carry outlines, where scaling the color bitmap still beats falling back to a monochrome outline. Happy to reshape it if you would rather draw the line elsewhere.

RazrFalcon:
Good. The implementation looks much nicer and less invasive now.

PR #1128, opened 2026-08-31
`Text` feature represents shaping and layout, and text parsing is enabled by default

fundon:
The `text` feature represents shaping and layout, and text parsing is enabled by default.


crates/usvg/Cargo.toml
```diff
@@ -47,10 +47,10 @@ unicode-vo = { version = "0.1", optional = true }
 once_cell = "1.21"
 
 [features]
-default = ["svgz", "text", "system-fonts", "memmap-fonts", "writer"]
+default = ["svgz", "text", "system-fonts", "memmap-fonts"]
```
nicoburns: Why is `writer` no longer a default feature? Seems unrelated to this PR?

crates/usvg/src/parser/text.rs
```diff
@@ -137,9 +138,11 @@ pub(crate) fn convert(
         stroke_bounding_box: dummy,
         abs_stroke_bounding_box: dummy,
```
nicoburns: This is causing bounding boxes of all nodes to be expanded to include the origin when the `text` feature is disabled. Text nodes should be excluded from bbox computation in this case.

PR #1132, opened 2026-09-05
Add aarch64 release artifacts for Linux and Windows

harryzcy:
Adds native aarch64 builds to the release workflow.

- **Linux aarch64** on the `ubuntu-24.04-arm` runner → `resvg-linux-aarch64.tar.gz`, `usvg-linux-aarch64.tar.gz`
- **Windows aarch64** on the `windows-11-arm` runner → `resvg-win-aarch64.zip`, `usvg-win-aarch64.zip`

Updates the workflow's actions, two of which still declared the end-of-life node12 runtime:

- `alexellis/upload-assets` 0.2.2 → 0.5.0 (node12 → node24)
- `battila7/get-version-action` removed — unmaintained since 2021 and node12; the workflow only triggers on `v*` tags, so the version is just `${GITHUB_REF_NAME#v}`
- `actions/checkout` v2 → v7, `softprops/action-gh-release` v2 → v3

Renames the existing job to explicitly say `x86_64` and differentiate `aarch64`

Closes #968

PR #1133, opened 2026-09-08
Convert CI release job to use a Matrix

nicoburns:
Some CI cleanup suggested by LLM review of https://github.com/linebender/resvg/pull/1132. Has been verified as working on my fork.

## Changes made

- Collapse the six per-platform release-* jobs into one release-binaries matrix job over {linux, win, macos} × {x86_64, aarch64}. Asset names are derived from the matrix (`resvg-$SUFFIX.{tar.gz,zip}`). This renamea `resvg-win64.zip` / `usvg-win64.zip` to `resvg-win-x86_64.zip` / `usvg-win-x86_64.zip` to match the linux-* / macos-* / win-aarch64 naming. This is a breaking change for anyone scripting downloads of the old names.
- Drop the crates/c-api release builds (and the ilammy/msvc-dev-cmd step): their output was never collected or uploaded, so they only cost CI time.
- Split the vendored sources archive and the Windows Explorer thumbnailer into their own jobs
- Upload assets with `softprops/action-gh-release@v3` files: (already used to create the release) instead of the third-party `alexellis/upload-assets`.

### DIFF
--- a/.github/workflows/tagged-release.yml
+++ b/.github/workflows/tagged-release.yml
@@ -5,6 +5,9 @@ on:
     tags:
       - "v*"
 
+permissions:
+  contents: write
+
 env:
   CARGO_TERM_COLOR: always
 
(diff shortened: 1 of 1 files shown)


LaurenzV (approved):
If you've tested it LGTM!

PR #1135, opened 2026-09-13
fix(usvg): inherit `gradientTransform` and `patternTransform` via `href`

### LINKED ISSUE #1134: `gradientTransform` and `patternTransform` are not inherited through `href`
A gradient or pattern that uses another one as a template through `href`/`xlink:href` does not inherit its `gradientTransform` or `patternTransform`. Stops, coordinates, `gradientUnits` and `spreadMethod` are inherited correctly; only the transform is dropped.

### How to reproduce

1. Save this SVG as `gradient-href.svg`:

   ```xml
   <svg width="400" height="400" viewBox="0 0 200 200" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
     <!-- lg1 defines the colors and rotates the gradient 90 degrees: it should run top (gold) to bottom (navy) -->
     <linearGradient id="lg1" gradientTransform="rotate(90 0.5 0.5)">
       <stop offset="0" stop-color="gold"/>
       <stop offset="1" stop-color="navy"/>
     </linearGradient>
     <!-- lg2 reuses lg1 via href and sets nothing itself, so it must inherit the rotation -->
     <linearGradient id="lg2" xlink:href="#lg1"/>
     <rect x="20" y="20" width="160" height="160" fill="url(#lg2)"/>
   </svg>
   ```

2. Render it with resvg:

   ```sh
   resvg gradient-href.svg resvg-main.png
   ```

3. For comparison, render it with another renderer, for example librsvg:

   ```sh
   rsvg-convert gradient-href.svg -o rsvg-convert.png
   ```

`lg2` sets no attributes of its own, so it should inherit both the stops and the 90° rotation from `lg1`, and the gradient should run from gold at the top to navy at the bottom. resvg picks up the stops but not the rotation, and renders the gradient left to right.

| Before: resvg `main` (2f1aa34f) | After: resvg with the proposed fix |
|---|---|
| Before: resvg main renders the gradient left to right | After: resvg with the fix renders the gradient top to bottom |

The "after" image matches the rsvg-convert 2.58 output to within 1/255 per color channel (gradient rounding). The "before" image differs from it by up to 255.

The same happens with `radialGradient` (an inherited `translate` is lost) and `pattern` (a referencing pattern renders unrotated).

### Expected behavior

SVG 2 lists the attributes copied from a template element (https://www.w3.org/TR/SVG2/pservers.html):

- `linearGradient`: `x1`, `y1`, `x2`, `y2`, **`gradientTransform`**, `gradientUnits`, `spreadMethod`
- `radialGradient`: `cx`, `cy`, `r`, `fx`, `fy`, `fr`, **`gradientTransform`**, `gradientUnits`, `spreadMethod`
- `pattern`: `x`, `y`, `width`, `height`, `viewBox`, `preserveAspectRatio`, **`patternTransform`**, `patternUnits`, `patternContentUnits`

SVG 1.1 says the same in general terms: attributes defined on the referenced element but not on this element are inherited.

### History

This is a regression from 9101f30a ("Add support for the `transform-origin` attribute", November 2023). It was first released in v0.37.0 and is still present in v0.48.1.

Up to v0.36.0, all three converters in `paint_server.rs` (then in `crates/usvg-parser/`) read the transform through `resolve_attr`, which follows `href`:

```rust
let transform = resolve_attr(node, AId::GradientTransform)
    .attribute(AId::GradientTransform)
    .unwrap_or_default();
```

To apply `transform-origin`, that commit replaced this with `node.resolve_transform(AId::GradientTransform, state)` (and the `PatternTransform` equivalent in `convert_pattern`). `resolve_transform` only reads the attribute from `node` itself, so the `href` lookup was lost. The `GradientTransform` entries in `resolve_lg_attr` and `resolve_rg_attr` are still present but no longer used.

None of the existing `attributes-via-xlink-href` tests set a transform on the referenced element, so the regression was not caught.

### Proposed fix

In `convert_linear`, `convert_radial` and `convert_pattern` (`crates/usvg/src/parser/paint_server.rs`), resolve the element that defines the transform with `resolve_attr` first, then call `resolve_transform` on it:

```diff
-    let transform = node.resolve_transform(AId::GradientTransform, state);
+    let transform =
+        resolve_attr(node, AId::GradientTransform).resolve_transform(AId::GradientTransform, state);
```

```diff
-    let transform = node.resolve_transform(AId::PatternTransform, state);
+    let transform =
+        resolve_attr(node, AId::PatternTransform).resolve_transform(AId::PatternTransform, state);
```

This keeps `transform-origin` support, and a transform set on the element itself still takes precedence. With this change the reproducer matches rsvg-convert, and the existing test suite passes with no reference image changes.

killerdevildog:
Fixes #1134.

A gradient or pattern that references another one through `href` did not inherit its `gradientTransform` or `patternTransform`. For example, `<linearGradient id="lg2" xlink:href="#lg1"/>`, where `lg1` has `gradientTransform="rotate(90 0.5 0.5)"`, rendered left to right instead of top to bottom.

This regressed in 9101f30a (first released in v0.37.0), which replaced `resolve_attr(node, AId::GradientTransform).attribute(..)` with `node.resolve_transform(..)` to support `transform-origin`. `resolve_transform` only reads the element itself, so the `href` lookup was lost and the `GradientTransform` entries in `resolve_lg_attr` and `resolve_rg_attr` became unused. Resolving the defining element with `resolve_attr` first restores inheritance and keeps `transform-origin` support. A transform set on the element itself still takes precedence. When the transform is inherited, `transform-origin` is read from the same element that defines the transform.

| Before | After |
|---|---|
| Before: resvg main renders the gradient left to right | After: resvg with the fix renders the gradient top to bottom |

This PR changes only `crates/usvg/src/parser/paint_server.rs`.

Validation: the reproducer from #1134 fails before the change and passes afterward; resvg's output now matches rsvg-convert 2.58 to within 1/255 per color channel (before the change it differed by up to 255). Local paired cases covering linear, radial and pattern templates, a radial gradient using a linear template, a two-level `href` chain, and an element that overrides the inherited transform each render pixel-identically to the same transform set directly on the element. `cargo test --release -p resvg -p usvg` passes (1731 integration tests, no reference image changes) and `cargo fmt --check` is clean. The local regression tests are not included in this PR. Tested on Linux only.


killerdevildog:
I'll work on it tonight.

PR #1136, opened 2026-09-15
tests: load test fonts in a deterministic order

luisbg:
`fontdb::load_fonts_dir` walks the directory with `std::fs::read_dir`, which outputs entries in filesystem order. That order is not stable across machines, and the test harness depends on it.

Replace `load_fonts_dir` with an explicit sorted list so order is identical everywhere. The whole suite passes unchanged, so no stored reference image depended on the previous ordering.


DJMcNab (approved):
Sure. Test code is low enough risk, especially if CI still passes...

This probably should use an unstable sort, but I can't be bothered fixing that.

